PR #12, opened 2022-11-28
Adding some basic Excalidraw-ings

gunnarmorling:
Hey @tombentley, @k-wall, just came across this drawing from a while ago. You can load it into excalidraw.com to modify it. I thought it might be useful to have it here for reference.


SamBarker (approved):
LGTM.

Thanks @gunnarmorling

PR #24, opened 2023-02-27
Proposal template

k-wall:
Using Github design discussions are get unwieldy when there have been many comments.  It's hard to see where agreement has been reached and where not.  Let's try a proposal mechanism.  I've based the template on Strimzi's to get us going.


SamBarker (approved):
As far as template goes I think its a good start.

I remain skeptical as to the value of proposals but happy to try it. 

I think the lack of clarity in the multi tenant MVP discussion is reflective of the lack of established clarity rather than the discussion mechanism. Maybe a PR on the proposal will work as a forcing function to bring consensus and action.

PR #26, opened 2023-03-03
Create 001-ClusterRepresentation.md

k-wall:


proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+#### Downstream Clusters
+
+Conceptually the _downstream cluster_ represents the cluster that the kafka client connects to.  The downstream clusters list specifies
+all the kafka clustsers are being presented by Kroxylicious. 
+
+A _downstream clustser_ references exactly one _upstream cluster_. 
+
+A _downstream cluster_ may reference a `filter chain`.  This provides zero or more filters that the RPCs will pass through before arriving at the upstream cluster.
+
+A _downstream cluster_ enumerates the downstream brokers that comprise the downstream cluster.
```
robobario, line 67: Could we add something here explicitly stating the relationship between Downstream Cluster and Endpoint?

Something like `A downstream cluster is associated with a single Endpoint. Multiple downstream clusters may refer to the same Endpoint`

Same with TLS like `A downstream cluster can provide TLS configuration which is used for XYZ`

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+##### Downstream Brokers
+
+Each broker has a reference to an endpoint that will provide the way in to the broker.   For the SNI case, there is also an addresses matcher.  It is an error
+for more than one broker to share an endpoint unless the endpoint type is TLS and the SNI matching address is defined.
+
+Each downstream broker references exactly one upstream broker which must exist within the referenced upstream cluster.  It is an error if there is not a
+one-to-one correspondence between downstream and upstream brokers.
+
+TLS key material may be provided either at the downstream cluster level or at the individual broker level.
```
robobario, line 77: What would the cluster-level TLS configuration be used for? We know the broker SNI hostnames, but would we know the bootstrap hostname to apply it to? Should a public hostname be part of the downstream cluster configuration?

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+### Responsibilities
+
+All specified endpoints will be bound.
+
+When a connection is made to an endpoint,  the system must resolve that connection to a downstream broker and hence a downstream cluster.
+
+To do this, it resolves the endpoint and any SNI information against the model.   This should yield exactly one broker belonging to a downstream cluster.
+It is an error otherwise and the connnection must be closed.
+
+If TLS is in use, the SSLContext can be generated from the downstream cluster definition.  This will be passed to Netty to let it complete the TLS handshake.
```
SamBarker, line 160: This feels very implementation specific compared to the other responsibilities. 

I see it as an important point to

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+A _downstream cluster_ enumerates the downstream brokers that comprise the downstream cluster.
+
+##### Downstream Brokers
+
+Each broker has a reference to an endpoint that will provide the way in to the broker.   For the SNI case, there is also an addresses matcher.  It is an error
+for more than one broker to share an endpoint unless the endpoint type is TLS and the SNI matching address is defined.
+
+Each downstream broker references exactly one upstream broker which must exist within the referenced upstream cluster.  It is an error if there is not a
+one-to-one correspondence between downstream and upstream brokers.
```
SamBarker, line 75: Does this hold true in the multi tenant use case? @robobario and I are imagining providing N `virtual` clusters compose of `3` brokers each being serviced by a **single** `physical` cluster composed of `6` brokers. 

We could establish fixed mappings from `virtual` brokers to `physical` brokers initially but when we use cruse control to rebalance topics across the `6 physical` brokers those fixed mapping no longer hold.  As the partitions could be moved to the brokers not included in the one to one mapping. We potentially need to be able to map multiple `physical` brokers to a single `virtual` broker. 


We may not need to support this the MVP but any design we come up with should be open to extending in this direction. In the MVP we could enforce that all `virtual` clusters have the same broker counts as the `physical` clusters.

robobario:
Thanks for the PR @k-wall :bow:

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+  - name: broker-2
+    address: broker-1-private:9092
+    ...
+    tls:
+      trust:
+```
+
+#### Filter Chain
+
+Currently kroxylicious provide a single filter chain.  It will refactored so that we can support many filter chains, each identified by name. The `downstream cluster` and `upstream cluster` may reference a filter chain.
```
SamBarker, line 139: I'm starting to think it might be better to express this as the `virtual` cluster defines a filter chain (potentially applied by the `frontendHandler` and the `physical` cluster defines a filter chain (again applied by the `backendHandler`). Those `virtual` and `physical` chain definitions could potentially be shared between multiples of each cluster type but there is no single over arching chain definition.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+    key:
+    cert:
+```
+
+
+#### Upstream Clusters
+
+An _upstream cluster_ represents a physical kafka cluster.   The upstream clusters list specifies all the kafka clustsers that are being exposed through Kroxylicious.  
+
+An _upstream cluster_ may reference a `filter chain`.  This provides zero or more *additional* filters that the RPCs will pass through before passing to the brokers
```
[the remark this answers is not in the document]
k-wall: yes, that matches my mental model.

proposals/001-ClusterRepresentation.md
k-wall: It is for the certificate per cluster use-case. The cert's SAN would need to list the FQDN of each broker within the cluster.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+If TLS is in use, the SSLContext can be generated from the downstream cluster definition.  This will be passed to Netty to let it complete the TLS handshake.
+
+The downstream broker and downstream cluster is used to identify the upstream broker and upstream cluster.
+
+The upstream cluster and downstream broker provide the filter chain.
+
+The handler chain connects to the upstream broker.
+
+The system must reload chnages to the model dynammically, without dropping established connections.
```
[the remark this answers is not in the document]
k-wall: > I think we should make it explicit that the reloaded config is applied by the existing connections.

yeah - I was pondering whether reconfiguring an upstream or downstream broker should close the established connections.  similarly for reconfiguring an upstream or downstream cluster should that close all the established connections of all the brokers?

I'm not sure how important it is to consider right now.  I think they might all be edge cases or require no action.

1. I don't think a use-case where a downstream cluster is pointed at a different upstream cluster is likely to materialise in the short term (tar pitting might be a use-case)
2. TLS cert renewal.  I don't think there is an necessity to disconnect existing connections.
3. Upstream cluster topology changes
- fewer brokers - connections will die as the physical upstream broker is stopped.
- more brokers - kafka's metadata discovery will handle this. No advantage gained by interrupting the connections to the existing brokers.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Broker Address Filter
+
+How will broker address filter map the RPCs that contain the upstream broker address to the downstream broker addresss that are resolveable to the client?
+
+In the case where kroxy is being placed in front of a kafka cluster spanning a three-AZ cluster, the kroxylicious instances in the AZ won't need configuration to
+*connect* to brokers in the other AZs.  However, BrokerAddressFilter will be to be capable of rewriting the the broker address for the whole cluster.
+This leads us to the conclusion that Broker Address filter need separate configuration independent of that what could be derived from the upstream/downstream cluster mapping.
+
+TODO: Maybe used a regexp based mapping would be sufficient
```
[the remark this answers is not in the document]
k-wall: Yes I believe a mapping will work.  The mapping will need the port numbers.  There's the possibility that downstream and upstream port numbers would be different.

I've been thinking about where this mapping lives.   If it is exists as filter configuration, then in the multitenant case, we'd need a separate filter configuration for each tenant.  Whilst that would work, I'm wondering whether the cluster configuration should have have a map of arbitrary configuration that is made available to the filters at runtime (somehow).  

We could use this to pass extra information to the filters like the MultiTenantFilter (a tenant key) and BrokerAddressFilter (a mapping).

```yaml
- name: tenant1-cluster
  upstreamClusterRef: my-private-cluster
  filterChainRef: my-upstream-filter-chain
  filterVariables:
      io.kroxylicious.tenantKey: xyz4a
      io.kroxylicious.brokerAddressMapping: 
        broker-1.azA.internal.svc:1234: broker-A.myvanitydomain.com:9080
```

I think some of the variable name would be 'well-known' (we reserve names variable names beginning with `io.kroxylicious`.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+```yaml
+- name: my-public-cluster
+  upstreamClusterRef: my-private-cluster
+  filterChainRef: my-upstream-filter-chain
+  brokers:
+  - name: broker-1
+    endpointRef: my-tls-endpoint
+    sniMatchAddress: broker-1-public.example.com
+    upstreamBrokerRef: broker-1
```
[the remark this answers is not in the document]
k-wall: Yes, do you see a problem with that assumption?

proposals/001-ClusterRepresentation.md
k-wall: > Does this hold true in the multi tenant use case? @robobario and I are imagining providing N `virtual` clusters compose of `3` brokers each being serviced by a **single** `physical` cluster composed of `6` brokers.

Can that work?   I had been working under the (unspoken) assumption that a virtual cluster had n virtual brokers where n matched the number of the physical brokers on the cluster. It is possible I'm the outlier here.

> We could establish fixed mappings from virtual brokers to physical brokers initially but when we use cruise control to rebalance topics across the 6 physical brokers those fixed mapping no longer hold.

Exactly.

> We potentially need to be able to map multiple physical brokers to a single virtual broker.

I'm not sure that could work.  When a topic is created, the kafka client decides which broker each partition/replicas go.   If there's not a one to one mapping between virtual broker and physical ones I think this would break down.

It might be possible to have kroxy expose only those brokers hosting partitions where the tenant has partitions but this does seen to bring a fair amount of extra complexity.

What's the advantage for exposing only a subset of the physical brokers?

proposals/001-ClusterRepresentation.md
k-wall: Yes, maybe.  if each virtual cluster/cluster declares the filter chain in its own right, there's no problem passing cluster specific configuration.

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
SamBarker: I was assuming we would generate the mapping table from the configuration rather than listing it explicitly as we are currently talking about having exhaustive mappings. 

But yeah maybe including it in the config is just simpler and easier to deal with for now.

proposals/001-ClusterRepresentation.md
robobario: Okay, I was more thinking of if we have a separate bootstrap host like:
`my-kafka-a.b.c.robeyoun-kafka.public.host`
and then per-broker hosts like `broker-1.my-kafka-a.b.c.robeyoun-kafka.public.host`

Where would we know how to map the bootstrap host to the downstream cluster SSL configuration?

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
+# Cluster Representation in the Model
+
+This proposal introduces the concept of _cluster_ into the model.
+
+A _downstream cluster_ is a kafka cluster that the clients connect to.   The downstream cluster can be thought of as a 'virtual' cluster or a 'facade' to
```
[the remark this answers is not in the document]
SamBarker: Yeah. I'm much less sure about `physical` too. Toyed with `underlying` and `target` but `physical` worked well for as being the opposite of `virtual` and really drives the point home for me. 

As far as each Kroxy is concerned each upstream cluster is the `physical` one even if it happens to be another kroxy instance thats an implementation detail that doesn't concern the given instance (otherwise we have a leaky abstraction - but I can see arguments for co-operation between proxies...).

proposals/001-ClusterRepresentation.md
robobario: edit: maybe we can model bootstrap as a downstream and upstream broker.

proposals/001-ClusterRepresentation.md
SamBarker: On reflection I don't think `N physical` brokers to `1 virtual` broker doesn't work and is unnecessary but I think we do need to support a subset of brokers (at some point).

The use-case I am thinking of is something like offering small clusters on a much larger shared cluster, in the tens or hundreds of brokers. I don't see it making any realistic sense to expose hundreds of brokers through the proxy to a single `virtual` cluster. If we expand to that point having a static map of `virtual` clusters to `physical` brokers becomes very difficult to manage. 

For the record I'm not saying we need to fix this during MVP, just that we need to keep this end goal in mind and keep options open that make that support easier.

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
devguyio: I struggled with the `upstream` and `downstream` terms as well, I think `physical` and `virtual` are much better replacements and are explicit in nature to spare the users any non-intuitive mental model mapping (e.g. "oh it mentions "cluster" only, so this must be the physical one, since the other one is called "virtual" ) 

I think either `virtual` and `physical` or `virtual` and `target` would be good alternatives

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+one or more _upstream clusters_.    An _upstream cluster_ represents the physical kafka cluster.  There is a one to one correspondance between
+a _upstream cluster_ and a physical cluster in the world.
+
+There will be a mechansim to route flexibly between downstreams and upstreams.  This will allow useful topologies to built to serve different use-cases.
```
devguyio, line 11:
```suggestion
There will be a mechansim to route flexibly between downstreams and upstreams.  This will allow building useful topologies to serve different use-cases.
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+- *one to one* - the simple exposure of a physical cluster 
+- *many to one* - kroxy presents many downstream clusters which route to a single upstream - this would support a multitenant use-cases 
+
+(There's a possiblity of *one-to-many* which a single downstream routes to several upstreams as if they were one, which might be useful for presenting several
+kafka clusters as if they were one, however, supporting transactions across two clusters would be difficult.  We consider won't this use-case futrther).
```
devguyio, line 17:
```suggestion
kafka clusters as if they were one, however, supporting transactions across two clusters would be difficult.  We won't consider this use-case further).
```

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
k-wall: I'll adopt virtual/physical nomenclature.

proposals/001-ClusterRepresentation.md
k-wall: > For the record I'm not saying we need to fix this during MVP, just that we need to keep this end goal in mind and keep options open that make that support easier.
+1

I'll call that out as a goal.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
...
+
+
+
+
+#### Endpoints
+
+Currently `KafkaProxy` uses the `ProxyConfig` object to get a single address to bind to.  
+
+* For the non-TLS case, multiple endpoints are required to support >1 brokers. 
+* For TLS case, SNI gives a mechanism to route the traffic to a >1 brokers even when a single endpoint is used, however there may be performance or operational reasons to favour the ability to define endpoints.
```
devguyio, line 48:
```suggestion
* For TLS case, SNI gives a mechanism to route the traffic to >1 brokers even when a single endpoint is used, however there may be performance or operational reasons to favour the ability to define endpoints.
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,305 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
+# Cluster Representation in the Model
+
+This proposal introduces the concept of _cluster_ into the model.
+
+A _virtual cluster_ is a kafka cluster that the clients connect to.   The virtual cluster can be thought of as a 'virtual' cluster or a 'facade' to
```
devguyio, line 7:
```suggestion
A _virtual cluster_ is a Kafka cluster that the clients connect to.  It can be thought of as a 'logical' cluster or a 'facade' to
```

Thinking either use the word logical here or just remove this part and make it `can be thought of as a facade..` .

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,302 @@
...
+
+```yaml
+- name: my-public-cluster
+  physicalClusterRef: my-private-cluster
+  filterChainRef: my-upstream-filter-chain
+  brokers:
+  - name: broker-1
+    endpointRef: my-tls-endpoint
+    sniMatchAddress: broker-1-public.example.com
+    upstreamBrokerRef: broker-1
```
devguyio, line 94:
```suggestion
    physicalBrokerRef: broker-1
```

I presume that we'll use the new terminology

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,302 @@
...
+  # cluster wide TLS cert
+  tls:
+    key:
+    cert:
+```
+
+
+#### Physical Clusters
+
+An _physical cluster_ is the in-model representation of a running kafka cluster.   The physical clusters list specifies all the kafka clusters that known to Kroxylicious.
```
devguyio, line 110:
```suggestion
A _physical cluster_ is the in-model representation of a running kafka cluster.   The physical clusters list specifies all the kafka clusters that are known to Kroxylicious.
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,302 @@
...
+
+An _physical cluster_ may reference a `filter chain`.  This provides zero or more *additional* filters that the RPCs will pass through before passing to the brokers
```
devguyio:
```suggestion
A _physical cluster_ may reference a `filter chain`.  This provides zero or more *additional* filters that the RPCs will pass through before passing to the brokers
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,306 @@
...
+of the physical cluster.  
+
+An _physical cluster_ enumerates the brokers that comprise the physical cluster.
```
devguyio, line 116:
```suggestion
A _physical cluster_ enumerates the brokers that comprise the physical cluster.
```

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
k-wall: > I was assuming we would generate the mapping table from the configuration rather than listing it explicitly as we are currently talking about having exhaustive mappings.

That was my initial thinking too.  I don't think that works. 

- Firstly, in the case where you have an AZ specific kroxy deployment, the mapping needs to know about the brokers on the other AZs.
- Secondly, there's no guarantee that the endpoint address/port that kroxy needs to bind is the same as the address/port that actually needs to be advertised to the client.

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
k-wall: Done

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
k-wall: Done

proposals/001-ClusterRepresentation.md
k-wall: Done.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,306 @@
...
+for more than one broker to share an endpoint unless the endpoint type is TLS and the SNI matching address is defined.
+
+Each virtual broker references exactly one physical broker which must exist within the referenced physical cluster.  It is an error if there is not a
+one-to-one correspondence between downstream and upstream brokers.
+
+TLS key material may be provided either at the virtual cluster level or at the individual broker level.
+
+
+```yaml
+- name: my-public-cluster
```
devguyio, line 83: I found it confusing when the yaml didn't include the `virtualClusters` key. I think it's more clear if it's something like this:
```yaml
...
virtualClusters:
- name: my-public-cluster
  upstreamClusterRef: my-private-cluster
  filterChainRef: my-downstream--filter-chain
  tls:
    key:
    cert:
  brokers:
 name: my-public-cluster
  upstreamClusterRef: my-private-cluster
  filterChainRef: my-upstream-filter-chain
  brokers:
  - name: broker-1
    endpointRef: my-tls-endpoint
    sniMatchAddress: broker-1-public.example.com
    physicalBrokerRef: broker-1
    # or brokers themselves provide the key/cert
    - tls:
        key:
        cert:
  - name: broker-2
    ...
  # cluster wide TLS cert
  tls:
    key:
    cert:
    ...
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,306 @@
...
+An _physical cluster_ enumerates the brokers that comprise the physical cluster.
+
+#### Physical Brokers
+
+Each physical broker specifies the address of a broker and whether TLS is to be used.  It will also specify trust settings.
+
+TLS trust material may be provided at the upstream broker level or at the individual broker level.
+
+```yaml
+- name: my-private-cluster
```
devguyio, line 125: same comment as above about including the `physicalClusters:` key

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,306 @@
...
+When a connection is made to an endpoint,  the system must resolve that connection to a virtual broker and hence a virtual cluster.
+
+To do this, it resolves the endpoint and any SNI information against the model.   This should yield exactly one broker belonging to a virtual cluster.
+It is an error otherwise and the connnection must be closed.
+
+If TLS is in use, the SSLContext can be generated from the virtual cluster definition.  This will be passed to Netty to let it complete the TLS handshake.
+
+The virtual broker and virtual cluster is used to identify the physical broker and physical cluster.
+
+The physical cluster and downstream broker provide the filter chain.
```
devguyio:
```suggestion
The physical cluster and physical broker provide the filter chain.
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,306 @@
...
+
+The handler chain connects to the upstream broker.
+
+The system must reload chnages to the model dynammically, without dropping established connections.
+
+### Broker Address Filter
+
+How will broker address filter map the RPCs that contain the upstream broker address to the downstream broker addresss that are resolveable to the client?
```
devguyio:
```suggestion
How will broker address filter map the RPCs that contain the physical broker address to the virtual broker addresss that are resolveable to the client?
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,306 @@
...
+
+In the case where kroxy is being placed in front of a kafka cluster spanning a three-AZ cluster, the kroxylicious instances in the AZ won't need configuration to
+*connect* to brokers in the other AZs.  However, BrokerAddressFilter will be to be capable of rewriting the the broker address for the whole cluster.
```
devguyio:
```suggestion
*connect* to brokers in the other AZs.  However, BrokerAddressFilter will have to be capable of rewriting the broker address for the whole cluster.
```

devguyio (commented):
A batch of comments

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,341 @@
...
+
+#### Model Changes
+
+The system must reload chnages to the model dynammically, without dropping established connections.
+
+#### Broker Address Filter
+
+The BrokerAddressFilter must map the RPC reponses that contain the physical broker address to the virtual broker addresses that are resolveable to the client.  DescribeCluster response is an example of an RPC that needs to be mapped.
+
+In the case where Kroxylicious is being placed in front of a kafka cluster spanning a three-AZ cluster, the kroxylicious instances in the AZ won't need configuration to *connect* to brokers in the other AZs.  However, BrokerAddressFilter must be capable of rewriting the the broker address for the whole cluster.
```
devguyio, line 179:
```suggestion
In the case where Kroxylicious is being placed in front of a kafka cluster spanning a three-AZ cluster, the kroxylicious instances in the AZ won't need configuration to *connect* to brokers in the other AZs.  However, BrokerAddressFilter must be capable of rewriting the broker address for the whole cluster.
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,341 @@
...
+This leads us to the conclusion that Broker Address filter need separate configuration, independent of that what could be derived from the virtual/physical cluster mapping.
+
+The BrokerAddressFilter will accept a mapping, which will map the upstream broker addresses to routable addresses that correspond to the virtual brokers of the
```
devguyio:
```suggestion
The BrokerAddressFilter will accept a mapping, which will map the physical broker addresses to routable addresses that correspond to the virtual brokers of the
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,341 @@
...
+        broker-2.azB.internal.svc:19080: broker-B.myvanitydomain.com:9092
+        broker-3.azC.internal.svc:19080: broker-C.myvanitydomain.com:9092
+```       
+
+
+### Example Config Files
+
+#### One to one case.
+
+# Single Cluster with 3 broker cluster, exposed via kroxy
```
devguyio, line 200:
```suggestion
# Single Cluster with 3 brokers, exposed via kroxylicious
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+
+```yaml
+endpoints:
+- name: my-tls-endpoint
+  type: TLS
+  bindAddress: 0.0.0.0:9092
+virtualClusters:
+- name: my-public-cluster
+  upstreamClusterRef: my-private-cluster
```
devguyio:
```suggestion
  physicalClusterRef: my-private-cluster
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+        broker-2.azB.internal.svc:19080: broker-2.my-tenant1.kafka.com:9092
+        broker-3.azC.internal.svc:19080: broker-3.my-tenant1.kafka.com:9092
+  - type: MultiTenantFilter
+    config:
+      tenantKey: z4de # prefix used to prefix objects owned by the tenant.
+  brokers:
+  - name: broker-1
+    sniMatchAddress: broker-1.my-tenant1.kafka.com
+    endpointRef: my-tls-endpoint
+    upstreamBrokerRef: private-broker-1
```
devguyio:
```suggestion
    physicalBrokerRef: private-broker-1
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+  - name: broker-2
+    sniMatchAddress: broker-2.my-tenant1.kafka.com
+    endpointRef: my-tls-endpoint
+    upstreamBrokerRef: private-broker-2
```
devguyio:
```suggestion
    physicalBrokerRef: private-broker-2
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+  - name: broker-3
+    sniMatchAddress: broker-3.my-tenant1.kafka.com
+    endpointRef: my-tls-endpoint
+    upstreamBrokerRef: private-broker-3
```
devguyio:
```suggestion
    physicalBrokerRef: private-broker-3
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+- name: my-tenant2
+  upstreamClusterRef: my-big-cluster
```
devguyio:
```suggestion
  physicalClusterRef: my-big-cluster
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+        broker-2.azB.internal.svc:19080: broker-2.my-tenant2.kafka.com:9092
+        broker-3.azC.internal.svc:19080: broker-3.my-tenant2.kafka.com:9092
+  - type: MultiTenantFilter
+    config:
+      tenantKey: ab7ge
+  brokers:
+  - name: broker-1
+    sniMatchAddress: broker-1.my-tenant2.kafka.com
+    endpointRef: my-tls-my-tls-endpoint
+    upstreamBrokerRef: private-broker-1
```
devguyio:
```suggestion
    physicalBrokerRef: private-broker-1
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+  - name: broker-2
+    sniMatchAddress: broker-2.my-tenant2.kafka.com
+    endpointRef: my-tls-endpoint
+    upstreamBrokerRef: private-broker-2
```
devguyio:
```suggestion
    physicalBrokerRef: private-broker-2
```

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,343 @@
...
+  - name: broker-3
+    sniMatchAddress: broker-3.my-tenant2.kafka.com
+    endpointRef: my-tls-endpoint
+    upstreamBrokerRef: private-broker-3
```
devguyio:
```suggestion
    physicalBrokerRef: private-broker-3
```

devguyio (approved):
Thanks for putting this together @k-wall ! Here are my thoughts
* I think this is a very good start to expand Kroxylicious beyond a single cluster model.
* I think this is a good level of abstraction that would still keep kroxylicious unopinionated about its usecase yet dynamically configurable. 
* I can imagine there are future improvements that can take place on the model, the configuration syntax, static vs dynamic and discoverable configuration etc. But this is for me a good starting point for the MVP.

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
k-wall: I'm now passing separate configuration to BrokerAddressFilter

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
SamBarker: I think we've addressed the underlying challenges with this else where.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,344 @@
...
+This leads us to the conclusion that Broker Address filter need separate configuration, independent of that what could be derived from the virtual/physical cluster mapping.
+
+The BrokerAddressFilter will accept a mapping, which will map the addresses returned by the upstream broker to routable addresses that route to the virtual brokers of the cluster in question.
+
+To do this, we can use the existing filter configuration mechanism.
+
+```
+  - type: BrokerAddress
+    config:
+      brokerAddressMapping:
```
SamBarker, line 193: If we add the `advertisedAddress` to the `physical broker` definition we already have the relationship established and I worry about having to generate the mappings twice in the config file. 

We could then provide the `brokerAddressMapping` as a derived property of the `virtual cluster` configuration when injected into the filters. However that implies the `virtual config` stops being a dumb value object and starts being a right domain object.

As an aside this thought was leading down the track of wondering about how to do Inversion Of Control (IOC) or Dependency Injection (DI) with filter instances. However I think making the derived address mapping a property of the `virtual cluster` config side steps that issue and leaves it in the long grass.

SamBarker (commented):
Like @devguyio I approve in principal, I think there a few details left to iron out around particularly around the brokerAddress mapping.

tombentley:
1. It's not clear to me whether we really need virtual and physical brokers to be things surfaced in the configuration. I can see that doing it that way _starts out_ simpler, but that's because you've hand-waved some of the essential complexity into an operator which will need to orchestrate reconfiguration, coordinating with orchestration of the broker cluster(s). 
\
One possible alternative would be to allocate endpoints automatically, based on the information about `Nodes` which propagates to clients via `MetadataResponse`, `DescribeClusterResponse` and `FindCoordinatorResponse`. You'd need a suitable function (a bijection, actually) for allocating the SNI matcher and port (in particular it needs to guarantee that no two brokers, even from different physical clusters, can map to the same endpoint). But that's relatively trivial. You'd also need the machinery to trigger the socket bind based on that. However, I think this might significantly simply operations:

  * Scaling in a broker cluster doesn't require external orchestration of the proxy cluster. New endpoints get created automatically. 
  * There is less validation of the configuration (i.e. you can assume the known+tested properties of the endpoint allocation bijection)
  * Even under the brokers-as-configurables model, this bijection would eventually need to exist anyway, except it would be in the operator.

2. I think we need to support mapping Kafka `clusterId`s. I.e. A `clusterId` needs to be a property of both virtual and physical clusters, with support for mapping between the two. This is necessary so you can't infer multitenancy from observing the same clusterId (e.g. in `MetadataResponse`) for two distinct virtual clusters which happen to (currently) share the same physical cluster. 

3. It's not clear to me whether what's proposed here makes it harder to implement "simple proxies". I think it would be good to think about how minimal we could make the configuration for the simple proxy case.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,344 @@
...
+
+A _virtual cluster_ enumerates the virtual brokers that comprise the virtual cluster.
+
+##### Virtual Brokers
+
+Each broker has a reference to an endpoint that will provide the way in to the broker.   For the SNI case, there is also an addresses matcher.  It is an error
+for more than one broker to share an endpoint unless the endpoint type is TLS and the SNI matching address is defined.
+
+Each virtual broker references exactly one physical broker which must exist within the referenced physical cluster.  It is an error if there is not a
+one-to-one correspondence between physical and virtual brokers.
```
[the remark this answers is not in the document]
k-wall: By physical broker here I am referring to the model object, rather than the the actual Kafka Cluster. Am I trying to say that the brokers defined by virtual cluster configuration needs to match the brokers the defined by the physical cluster model object.

For a scaling event, after scaling up the number of brokers on the Apache Kafka Cluster, an operator would update both the virtual cluster definition and the physical cluster definition consumed by Kroxy in lock step.

At the heart of this proposal is that the user of kroxy configures kroxy to know about the target cluster topology (i.e. static configuration enumerating the target cluster's brokers) and that configuration can be updated at runtime.    An alternative (not explored by this proposal) is let kroxy discover the broker topology of the target cluster.   Kroxy configuration would just need a bootstrap url of the target cluster.  If we took this option, we would need to require TLS connections at the downstream (client side).  This is so that SNI could be used to distinguish the broker traffic arriving at a single endpoint..

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,344 @@
...
+
+
+#### Endpoints
+
+Currently `KafkaProxy` uses the `ProxyConfig` object to get a single address to bind to.  
+
+* For the non-TLS case, multiple endpoints are required to support >1 brokers. 
+* For TLS case, SNI gives a mechanism to route the traffic to >1 brokers even when a single endpoint is used, however there may be performance or operational reasons to favour the ability to define endpoints.
+
+`ProxyConfig` will be replaced by an list of _Endpoint_ objects.
```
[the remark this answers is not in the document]
k-wall: I was thinking the endpoint specify if it was TLS or not. It would probably also be where we configure cipher suites etc though I haven't specified this.  If we pushed the endpoint configuration into the upstream, we'd need to resolve any conflicts. I think having named endpoints is simpler to avoid those issues.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,344 @@
...
+
+A physical cluster_ may define a chain of filters.  This provides zero or more filters that the RPCs will pass through as they traverse the physical cluster.
+(Note that it is possible for both a filter chain to be define at both the virtual cluster and physical cluster level.  RPCs pass through both sets of filters
+in the order they are defined).
+
+An _physical cluster_ enumerates the brokers that comprise the physical cluster.
+
+#### Physical Brokers
+
+Each physical broker specifies the address of a broker and whether TLS is to be used.  It will also specify trust settings.
```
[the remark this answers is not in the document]
k-wall: I don't understand how that could work.

The virtual brokers needs to be in charge of the advertised address.   For the many-to-one case, the my broker1 FQDN needs to be different to your broker1 FQDN.

proposals/001-ClusterRepresentation.md
```diff
@@ -0,0 +1,344 @@
...
+  - type: Filter1
+    config:
+       foo: bar
+  - type: Filter2
+  tls:
+    trust:
+    ...
+  brokers:
+  - name: broker-1
+    address: broker-1-private:9092
```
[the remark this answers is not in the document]
k-wall: Like I said above I don't see how having advertisedAddress as an attribute of the physical broker helps.

k-wall:
Thanks for the great reply.


> 
One possible alternative would be to allocate endpoints automatically, based on the information about Nodes which propagates to clients via MetadataResponse, DescribeClusterResponse and FindCoordinatorResponse. You'd need a suitable function (a bijection, actually) for allocating the SNI matcher and port (in particular it needs to guarantee that no two brokers, even from different physical clusters, can map to the same endpoint). But that's relatively trivial. You'd also need the machinery to trigger the socket bind based on that.

The design was avoiding the need to dynamically bind the listening port for the virtual brokers.  That's what lead to statically defining the virtual brokers, and as a consequence lead, to static definition of the physical brokers too.  I'd love to avoid that configuration too.

Why do I think dynamic ports are a problem?  If we have kroxy dynamically assigning ports, does that not present problems exposing a service at a kube level?  I imagine a kroxy operator will add kube services corresponding to each virtual broker to route the traffic to the kroxy pod.  We will want to be able to scale kroxy replicas.  If we have a reliance on  dynamically assigned ports, different kroxy instance will likely bind to different port.   I'm not aware of a way to express that at Kubernetes service.   The `targetport` for tenant A's broker1 needs to be the same for every kroxy replica.

So I think for good kube integration, kroxy needs to support a statically defined port binding.

If we were willing to accept that kroxy's downstream exclusively supported TLS, we could use a statically defined listening port and use SNI to route to the different virtual cluster and their brokers.  I think it is likely production use-cases will be TLS, so I don't really see that as a problem to production users.

However, the development/testing/demos on a laptop, having plain connections without the faff on TLS certs and `/etc/hosts` munges is really appealing.

We could try to find a way to express both binding models.  WDYT?


1





  
If you had kroxy dynamically assigning

k-wall:
> I think we need to support mapping Kafka clusterIds. I.e. A clusterId needs to be a property of both virtual and physical clusters, with support for mapping between the two. This is necessary so you can't infer multitenancy from observing the same clusterId (e.g. in MetadataResponse) for two distinct virtual clusters which happen to (currently) share the same physical cluster.

That's a good point, I hadn't thought of that.  We could have a filter that intercepts the `DescribeClusterResponse`and produces a stable clusterid from the digest of the physical cluster's cluster id and some immutable property(s) of the upstream cluster definition.  That might be the FQDN of the cluster's bootstrap or in the case of multi-tenant a tenant prefix).   We'd need to ensure upstream cluster id stability across kroxy replicas and pod restarts.

k-wall:
> You'd also need the machinery to trigger the socket bind based on that.

I think that's doable.  Kroxy would need to bind listening sockets corresponding for the upstream clusters' bootstrap addresses.  Kroxy would somehow need to snoop the `describecluster` or `metadata` response and use that as a signal to bind the listening sockets (for the brokers). Then on each subsequent connection it needs to validate if the set of required set of listening sockets has changed.

proposals/001-ClusterRepresentation.md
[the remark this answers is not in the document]
SamBarker: Yeah ok. Lets leave them separated.

SamBarker:
>One possible alternative would be to allocate endpoints automatically, based on the information about Nodes which propagates to clients via MetadataResponse, DescribeClusterResponse and FindCoordinatorResponse.

I think this is pointing, however much better expressed, to the same place as my suggestion of providing Kroxylicious with a set of credentials and a bootstrap address for the physical cluster and using that to generate the physical broker topology. I know we have said dealing with subsets of brokers in the Multi tenant case is out of scope but would this change make that substantially harder? I'm not sure I 

Removing the definition of `virtual brokers` from the configuration model also implies that Kroxylicious becomes the arbiter of the broker hostname, the domain comes from the `virtual cluster` definition. I guess we could add some sort of plug-able name generation scheme.

showuon (commented):
Interesting proposal! 
question:
1. It looks like the physical clusters are pre-defined, that is, we have 3 clusters, and each have 3 brokers. So virtual clusters will map to one of the 3 clusters. Is that right? 
2. Why can't we have a large cluster with N brokers. And the virtual cluster is just mapping to some of the brokers from the large cluster? That should be reduce the resource since all brokers share with the same controller.
3. with <2> above, when we found 1 broker is under heavy load, we can easily move some replicas to another broker, and then change the mapping in virtual cluster. And also we can share the ACLs for the same user.

Thoughts?

k-wall:
It'll be cleaner if I start I new PR.

PR #51, opened 2025-02-03
Operator API - A Home Grown alternative

k-wall:
I'm not sure that '[Gateway API flavoured](#52 )' approach for modelling ingress is the right one for Kroxylicious Operator.
I think that our ingress API should work with Gateway APIs (and OpenShift Ingress), rather than be one itself.  So, this 
has prompted me to think about an alternative API design.

In this design we have a `Proxy` and `ProxyIngress` (responsibility of the infra admin), and a `VirtualCluster` and the `FIlter` (responsibility of the application developer).

### DIFF
--- a/proposals/kroxylicious-operator-api-alternative.md
+++ b/proposals/kroxylicious-operator-api-alternative.md
@@ -1,0 +1,692 @@
+# Kroxylicious Operator API - Alternative Proposal.
+
+<!-- TOC -->
+* [Kroxylicious Operator API - Alternative Proposal.](#kroxylicious-operator-api---alternative-proposal)
+* [Background](#background)
+* [CRDs](#crds)
+* [Personas / Responsibilities](#personas--responsibilities)
+  * [Infrastructure admin](#infrastructure-admin)
+  * [Developer](#developer)
+* [API](#api)
+  * [Proxy](#proxy)
+  * [ProxyIngress](#proxyingress)
+  * [VirtualCluster](#virtualcluster)
+  * [Filter](#filter)
+* [Worked examples](#worked-examples)
+  * [On Cluster Traffic - plain downstream & upstream](#on-cluster-traffic---plain-downstream--upstream)
+  * [On Cluster Traffic - tls downstream & upstream](#on-cluster-traffic---tls-downstream--upstream)
+    * [On Cluster Traffic - tls downstream & upstream - varation using OpenShift Cluster CA generated cert](#on-cluster-traffic---tls-downstream--upstream---varation-using-openshift-cluster-ca-generated-cert)
+  * [Off Cluster Traffic (OpenShift Route)](#off-cluster-traffic-openshift-route)
+  * [Off Cluster Traffic (Load Balancer)](#off-cluster-traffic-load-balancer)
+  * [Upstream specified by Kafka CR](#upstream-specified-by-kafka-cr)
+* [Phased High Level Implementation Plan](#phased-high-level-implementation-plan)
+* [Not in Scope](#not-in-scope)
+<!-- TOC -->
+# Background
+
+Keith isn't sure that '[Gateway API flavoured approach](https://github.com/kroxylicious/design/pull/52)' for modelling
+ingress is the right one for Kroxylicious Operator. He thinks that our ingress API should work with Gateway APIs, rather than be one itself.  So, this 
+has prompted him to think about an alternative API design.
+
+# CRDs
+
+* Proxy CR - an instance of the Kroxylicious
+* ProxyIngress CR - Defines a way to access a Proxy
+* VirtualCluster CR - a virtual cluster
+* Filter CR - a filter definition
+
+![image](https://github.com/user-attachments/assets/8767da6b-3fc4-4fbc-99bc-64e93c06bc70)
+
+https://excalidraw.com/#json=CVAZMl-MuN1QP2cIW3hcA,cEWi5VjUg1mK2N1lvU-_xQ
+
+
+# Personas / Responsibilities
+
+## Infrastructure admin
+
+Responsible for the proxy tier and how proxies are exposed to the network.
+
+## Developer
+
+Responsible for the configuration of the virtualcluster and filters.
+Responsible for providing the virtualcluster TLS key/trust material (in Secret and ConfigMap respectively).
+Responsible for providing any key/trust material required to connect to the target cluster (in Secret and ConfigMap respectively).
+
+# API
+
+## Proxy
+
+The Proxy CR is the responsibility of the Infrastructure admin.
+
+Declares an instance of the proxy and defines the ingress mechanisms available to it.
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: Proxy
+metadata:
+  name: myproxy
+spec:
+
+   # Optional - controls the Proxy infrastructure.
+   infrastructure:
+      # Influences the annotations and labels on all resources created by the operator.
+      annotations:
+         x: y
+      labels:
+         x: y
+      proxy:
+        # Controls specific to operand deployment
+        replicas: 2
+        resources:
+          requests:
+            memory: "64Mi"
+            cpu: "250m"
+#  # Optional: default ordered list of filters to be used by a virtualcluster that doesn't apply it own.  (this doesn't fit well with the Infra role - drop it?).
+#  defaultFilterRefs:
+#  - group: filter.kroxylicious.io
+#    kind: Filter
+#    name: encryption```
+
+status:
+   # status of the proxy tbd
+   conditions:
+   - status: x
+     type: y
+     reason: r
+   # per listener status 
+   ingresses:
+   - name: myclusterip
+     conditions:
+     - ...
+```
+
+## ProxyIngress
+
+The ProxyIngress CR is the responsibility of the Infrastructure admin.
+
+It declares a named ingress mechanism - in other words, a way for traffic to get to a proxy.  ProxyIngress knows about four types:
+
+* clusterIP - for on-cluster traffic realised using Kubernetes ClusterIP Service. Supports TCP or TLS.
+* loadBalancer - for off-cluster realised using a Kubernetes LoadBalancer Service.  Supports TLS only.
+* openShiftRoute - for off-cluster realised using OpenShift Routes (OpenShift specific)
+* gateway - for off-cluster realised using Gateway API.  Supports `TLSRoutes` with listener type of `passthrough`.
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: ProxyIngress
+metadata:
+  name: myclusterip
+spec:
+  # reference to the proxy that will use this ingress.
+  proxyRef:
+    kind: Proxy  # if present must be Gateway, otherwise defaulted
+    group: proxy.kroxylicious.io # if present must be proxy.kroxylicious, otherwise defaulted=
+    name: myproxy  # name of proxy
+
+  # oneOf: clusterIP, loadBalancer, openShiftRoute, gateway
+  clusterIP:
+    protocol: TCP|TLS (ClusterIP supports both)
+    port: # optional, defaults to 9082
+  loadBalancer:
+    bootstrapAddressPattern: $(vitrtualClusterName).kafka.com 
+    brokerAddressPattern: broker-$(nodeId).$(vitrtualClusterName).kafka.com
+    port: # optional, defaults to 9082
+  openShiftRoute:
+    # ingress controller name (optional, if omitted defaults to default in the openshift-ingress-operator namespace).
+    resourceRef:
+      kind: IngressController  # if present must be IngressController, otherwise defaulted to IngressController
+      group: operator.openshift.io # if present must be operator.openshift.io, otherwise defaulted to operator.openshift.io
+      name: myingresscontroller
+      namespace: openshift-ingress-operator # namespace of the gateway, if omitted assumes namespace of this resource
+  gateway:
+    type: TLSRoute
+    resourceRef:
+      kind: Gateway  # if present must be Gateway, otherwise defaulted to Gateway
+      group: gateway.networking.k8s.io # if present must be gateway.networking.k8s.io, otherwise defaulted to gateway.networking.k8s.io
+      name: foo-gateway  # name of gateway resource to be used
+      namespace: namespace # namespace of the gateway, if omitted assumes namespace of this resource
+      sectionName: mytlspassthrough # refers to a TLS passthrough listener defined in the Gateway
+
+  # Optional
+  infrastructure:
+    # Controls the Admin influence the annotations/labels created on the ingress objects (Service, OpenShift Routes, TLSRoutes etc).
+    # This lets the Infrastructure admin  specify things like:
+    # https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#lb-type
+    annotations:
+       x: y
+    labels:
+       x: y
+status:
+   # describes the validity state of the ingress.  For instance:
+   # - for openShiftRoute the operator will verify that the ingress controller exists.
+   # - for gateway, it will verify that the gateway exists and the listener is defined as TLS passthrough.
+   conditions:
+   - status: x
+     type: y
+     reason: r 
+```      
+
+## VirtualCluster
+
+The VirtualCluster CR is the responsibility of the Developer.
+
+Declares a virtualcluster.
+
+A Virtualcluster is associated with exactly one Proxy.
+
+It enumerates the ingresses that are to be used by the virtualcluster.  It also supplies the virtual cluster ingress specific information
+such as the TLS certificates.
+
+The virtualcluster has a reference to a single target cluster which may be expressed using either a reference to a Strimzi Kafka object, or generic bootstraping information.
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: VirtualCluster
+metadata:
+  name: mycluster
+spec:
+  proxyRef:
+    kind: Proxy  # if present must be Proxy, otherwise defaulted
+    group: proxy.kroxylicious.io # if present must be proxy.kroxylicious.io, otherwise defaulted
+    name: myproxy
+ 
+  # list of ingresses the virtual cluster wishes to use.  Each ingress selected must be one associated with
+  # the proxy.  It is legal for the virtualcluster to choose only a subset.
+  ingresses:
+  - name: myclusterip
+    tls:
+      # server certificate - grab bag matched by hostname (what about key/keystore password?)
+      # secrets provided by the Developer.
+      certificateRefs:
+      - kind: Secret # if present must be Secret, otherwise defaulted to Secret
+        group: ""  # if present must be "", otherwise defaulted to ""
+        name: servercert
+        namespace: # namespace of the secret, if omitted assumes namespace of this resource
+      # peer trust
+      # configMap provided by the Developer
+      trustAnchorRefs:
+      - kind: ConfigMap # if present must be ConfigMap, otherwise defaulted to ConfigMap
+        group: ""  # if present must be "", otherwise defaulted to ""
+        name: trustbundle
+        namespace: # namespace of the configmap, if omitted assumes namespace of this resource
+      trustOptions:
+       clientAuth: REQUIRED
+    infrastructure:
+      # Controls that let the Developer influence the annotations/labels created on the ingress objects (Service, OpenShift Routes, TLSRoutes etc).
+      # If an annotation or label supplied here duplicates one provided by the ProxyIngress, the one defined by the ProxyIngress takes precendence. 
+      # The specific use case for this is https://docs.openshift.com/container-platform/4.17/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate
+      # where the developer wishes their VC to use Certificates created by OpenShift (useful for ClusterIP type TLS).
+      annotations:
+         x: y
+      labels:
+         x: y
+
+  # Points to the cluster being proxied.  Can either be Strimzi Kafka resource or endpoint details.
+  targetCluster:
+    # one of: resourceRef or bootstrapping
+    resourceRef:
+      kind: Kafka  # must be Kafka
+      group: strimzi.io # must be strimzi.io
+      name: my-cluster
+      listenerName: listener # name of strimzi listener 
+    bootstrapping:
+       bootstrap: bootstrap:9092
+       protocol: TCP|TLS
+       nodeIdRanges:
+       - name: mybrokers
+         range:
+          startInclusive: 0
+          endExclusive: 3
+    tls:
+      # Optional - client auth
+      # secret provided by the Developer.
+      certificateRef:
+        kind: Secret # if present must be Secret, otherwise defaulted to Secret
+        group: ""  # if present must be "", otherwise defaulted to ""
+        name: servercert
+        namespace: # namespace of the secret, if omitted assumes namespace of this resource
+      # Optional - peer trust
+      # configMap provided by the Developer.
+      trustAnchorRefs:
+      - kind: ConfigMap # if present must be ConfigMap, otherwise defaulted to ConfigMap
+        group: ""  # if present must be "", otherwise defaulted to ""
+        name: trustbundle
+        namespace: # namespace of the configmap, if omitted assumes namespace of this resource
+
+  # ordered list of filters to be used by the virtualcluster
+  filterRefs:
+  - group: filter.kroxylicious.io
+    kind: Filter
+    name: encryption
+status:
+   # overall status
+   conditions:
+   - status: x
+     type: y
+     reason: r 
+   # per ingress status 
+   ingresses:
+   - name: myclusterip
+     bootstrap: xxx.local:9082   # the bootstrap address
+     protocol: TCP|TLS
+     conditions:
+     - ...
+   # per filter status 
+   filters:
+   - name: myclusterip
+     conditions:
+     - ...
+```
+
+## Filter
+
+The VirtualCluster CR is the responsibility of the Developer - As per current implementation.
+
+```yaml
+apiVersion: filter.kroxylicious.io/v1alpha1
+kind: Filter
+metadata:
+  name: myfilter
+spec:
+  type: io.kroxylicious.filter.encryption.RecordEncryption
+  config:
+    kms: Foo
+    kmsConfig: {}
+    selector: Bar
+    selectorConfig: {}
+```
+
+# Worked examples
+
+## On Cluster Traffic - plain downstream & upstream
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: Proxy
+metadata:
+  name: myproxy
+spec: {}
+```
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: ProxyIngress
+metadata:
+  name: myclusterip
+spec:
+  proxyRef:
+    name: myproxy
+  clusterIP:
+    protocol: TCP
+    port: 9082
+```
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: VirtualCluster
+metadata:
+  name: mycluster
+spec:
+  proxyRef:
+    name: myproxy
+    kind: Proxy
+    group: proxy.kroxylicious.io
+
+  ingress:
+  - name: myclusterip
+
+  targetCluster:
+    bootstrapping:
+       bootstrap: upstream:9092
+       protocol: TCP
+       nodeIdRanges:
+       - name: mybrokers
+         range:
+          startInclusive: 0
+          endExclusive: 3
+  filterRefs:
+  - group: filter.kroxylicious.io
+    kind: Filter
+    name: encryption 
+```
+
+What would operator create:
+* kroxylicious deployment, 1 replica
+* proxy relies on port based routing (needs new PortPerNode scheme that allows the Operator to control the ports precisely.  This is needed to prevent the potential for "crossed-lines" during reconcilations.)
+* 1 ClusterIP services (with n+1 ports) - operator needs to _deterministically_ assign a block of ports for this VC and create the port/target mapping in the Service accordingly.
+* Kafka Clients connect to serviceaddress:9082 
+
+## On Cluster Traffic - tls downstream & upstream
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: Proxy
+metadata:
+  name: myproxy
+spec: {}
+```
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: ProxyIngress
+metadata:
+  name: oncluster
+spec:
+  proxyRef:
+    name: myproxy
+
+  clusterIP:
+    protocol: TLS
+    port: 9082
+```
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: VirtualCluster
+metadata:
+  name: mycluster
+spec:
+  proxyRef:
+    name: myproxy
+    kind: Proxy
+    group: proxy.kroxylicious.io
+
+  ingress:
+  - name: myclusterip
+    tls:
+      certificateRef:
+         ...
+
+  targetCluster:
+    bootstrapping:
+       bootstrap: upstream:9092
+       protocol: TLS
+       nodeIdRanges:
+       - name: mybrokers
+         range:
+          startInclusive: 0
+          endExclusive: 3
+
+  filterRefs:
+  - group: filter.kroxylicious.io
+    kind: Filter
+    name: encryption 
+```
+
+What the Developer would provide:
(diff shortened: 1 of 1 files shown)


proposals/operator-ingress-sketch.md
```diff
@@ -0,0 +1,189 @@
...
+    type: ClusterIP|LoadBalancer
+    brokerAddressPattern: $(node)-${vitrtualClusterName}.example.com  (LoadBalancer only)
+    protocol: TCP|TLS (ClusterIP supports both, LoadBalancer TLS only)
+    port: # optional, defaults to 9082
+  openShiftRoute:
+    ingressControllName: # ingress controller name (optional)
+  gateway:
+    type: TLSRoute
+    # think about this
+```
```
tombentley, line 67: No status?

proposals/operator-ingress-sketch.md
```diff
@@ -0,0 +1,189 @@
...
+```yaml
+apiVersion: proxy.kroxylicious.io/v1
+kind: VirtualCluster
+metadata:
+  name: mycluster
+spec:
+  proxyReference:
+    kind: ProxyIngress
+    group: proxy.kroxylicious.io
+    name: myproxy
```
tombentley, line 82: Does this point to the Proxy, or the ProxyIngress (the `name` and the `type` disagree with the above yamls, so it's not clear).

proposals/operator-ingress-sketch.md
```diff
@@ -0,0 +1,189 @@
...
+  # ingress declared at by the proxy.  If the proxy is using the default ingress, that is what the virtualcluster will
+  # use too.
+  ingresses:
+  - name: myclusterip
+
+  # Points to the cluster being proxied.  Can either be a external service or a Strimzi Kafka resource.
+  targetClusterRef:
+    kind: Kafka|Service
+    group: strimzi.io|<empty>
+    name: name
```
tombentley, line 94: Why does this need to reference another Kube resource?

proposals/operator-ingress-sketch.md
```diff
@@ -0,0 +1,189 @@
...
+* Infrastructure admin - responsible for the proxy tier and how it exposes itself to the network.
+* Developer - responsible for declaraction of the virtualcluster
+
+## API
+
+### Proxy
+
+Owned by the Infrastructure admin.
+
+Declares an instance of the proxy.  It is valid for this to be an empty object.  The empty object
```
tombentley, line 18: > It is valid for this to be an empty object.

I'm not sure this is really something worth striving for. The more you rely on defaulting the more cryptic things look to someone who doesn't know the API. While a single replica is a fairly guessable default I'm not so sure that an uninformed reader would guess "If no ingresses are defined, the operator will behave as if a plain clusterip ingress is defined called default."

proposals/operator-ingress-sketch.md
```diff
@@ -0,0 +1,189 @@
...
+  targetClusterRef:
+    name: my-cluster
+    group: strimzi.io|<empty>
+    kind: strimzi
+    listener: my-strimzi-listener
+```
+
+What would operator create:
+* kroxylicious deployment, 1 replica
+* n+1 ClusterIP services (1 per broker + 1 bootstrap).  It would watch the strimzi resource in order to know n
```
tombentley, line 144: > It would watch the strimzi resource in order to know n

If we're going to go this route then ading this functionality to the proxy is something we can progress in parallel to bikeshedding on this API.

proposals/operator-ingress-sketch.md
```diff
@@ -0,0 +1,189 @@
...
+  name: myproxy
+spec:
+  ingresses:
+  - name: routes
+    openShiftRoute: {}
+status:
+  ingresses:
+    - name: default
+      openShiftRoute:
+        ingressControllName: default
```
tombentley, line 162:
```suggestion
        ingressControllerName: default
```

proposals/operator-ingress-sketch.md
k-wall: They'll be a status section yes.

proposals/operator-ingress-sketch.md
k-wall: Let me explain my motivation. I want to be able to stand up for a demo and say "let me show you how you add encryption at rest to your kafka" and then proceed to do it with as few lines of yaml as possible - something that people can hold in their heads.  I figured that a proxy using plain protocol, exposed by services for on-cluster traffic, it is the simplest useful thing. I've tried to make that the easy thing whilst supporting more complex ingress patterns.

Maybe what I am suggesting here is not the right realisation of the idea.   I'm happy to keep iterating to get to something less surprising.

proposals/operator-ingress-sketch.md
k-wall: You mean have the proxy probe the target cluster and discover the broker topology and somehow have the operator discover that from the proxy? REST API?  I don't think that's completely straightforward - I see a cold start issue.

1. Cluster gets started (3 brokers + bootstrap route)
2. Proxy gets started.  Proxy doesn't know how many brokers cluster has.
3. Operator establishes proxy bootstrap route.
4. Client connects through proxy to bootstrap, proxy forwards connection.
5. Proxy learns cluster has 3 brokers.
6. Operator learns (polling?) the Proxy that the Cluster has three brokers.  
7. Operator establishes broker routes.

Until we've reached step 7, we'll see connection failures.

proposals/operator-ingress-sketch.md
k-wall: That was a typo.  The VirtualCluster refers to the Proxy.

proposals/operator-ingress-sketch.md
k-wall: I stepped away from this direction.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,638 @@
...
+ # Optional
+ infrastructure:
+    # Controls the Admin influence the annotations/labels created on the ingress objects (Service, OpenShift Routes, TLSRoutes etc).
+    # This lets the Infrastructure admin  specify things like:
+    # https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/#lb-type
+    annotations:
+       x: y
+    labels:
+       x: y
+status:
```
k-wall, line 128: What about TLS cipher suite and protocol controls?  Where do they live?

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,650 @@
...
+     type: y
+     reason: r
+   # per listener status 
+   ingresses:
+   - name: myclusterip
+     condidtions:
+     - ...
+```
+
+### ProxyIngress
```
k-wall, line 72: I'm tempted to call this a ProxyListener.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,652 @@
...
+          endExclusive: 3
+  filterRef:
+  - group: filter.kroxylicious.io
+    kind: Filter
+    name: encryption 
+```
+
+What would operator create:
+* kroxylicious deployment, 1 replica
+* proxy relies on port based routing (needs new PortPerNode scheme that allows the Operator to control the ports precisely.  This is needed to prevent the potential for "crossed-lines" during reconcilations.)
```
k-wall, line 320: Why don't the existing PortPerBroker schemes work?   The issue is during a reconciliation where the number of nodes in target cluster has changed, some proxy replicas will be still using previous port assignments, whereas some will be using the new.  That risks that the client might connect to the wrong broker (or even wrong cluster).

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,652 @@
...
+
+  filterRef:
+  - group: filter.kroxylicious.io
+    kind: Filter
+    name: encryption 
+```
+
+What would operator create:
+* kroxylicious deployment, 1 replica
+* proxy relies on port based routing  (needs new PortPerNode scheme that allows the Operator to control the ports precisely.  This is needed to prevent the potential for "crossed-lines" during reconcilations.)
```
k-wall, line 383: This could use SNI based routing, but I figure since we've had to use port based routing for TCP, we may as well use port based routing for TLS too.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+      name: myingresscontroller
+      namespace: openshift-ingress-operator # namespace of the gateway, if omitted assumes namespace of this resource
+  gateway:
+    type: TLSRoute
+    resourceRef:
+      kind: Gateway  # if present must be Gateway, otherwise defaulted to Gateway
+      group: gateway.networking.k8s.io # if present must be gateway.networking.k8s.io, otherwise defaulted to gateway.networking.k8s.io
+      name: foo-gateway  # name of gateway resource to be used
+      namespace: namespace # namespace of the gateway, if omitted assumes namespace of this resource
+      sectionName: mytlspassthrough # refers to a TLS passthrough listener defined in the Gateway
```
tombentley, line 117: Can you elaborate for each of these mechanisms what resources are expected to already exist/are the responsibility of the user, vs what resources will be managed by the operator?

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+
+  # Points to the cluster being proxied.  Can either be Strimzi Kafka resource or endpoint details.
+  targetCluster:
+    # one of: resourceRef or resource
+    resourceRef:
+      kind: Kafka  # must be Kafka
+      group: strimzi.io # must be strimzi.io
+      name: my-cluster
+      listenerName: listener # name of strimzi listener 
+    resource:
```
tombentley, line 199: `resource` seems like a bit of an odd name to me, because there is no `resource` in the k8s sense.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+       bootstrap: bootstrap:9092
+       protocol: TCP|TLS
+       nodeIdRanges:
+      - name: mybrokers
+        range:
+          startInclusive: 0
+          endExclusive: 3
+    tls:
```
tombentley, line 207: I wonder whether calling both the `targetCluster` and the `ingress` property `tls` is the right choice. (I know that's what we did for the proxy config itself, but I have by doubts there too). It means you have to look to see what the container object is to know what you're configuring. I can imagine people finding that confusing.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+    resourceRef:
+      kind: Kafka  # must be Kafka
+      group: strimzi.io # must be strimzi.io
+      name: my-cluster
+      listenerName: listener # name of strimzi listener 
+    resource:
+       bootstrap: bootstrap:9092
+       protocol: TCP|TLS
+       nodeIdRanges:
+      - name: mybrokers
```
tombentley, line 203: Is there a reason we need to name this?

proposals/kroxylicious-operator-api-alternative.md
robobario: Do you mean that the port assignment would be stateful? That the operator would consider the previous deployed state to ensure new nodes receive a previously un-allocated port? (or at least track the last-allocated-port and keep rolling forward)

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+    name: encryption 
+```
+
+What would operator create:
+* kroxylicious deployment, 1 replica
+* proxy relies on SNI based routing (use SNI scheme)
+* ClusterIP service called `myopenshiftroute-service` (shared)
+* n+1 OpenShift Routes configured for pass-through. The `to` points to the `myopenshiftroute-service`.
+
+Note: that the operator would write the virtualcluster proxy config based on the hostnames assigned to the Routes by the Ingress Controller.
```
robobario, line 486: We need the openshift routes to match the virtual cluster broker address pattern. Would it be the Operator's responsibility to set up the Route hosts, or should the users configure something like `broker-${nodeId}.abc.def.${ingressDomain}`?

proposals/kroxylicious-operator-api-alternative.md
SamBarker: For me the main motivation is clarifying that your configuring TLS in each of these places and you have exactly the same set of TLS options in each. There is nothing more annoying that failing to configure property X in context B when it works exactly as you want in Context A.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+### VirtualCluster
+
+The VirtualCluster CR is the responsibilty of the the Developer.
+
+Declares a virtualcluster.
+
+A Virtualcluster is associated with exactly one Proxy.
+
+It enumerate the ingresses that are to be used by the virtualcluster.  It also supplies the virtual cluster ingress specific information
+such as the TLS certificates.
```
robobario, line 147: TLS config and target cluster location feels more like an `Infrastructure admin` concern

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+* gateway - for off-cluster realised using Gateway API.  Supports `TLSRoutes` with listener type of `passthrough`.
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1
+kind: ProxyIngress
+metadata:
+  name: myclusterip
+spec:
+  # reference to the proxy that will use this ingress.
+  proxyRef:
```
robobario, line 90: If ProxyIngress is associated with one Proxy and they're both infrastructure-admin oriented why not combine them in the one CR?

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+    kind: Proxy  # if present must be Gateway, otherwise defaulted
+    group: proxy.kroxylicious.io # if present must be proxy.kroxylicious, otherwise defaulted=
+    name: myproxy  # name of proxy
+
+  # oneOf: clusterIP, loadBalancer, openShiftRoute, gateway
+  clusterIP:
+    protocol: TCP|TLS (ClusterIP supports both)
+    port: # optional, defaults to 9082
+  loadBalancer:
+    bootstrapAddressPattern: $(vitrtualClusterName).kafka.com
```
robobario, line 100: So `virtualClusterName` is an Operator only concept that lets one Ingress be used for multiple VirtualClusters?

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+        namespace: # namespace of the secret, if omitted assumes namespace of this resource
+      # Optional - peer trust
+      trustAnchorRefs:
+       - kind: ConfigMap # if present must be ConfigMap, otherwise defaulted to ConfigMap
+         group: ""  # if present must be "", otherwise defaulted to ""
+         name: servercert
+         namespace: # namespace of the configmap, if omitted assumes namespace of this resource
+
+    # ordered list of filters to be used by the virtualcluster
+    filterRefs:
```
SamBarker, line 222: I could have screwed up but I see the filterRefs as being scoped to the target cluster and I think they should be scoped to the virtual cluster. 

I'd also argue that the property should be `filterChain` to make the ordering concerns clear and obvious.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+kind: Filter
+metadata:
+  name: myfilter
+spec:
+  type: io.kroxylicious.filter.encryption.RecordEncryption
+  config:
+    kms: Foo
+    kmsConfig: {}
+    selector: Bar
+    selectorConfig: {}
```
SamBarker, line 261: What is your thinking here around the schema for `Kind: filter`? Are all filters going to have standard  generic schema? e.g. `type` + arbitrary config? 

Or should we be offering a higher level CR for Filters shipped by the project? 

I started wondering about this from the angle of what happens if we want to have multiple filter instances pointed at the same KMS possibly even a corporate standard KMS instance that the infrastructure persona manages.  Should the filter also support a `kmsRef` or just a `configRef`? that points to things configured once?

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,638 @@
...
+     # peer trust
+     trustAnchorRefs:
+      - kind: ConfigMap # if present must be ConfigMap, otherwise defaulted to ConfigMap
+        group: ""  # if present must be "", otherwise defaulted to ""
+        name: servercert
+        namespace: # namespace of the configmap, if omitted assumes namespace of this resource
+      trustOptions:
+       clientAuth: REQUIRED
+    infrastructure:
+      # Controls that let the Developer influence the annotations/labels created on the ingress objects (Service, OpenShift Routes, TLSRoutes etc).
```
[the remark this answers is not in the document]
robobario: Might need more of an idea how we would use serving certs. My read of it is we could potentially use them for in-cluster TLS and maybe in combination with openshift routes using [TLS re-encryption](https://docs.openshift.com/container-platform/4.17/networking/routes/secured-routes.html#nw-ingress-creating-a-reencrypt-route-with-a-custom-certificate_secured-routes)?

So if we have N brokers, we would have N Services all annotated for serving certs. This would generate a PEM/cert for each of the N services. We configure the proxy to load all the generated PEMs into that virtual cluster (some work required to take multiple PEM files and construct a single KeyStore).

So I think it's pretty painful, you need to be able to target a secret name to each of the services maybe using a pattern. Might be better to have the Operator handle this for you with another Ingress strategy and have the operator look after secret names for the user. The end result could be pretty cool though, secure OpenShift Routes for off-cluster without having to set up cert-manager, and secure in-cluster access.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+## API
+
+### Proxy
+
+The Proxy CR is the responsibilty of the the Infrastructure admin.
+
+Declares an instance of the proxy and defines the ingress mechanisms available to it.
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1
```
SamBarker, line 33: I appreciate  this is just a sketch but we should be starting with an `v1alpha1` version designator

proposals/kroxylicious-operator-api-alternative.md
SamBarker: I certainly think Listener is better than ingress, though does `listener` clash with Kafka listeners (for our user base)? Maybe `proxyAccess`

proposals/kroxylicious-operator-api-alternative.md
SamBarker: Does life get easier if we just mandate SNI (for day 1) and then worry about stateful port assignment later?

SamBarker (commented):
Thanks @k-wall without digging into the details this feels like the use case persona driven API I'd been hoping to see.

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
k-wall: I don't think they have a role to play with OpenShift Routes in our use-case.   Kroxylicious need passthrough behaviour in order to get the SNI, so I don't think rencrypt is ever a mode we can actually make use of.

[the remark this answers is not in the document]
k-wall: > Might need more of an idea how we would use serving certs.

The use-case is for on-cluster services secured with TLS.

The OpenShift cluster operator takes the responsibility for automatically creating a certificate and periodically rolling it.   You gesture to the OpenShift cluster operator that you want a certificate with an annotation added to the service.  The cluster operator creates a secret containing the key-material.   Your operand needs to serve that certificate. The client use https://docs.openshift.com/container-platform/4.16/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate so that it trusts the cluster operator's CA.

> So if we have N brokers, we would have N Services all annotated for serving certs.

There's a choice there.

1. We could create n+1 services.  Each has the annotation.  The cluster operator create n+1 secrets.   We'd somehow mount those into Kroxylicious and give the SNI handler the smarts to serve the right certificate.   The client will be connecting through different services bootstrap-myvcluster, broker0-myvcluster, etc.. so we can then use SNI to route the traffic within the Proxy.

2. The alternative it is  create 1 service (with n+1 port mappings) and use port based routing.  That means there's only 1 service to annotate and one 1 certificate to mount.   The client now connects to myvcluster:2000 and gets back brokers myvcluster:2001..myvcluster:200n.

I think either way works.   In the proposal, I was suggesting the second.   I said we need a new PortPerBroker scheme in order to make sure the port/broker mapping says stable across cluster topology changes.

proposals/kroxylicious-operator-api-alternative.md
k-wall: Naming is so hard!

> I certainly think Listener is better than ingress, though does listener clash with Kafka listeners 

Isn't it the proxy's analogue for kafka's listener?   So maybe the clash is exactly what we want.

> Maybe proxyAccess

I think I prefer either Ingress or Listener over Access.  Access makes with think of ACLs.

proposals/kroxylicious-operator-api-alternative.md
k-wall: > Do you mean that the port assignment would be stateful? That the operator would consider the previous deployed state to ensure new nodes receive a previously un-allocated port? (or at least track the last-allocated-port and keep rolling forward)

Yes, exactly.  I'd initially thought that the operator would just use the PortPerBrokerClusterNetworkAddress scheme.     I then got thinking about the case where the broker topology of the target cluster changed (or even a new virtual cluster added).   If proxy is replicas > 1, I think all kinds of bad things happen depending on the relative ordering of the port assignments in the services being updated and the replicas being restarted.   I worried that :1234 which previously pointed at broker 3 now points to broker 4, or even a entirely different cluster.  How's a client with a connection to :1234 going to respond.  And which client are we talking about (java, sarama)?   That lead me to the thought that port/broker assignments need to be sticky.

So that got me thinking about a new  PortPerBroker schema that allows the Operator to precisely control the port/broker assignments.   It could then use the previous port/broker assignment state as input to the next state.

> Does life get easier if we just mandate SNI (for day 1) and then worry about stateful port assignment later?

That'd crossed my mind too.

We could just change the implementation roadmap - implement ClusterIP/TLS first (using SNI), then come back to ClusterIP/TCP later.

The disadvantages:

1) it makes https://docs.openshift.com/container-platform/4.16/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate slightly harder - more secrets to juggle (see the conversation above).

2) Being able to proxy a cluster without TLS sure makes a demo easy to follow.

proposals/kroxylicious-operator-api-alternative.md
k-wall: I have talked about that in worked examples.  In each case, I stated what resources the operator would create.

proposals/kroxylicious-operator-api-alternative.md
k-wall: I've renamed it `bootstrapping`.

proposals/kroxylicious-operator-api-alternative.md
k-wall: In the proxy config, I think you make a valid point.  

```yaml
  demo:
    tls:  <---   tls setting applying to the virtual cluster (TLS server role)
       x: ...
    targetCluster:
      bootstrapServers: localhost:9092
      tls: <---   tls setting applying to the upstream (TLS client role)
         x: ..
```

I'm not convinced we have same issue in this proposal as the two tls nodes appear in separate branches of the tree.  I think there is much less potential for confusion.

```yaml
ingresses:
   name: foo
   tls:  <---   tls setting applying to the virtual cluster (TLS server role)
   
targetCluster:
   tls: <---   tls setting applying to the upstream (TLS client role)
```      
     
we could rename them ingressTls: and targetTls:   but is that really the kube way?  

WDYAT?

proposals/kroxylicious-operator-api-alternative.md
k-wall: I think we borrowed the idea from the Strimzi model when RangeAwarePortPerNodeCluster scheme was implemented. 

Naming the range isn't required for the function.  We could remove it.

proposals/kroxylicious-operator-api-alternative.md
k-wall: > We need the openshift routes to match the virtual cluster broker address pattern.

I don't think this is something we'd surface to users.

I was imagining it would be the Operator's responsibility. It would create routes with names in a well-known form:

broker-${nodeId}-${virtualcluster}
bootstrap-${virtualcluster}

and the Ingress Operator would assign the hostnames.

broker-${nodeId}-${virtualcluster}-${namespace}.${ingressDomain}
bootstrap-${virtualcluster}-${namespace}.${ingressDomain}

We'd need to document how the name will be formed so that the user can generate their certificate.

There's a potential to support vanity names (by setting the Routes spec.host field), but I think we'd leave that for an RFE.

proposals/kroxylicious-operator-api-alternative.md
k-wall: We definitely could.. But there are counterarguments:

* For OpenShift and Gateway, the user might need to use more than one controller (you could have two operators implementing the  Gateway API on the cluster.  so many instances of ProxyIngress would be required.
* CRs seem to work best when they focus tightly on a single concern.  Big APIs get unwieldy. 
* Decoupling might give us more wiggle/innovation room.

proposals/kroxylicious-operator-api-alternative.md
k-wall: `$(vitrtualClusterName)` would be understood by Kroxylicious, rather than the Operator.  Both  `$(vitrtualClusterName)`  and `$(nodeId)` would be passed down as literals into the proxy config file.  `$(vitrtualClusterName)` would be substituted for the virtual cluster's name.  $(nodeId) is as-is.

btw.  I did wonder about pushing bootstrapAddressPattern and brokerAddressPattern into the VirtualCluster.  That would do away with the need introduce the new pattern.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,654 @@
...
+      # If an annotation or label supplied here duplicates one provided by the ProxyIngress, the one defined by the ProxyIngress takes precendence. 
+      # The specific use case for this is https://docs.openshift.com/container-platform/4.17/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate
+      # where the developer wishes their VC to use Certificates created by OpenShift (useful for ClusterIP type TLS).
+      annotations:
+         x: y
+      labels:
+         x: y
+
+  # Points to the cluster being proxied.  Can either be Strimzi Kafka resource or endpoint details.
+  targetCluster:
```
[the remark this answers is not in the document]
k-wall: > I'm wondering if TargetCluster should be a CR as well less from a technical point of view but more of a who manages it.

Yeah,  I wondered about making `TargetClusterRef` its own CR too.    I went through an iteration where I had a `TargetClusterRef` CR.

```yaml
targetCluster:
  clusterRef:   <-- points at the cluster being proxied.
      kind: Kafka|TargetClusterRef
      group: kafka.strimzi.io|proxy.kroxylicious.io
      name: mycluster
   tls:
      clientCert:...
```

```yaml
kind:  TargetClusterRef
spec:
   bootstrap: xxx:1234
   protocol: TLS
   nodeIdRange: ....
   tls:
      trustAnchorsRef:...
```

The nice thing is that `ClusterRef` resources could be shared by many VirtualClusters, which seems quite nice for Multi Tenancy, for instance.

I'm not sure it is really clean cut who would be responsible for the `ClusterRef` object.   There are times when it seems Developer, others when it Infra Admin.

proposals/kroxylicious-operator-api-alternative.md
k-wall: > I could have screwed up but I see the filterRefs as being scoped to the target cluster and I think they should be scoped to the virtual cluster.

> I'd also argue that the property should be filterChain to make the ordering concerns clear and obvious.

I've used `*Refs` across the API, borrowing the terminology from `GatewayAPI`.   So that gives me `filterRefs`.   `filterChainRefs` wouldn't really make sense.  You'd expect to be listing FilterChain CRs.

proposals/kroxylicious-operator-api-alternative.md
k-wall: I haven't thought about this aspect at all this proposal!    The Filter definition you see is what we have on main right now.

In https://github.com/kroxylicious/design/blob/playing-with-operator-ingress-ideas/proposals/kroxylicious-operator-api-alternative.md#phased-high-level-implementation-plan I've said 'allow filters to accept secrets'.  Personally I feel that @robobario ["secret notation"](https://github.com/kroxylicious/kroxylicious/discussions/1613#discussioncomment-11328768) is good enough.    I don't think we want to start thinking about making KMSs toplevel right now.

proposals/kroxylicious-operator-api-alternative.md
robobario: Yep, I can see that trying to separate out frontend and backend tls config would end up with an unwieldy amount of CRs required

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
SamBarker: >I'm not sure it is really clean cut who would be responsible for the ClusterRef object. There are times when it seems Developer, others when it Infra Admin.

When would it be a developer responsibility? Isn't it always infra (even if the infra person is the same developer waring a different hat?)

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
robobario: another kink, the `tls.key` generated by the serving cert feature is in PKCS1 format. Netty would need bouncycastle to handle it.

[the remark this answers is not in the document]
robobario: Also I think you're right, it's hard to find it documented but I think re-encrypt routes expect HTTPS and can't handle TCP.  So I think `passthrough` is the only termination type that's relevant to us.

proposals/kroxylicious-operator-api-alternative.md
SamBarker: Hmm consistency vs domain specific naming. Tough one. I can certainly live with `filterRefs`, though maybe we should think about a `filterChain` CR as well, but I think `filterRefs` is the right level for MVP.

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
k-wall: Or does JDK now support PKCS1??   https://bugs.openjdk.org/browse/JDK-8023980

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
k-wall: I was thinking about the case where the Kafka Cluster being proxied was entirely the responsibility of the Developer team.  So a use-case where a app dev team is standing up a Kafka service for other in their organisation we say, schema enforcement.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,692 @@
...
+        resources:
+          requests:
+            memory: "64Mi"
+            cpu: "250m"
+#  # Optional: default ordered list of filters to be used by a virtualcluster that doesn't apply it own.  (this doesn't fit well with the Infra role - drop it?).
+#  defaultFilterRefs:
+#  - group: filter.kroxylicious.io
+#    kind: Filter
+#    name: encryption```
+
```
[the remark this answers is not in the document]
k-wall: I think the management port is an internal concern, so doesn't need to be surfaced in the API.
I think all we need is a switch to turn metrics on or off.   It should defect to on.
I don't see any reason to make the health probes configurable at this stage.

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
k-wall: I experimented with a TargetClusterRef in #53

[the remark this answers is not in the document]
SamBarker: > I was thinking about the case where the Kafka Cluster being proxied was entirely the responsibility of the Developer team. So a use-case where a app dev team is standing up a Kafka service for other in their organisation we say, schema enforcement.

I think we are on the same page then, but to me in that case the app dev team are waring the infra hat as well as the dev hat, and the proxied cluster is still an infra concern.

proposals/kroxylicious-operator-api-alternative.md
k-wall: I've fixed this in #53

SamBarker (approved):
Thanks for putting this together @k-wall it really helped clarify my/our thinking

PR #52, opened 2025-02-17
Draft: Operator proposal - align closely with Gateway API

robobario:


SamBarker (commented):
Thanks for all the hard work on this @tombentley & @robobario.

I was initially quite keen on the idea of aligning with GatewayAPI but reading this and comparing it with #51 I'm seeing a lot of compromises between what the GatewayAPI spec wants and what we want to offer. Thats both in terms of difficulty expressing things in a gateway API compliant way and in exposing Kafka the way we want to. 

What I'm not clear on at this stage is how many of the compromises outlined here actually carry over to #51 because you've done a lot more detailed work on this.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+```yaml
+# example with caCerts
+---
+kind: KafkaBackendTLSPolicy
+apiVersion: kroxylicious.io/v1alpha1
+metadata:
+  name: my-tls-policy
+spec:
+  targetRefs:
+    - kind: Service
```
k-wall, line 229: this assumes the backend is represented by a service.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+    - # A reference to the KafkaProxies this Route is a part of
+      group: kroxylicious.io # Must be kroxylicious.io
+      kind: KafkaGateway # Must be KafkaGateway
+      name: simple
+      sectionName: my-listener
+      brokers:
+        - advertisedBrokerIds: [ 1,2,3,4,5,6,7,8,9 ]
+          brokerIdentityMapping: identity | hash # to be added eventually, default is identity
+      hostnames:
+        # When a hostnamePattern ends with the cluster/namespace
```
k-wall, line 322: We are using a naming convention to let the user choose cluster ip or load balancer (or both).   How would we gesture that we want to use an alternative off-cluster ingress mechanism?  Say, a Gateway or an OpenShift Route?

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+```yaml
+metadata:
+  name: encryption
+  namespace: my-proxy
+spec:
+  type: io.kroxylicious.filter.encryption.RecordEncryption
+  config:
+    kms: Foo
+    kmsConfig:
+      myApiTokenFile: $(secretFile:secretName:subPath) # secret will be mounted into container and config replaced with path to secret
```
k-wall, line 385: I think this idea passes the 'good enough' test.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+tls:
+  key:
+    keyPairs:
+      - privateKeyFile: /opt/kroxylicious/server/key-material/tls.key
+        certificateFile: /opt/kroxylicious/server/key-material/tls.crt
+      - privateKeyFile: /opt/kroxylicious/server/key-material-2/tls.key
+        certificateFile: /opt/kroxylicious/server/key-material-2/tls.crt
+```
+
+Then the Proxy will be responsible for determining which certificate to use for the virtual cluster based on the SNI
```
k-wall, line 422: We need to remember to make sure the intermediate certificates too.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+      name: simple
+      sectionName: my-listener
+      brokers:
+        - advertisedBrokerIds: [ 1,2,3,4,5,6,7,8,9 ]
+          brokerIdentityMapping: identity | hash # to be added eventually, default is identity
+      hostnames:
+        # When a hostnamePattern ends with the cluster/namespace 
+        # domain we'd create ClusterIP Services automatically. Else
+        # we expose a single LoadBalancer service.
+        - foo-%.svc.ns.cluster.local
```
k-wall, line 325: I think the operator should publish the bootstrap addresses to the status section so a kubernetes based client app can read the bootstrap information for itself.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+- DNS twiddling is automated by the platform.
+
+Disadvantage:
+
+- It’s OpenShift only
+- We're proxying the proxy with HAProxy. More latency. More machinery.
+- Known issues with haproxy getting restarted every time a new Route is created = dropped Kafka connections. However
+  they seem to be fixing that (https://issues.redhat.com/browse/NE-879)!
+
+We could have a configurable list of ingress domains in KafkaGatewayParameters, then when we find a hostname that is a
```
k-wall, line 649: It is somewhat unfortunate to need to duplicate the ingressdomain.  If the ingressdomain changes for whatever, the configuration in the Kroxylicious CRs would be stale.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+subdomain of an ingress domain, the operator will create a single ClusterIP service and create passthrough routes
+following the same naming convention. So if the hostname in the CR is `my-ingress-route-%.<ingress-domain>` the Operator
+would create routes:
+
+`my-ingress-route-bootstrap.<ingress-domain>`
+`my-ingress-route-broker-0.<ingress-domain>`
+`my-ingress-route-broker-1.<ingress-domain>`
+`my-ingress-route-broker-2.<ingress-domain>`
+
+Then the user can bootstrap with `my-ingress-route-bootstrap.<ingress-domain>:443`
```
k-wall, line 659: That's slightly weird. The KafkaGateway's port is ignored in this case.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+            - kind: Service
+              namespace: kafka
+              name: my-cluster
+              port: 9093
+```
+
+Because the hostname is not suffixed with `svc.cluster.local` the operator will infer that we want to expose this
+hostname off-cluster.
+
+The operator will create a LoadBalancer Service named `my-route` mapping port 9092 to all the Proxy instances
```
k-wall, line 617: Sometimes you want to make hints to the loadbalancer to control its behaviour.  This is done via the Service's [annotations](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/). 

We've got the `infrastructure:` field at the Gateway, but that will apply to all the resources created by the operator.   AFAICS, we have no way to target an annotation at a particular resource.

proposals/001-operator-design.md
```diff
@@ -0,0 +1,818 @@
...
+        - name: my-rule
+          filters: [ ]
+          backendRefs:
+            - kind: Service
+              namespace: kafka
+              name: my-cluster
+              port: 9093
+```
+
+The Operator will manifest this as two virtual clusters in the proxy configuration, both with 9092 as the bootstrap
```
k-wall, line 718: I really want to fix this!

k-wall:
Thank you, this is a really through write up of this idea.

k-wall:
I too initially believed that modelling our API based on the Gateway API was the way to go.   I've changed my mind on that over the last few weeks.  Whilst I think the proposal is implementable and would function, I worry the result would difficult to use.

Thinking about my biggest areas of concern: 

1. I don't like using the suffix of the hostname field to decide which sort of ingress to create.  It works for `ClusterIP` and `LoadBalancer` but it does really extend to other forms of ingress.  For OpenShift Routes,  we'd need to hardcode router addresses (which might change).  I also have doubts about how it work would with a real Gateway API.  How could our operator know whether to follow its LoadBalancer or Gateway API path?  It'd wouldn't have sufficient information.
2. `KafkaGateway` has a port number, but in the case of OpenShift Routes or real  Gateway TLSRoute, we'd ignore it.   The port number is owned  by the Ingress Operator (443) or (real) Gateway Operator.
3. It seems common for ingress behaviours to controlled via labels or annotations.   Whilst we can apply labels via the infrastructure field in the KafkaGateway, but we've got no way to be specific and just apply a label to say, the ClusterIP's service.

proposals/001-operator-design.md
robobario: We could use an ExternalName service but it's a bit indirect yes

k-wall: https://kubernetes.io/docs/concepts/services-networking/service/#externalname doesn't include a port.  I think we'd need our reference object.

robobario:
Closing, we have decided #51 has fewer compromises and we will go in that direction

PR #53, opened 2025-02-18
Operator API - extract out a KafkaClusterRef - a reference to a cluster being proxied.

k-wall:
Operator API - extract out a KafkaClusterRef - a reference to cluster being proxied.

The `TargetClusterRef` is essentially a pointer to the cluster being proxied.  It might point at a Cloud-based Kafka service, a Kafka cluster running on bare metal.  It might point at a Kafka running on within a Kubernetes Cluster.  The VirtualCluster CR will reference a `TargetClusterRef` to describe the Kafka cluster it proxying. 

```
apiVersion: proxy.kroxylicious.io/v1alpha1
kind: VirtualCluster
metadata:
  name: mycluster
spec:
  ...
  targetCluster:
    clusterRef:
      kind:TargetClusterRef
      group: proxy.kroxylicious.io 
      name: my-cluster
```

with the TargetClusterRef CR looking like this:

```
apiVersion: proxy.kroxylicious.io/v1alpha1
kind: KafkaClusterRef
metadata:
  name: mycluster
spec:
   bootstrap: bootstrap:9092
   protocol: TCP|TLS
```

In the case where Kroyxlicious is used to proxy an instance of Kafka stood up by Strimzi, it is proposed that the targetCluster can reference the Kafka CR.

```
apiVersion: proxy.kroxylicious.io/v1alpha1
kind: VirtualCluster
metadata:
  name: mycluster
spec:
  ...
  targetCluster:
    clusterRef:
      kind:Kafka
      group: strimzi.io
      name: my-cluster
```


proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,764 @@
...
+      # where the developer wishes their VC to use Certificates created by OpenShift (useful for ClusterIP type TLS).
+      annotations:
+         x: y
+      labels:
+         x: y
+
+  # Points to the cluster being proxied.  Can either be Strimzi Kafka resource or endpoint details.
+  targetCluster:
+    clusterRef:
+      kind: Kafka|TargetClusterRef  # must be Kafka or TargetClusterRef
```
SamBarker, line 236: This is still an `or` because we can infer everything we need to know to connect to a Strimzi managed kafka from the Kafka CR?

I'm just wondering if TargetClusterRef should be the thing that provides a pointer to the Strimzi CR, or maybe any other operator CR.

k-wall: > This is still an or because we can infer everything we need to know to connect to a Strimzi managed kafka from the Kafka CR?

Yes.  If we have a Kafka reference and a listener name, we can suck everything we need to connect from Kafka's status _except_ the client-auth cert.     (aside: we learnt at the face to face that Strimzi depopulates the status: section if reconciliation is paused.   I'm not intending to think much about that problem (if that it what it is) right now)

> I'm just wondering if TargetClusterRef should be the thing that provides a pointer to the Strimzi CR, or maybe any other operator CR.

Maybe. The conciseness of avoiding TCR if Kafka is used is alluring though.   I'm still imagining that demo... "and ladies and gentleman, look how easy it is a to add record validation your Strimzi Kafka cluster.... et voila" :)

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,764 @@
...
+Cluster, it might be service running on bare metal, or a Kafka service of the cloud provider.
+
+The TargetClusterRef CR may the responsibility of a Developer or the Infrastructure admin.
+
+The TargetClusterRef is spec only.  It may be referenced by many VirtualCluster belonging to the same Proxy, or
+either VirtualClusters belonging to different proxies.
+
+```yaml
+apiVersion: proxy.kroxylicious.io/v1alpha1
+kind: TargetClusterRef
```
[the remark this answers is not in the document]
k-wall: Yes, I see those things too. There's a line to be walked - something that is implementable now vs thinking strategically about the cross component concerns.  I'm leaning towards the former right now.   The VC CR's uses a reference so we could plug in other resources at a later date

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,764 @@
...
+metadata:
+  name: mycluster
+spec:
+   bootstrap: bootstrap:9092
```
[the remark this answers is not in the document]
k-wall: I hadn't thought about that at all.  Doesn't this apply every place where we have the Proxy dialing out (KMS, upstream, Schema Registry).   I'm wondering if the control you are talking about needs to be higher up (Proxy CR).

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,764 @@
...
+   # per filter status 
+   filters:
+   - name: myclusterip
+     conditions:
+     - ...
+```
+
+## TargetClusterRef
+
+TargetClusterRef points to a remote Kafka cluster.  It might be a Kafka cluster stood up on remote Kubernetes
```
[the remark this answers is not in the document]
k-wall: I hope I have this use-case covered already. The VC's clusterRef field can point at either a Strimzi Proxy or a TCR.  The last worked example is that case.

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
robobario: How would you manifest a single proxy housing 2 virtual clusters with targets that require different dnsPolicies? The pod can only have one.

Maybe it's better off being configurable as part of the pod template for the proxy pod.

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
robobario: I think Tom is saying that `remote` suggests off-cluster. Maybe update to:

> TargetClusterRef points to a Kafka cluster.  It might be a Kafka cluster stood up on the same Kubernetes cluster, or a remote Kubernetes Cluster,

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -223,35 +232,19 @@ spec:
 
   # Points to the cluster being proxied.  Can either be Strimzi Kafka resource or endpoint details.
   targetCluster:
-    # one of: resourceRef or bootstrapping
-    resourceRef:
-      kind: Kafka  # must be Kafka
-      group: strimzi.io # must be strimzi.io
+    clusterRef:
+      kind: Kafka|TargetClusterRef  # must be Kafka or TargetClusterRef
```
robobario, line 236: the examples later on imply that we have default values for `kind` and `group`, defaulting to TargetClusterRef defails. We should doc that somewhere.

robobario (approved):
LGTM with nits, I like that this makes the configuration consistent across Kafka and TargetClusterRef (from VirtualCluster's point-of-view)

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
robobario: A suggestion from our [naming discussion](#55) is to name the kind something like `KafkaBootstrap` or `KafkaClusterRef`, leaving the `target` naming for the VirtualCluster.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -0,0 +1,764 @@
...
+metadata:
+  name: mycluster
+spec:
+   bootstrap: bootstrap:9092
+   protocol: TCP|TLS
+   nodeIdRanges:
+   - name: range1
+     range:
+      startInclusive: 0
+      endExclusive: 3
```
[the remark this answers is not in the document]
k-wall: > Just to be clear, we wouldn't need this for the Strimzi integration right? We would scrape the nodeIds out of the KafkaNodePool status.

Yes, that's right.  

Ingress types clusterIP, openShiftRoute, and gateway need to know the target broker topology so the operator can create the the right resources with the right content (Services, Routes or TLSRoutes).  loadBalancer does not need it.

proposals/kroxylicious-operator-api-alternative.md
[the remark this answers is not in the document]
k-wall: Let's see if we see this requirement from users.

proposals/diagrams/kroxylicious-operator-apis.png
robobario, line 1: the transparent background makes it hard to read in dark mode

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -498,21 +546,30 @@ spec:
...
 
+```yaml
+apiVersion: kroxylicious.io/v1alpha1
+kind: KafkaClusterRef
+metadata:
+  name: mytargetcluster
+spec:
+  bootstrap: upstream:9092
+  protocol: TLS
+  nodeIdRanges:
```
robobario, line 566: I'm wondering if `nodeIdRanges` should be here when we might need it for other ref types. Maybe it could be moved up on the `TargetCluster` object. It could be a useful stepping stone for Strimzi integration so that we can extract the addresses from the Kafka CR and not have to go delving for the nodeIds from the relevant KafkaNodePool.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -647,40 +718,7 @@ spec:
...
-
-* Use the Kafka CR status to provide things like TLS settings
-
-Parallel work:
-
-~1. Kroxylicious - server certificate grab bag support (serve the right certificate and intermediates based on SNI match)~ _(edit: not required)_
-1. Allow Kroxylicious to have multiple listeners per virtual cluster _routed to the same target cluster listener_.  This makes the cluster accessible by both on-cluster and off-cluster workloads.
-1. Allow KafkaProtocolFilters to reference secrets
-1. Proxy dynamically reloads files providing TLS material (i.e. allows certificates to be rolled).
+See https://github.com/orgs/kroxylicious/projects/6
```
robobario, line 721: should we link to a private project?

robobario (approved):
LGTM, +1 to the PR

proposals/kroxylicious-operator-api-alternative.md
robobario: it is public now :tada:

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -223,35 +232,19 @@ spec:
...
-      listenerName: listener # name of strimzi listener 
-    bootstrapping:
-       bootstrap: bootstrap:9092
-       protocol: TCP|TLS
-       nodeIdRanges:
-       - name: mybrokers
-         range:
-          startInclusive: 0
-          endExclusive: 3
+      listenerName: listener # name of strimzi listener, permitted only for Kafka (group strimzi.io)
```
SamBarker, line 239: Should be marked as optional right? Is it required for Strimzi references or can we default it?

proposals/kroxylicious-operator-api-alternative.md
SamBarker: If we can deterministically map based on ID's without user input then we should do that. I can see your point @robobario an if we can't do a deterministic & stable mapping without user config then we probably do need to move or duplicate it.

SamBarker (approved):
LGTM 

Thanks @k-wall

proposals/kroxylicious-operator-api-alternative.md
k-wall: Yeah, it is optional.

proposals/kroxylicious-operator-api-alternative.md
k-wall: We are suck it and see.

proposals/kroxylicious-operator-api-alternative.md
k-wall: It might also be optional if the Kafka object has only one listener, or perhaps, if there is only one on-cluster listener.

PR #54, opened 2025-02-25
Home Grown Ingress API - Correction - This proposal doesn't rely on the certificate 'grab-bag'

k-wall:
Correction - This proposal doesn't rely on the certificate 'grab-bag'


proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -677,10 +677,10 @@ spec:
 
 Parallel work:
 
-1. Kroxylicious - server certificate grab bag support (serve the right certificate and intermediates based on SNI match)
-2. Allow Kroxylicious to have multiple listeners per virtual cluster _routed to the same target cluster listener_.  This makes the cluster accessible by both on-cluster and off-cluster workloads.
-3. Allow Filters to reference secrets
-4. Proxy dynamically reloads files providing TLS material (i.e. allows certificates to be rolled).
+~1. Kroxylicious - server certificate grab bag support (serve the right certificate and intermediates based on SNI match)~ _(edit: not required)_
```
robobario, line 680: there's also a mention of the grab bag up in the [VirtualCluster example](https://github.com/kroxylicious/design/blob/main/proposals/kroxylicious-operator-api-alternative.md#virtualcluster) that could be removed

Does that mean we will start with only supporting a single certificateRef in the array? Would be good to comment that somewhere if so.

robobario (approved):
LGTM

proposals/kroxylicious-operator-api-alternative.md
k-wall: Yes, that's right.  I think we should keep the array even though we'll only support a single element.

PR #55, opened 2025-02-25
Operator crd naming

robobario:
Lets nail down the CRD names/groups before we start implementing


robobario:
I think we should also prefix the remaining CRDs for consistency, `KafkaVirtualCluster` and `KafkaFilter`. The only one I pause on is `KafkaTargetClusterRef`, getting cumbersome.

tombentley:
@robobario:

* `VirtualKafkaCluster` read better than `KafkaVirtualCluster`, but breaks the consistency.
*  I could like with `KafkaFilter` but wonder if `KafkaProtocolFilter` might be better.
* `KafkaTargetClusterRef`: what is this? That's not defined in the document AFAICS.

robobario:
> KafkaTargetClusterRef: what is this? That's not defined in the document AFAICS.

Ah yep, getting ahead of myself, I was thinking of the thing in #53.

tombentley:
> > KafkaTargetClusterRef: what is this? That's not defined in the document AFAICS.
> 
> Ah yep, getting ahead of myself, I was thinking of the thing in #53.

For that one, I don't think we should embed the role that it plays in a `VirtualKafkaCluster` into the type. I.e. it represents a Kafka cluster, the _target_ aspect belongs in the `VirtualKafkaCluster`. So it's a `KafkaCluster` or a `KafkaBootstrap` or something like that.

robobario:
With `KafkaVirtualCluster` vs `VirtualKafkaCluster` if the others are all Kafka prefixed I'd prefer the consistency of `KafkaVirtualCluster.

SamBarker:
> `VirtualKafkaCluster` read better than `KafkaVirtualCluster`, but breaks the consistency.

I'd always vote for readability over consistency. 

I'm not really sold on the idea of throwing Kafka on the front of everything. Maybe it makes kubectl / oc commands easier? but things are already in a group which could just as easily provide the `Kafka` context to all the types.

robobario:
If we only want to lower the chances of collisions then maybe I'd go with:

```
KafkaProxy
KafkaProxyIngress
ProtocolFilter
VirtualCluster
```

robobario:
after chatting I've updated it to:

```
KafkaProxy
KafkaProxyIngress
KafkaProtocolFilter
VirtualKafkaCluster
```

SamBarker (approved):
LGTM 

I think we wait for a +1 form @k-wall before merging.

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -30,10 +30,10 @@ has prompted him to think about an alternative API design.
...
-* Proxy CR - an instance of the Kroxylicious
-* ProxyIngress CR - Defines a way to access a Proxy
-* VirtualCluster CR - a virtual cluster
-* Filter CR - a filter definition
+* KafkaProxy CR - an instance of the Kroxylicious
+* KafkaProxyIngress CR - Defines a way to access a KafkaProxy
+* VirtualKafkaCluster CR - a virtual cluster
+* KafkaProtocolFilter CR - a filter definition
 
 ![image](https://github.com/user-attachments/assets/8767da6b-3fc4-4fbc-99bc-64e93c06bc70)
```
tombentley, line 38: @robobario do you want to update the diagram?

tombentley (approved):
LGTM, with one quibble

robobario:
Keith +1'd it in another channel

PR #56, opened 2025-03-03
Rename KafkaProxyIngress to KafkaGateway

robobario:
We have landed on [Gateway](https://github.com/kroxylicious/kroxylicious/pull/1848#issuecomment-2690076942) as our grudgingly preferred name for "the thing that presents one (or more) virtual kafka cluster to the network".

The Kafka protocol requires us to be able to identify each node that the client is attempting to connect to, so that we can forward the traffic to the correct upstream broker. Our strategies are to map the listening port, or the SNI hostname, to the upstream node id. Because one of theses strategies spans numerous ports, we found it awkward to name it a `listener` or `ingress`.

Gateway is similarly overloaded, but we think it's a better term for this melange of "how do we expose this cluster to the world", and "how do we map connections to nodes".

### DIFF
--- a/proposals/kroxylicious-operator-api-alternative.md
+++ b/proposals/kroxylicious-operator-api-alternative.md
@@ -9,7 +9,7 @@
   * [Developer](#developer)
 * [API](#api)
   * [KafkaProxy](#kafkaproxy)
-  * [KafkaProxyIngress](#kafkaproxyingress)
+  * [KafkaGateway](#kafkagateway)
   * [VirtualKafkaCluster](#virtualkafkacluster)
   * [KafkaProtocolFilter](#kafkaprotocolfilter)
 * [Worked examples](#worked-examples)
@@ -31,7 +31,7 @@ has prompted him to think about an alternative API design.
 # CRDs
 
 * KafkaProxy CR - an instance of the Kroxylicious
-* KafkaProxyIngress CR - Defines a way to access a KafkaProxy
+* KafkaGateway CR - Defines a way to access a KafkaProxy
 * VirtualKafkaCluster CR - a virtual cluster
 * KafkaProtocolFilter CR - a filter definition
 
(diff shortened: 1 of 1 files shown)


proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -113,11 +113,11 @@ It declares a named ingress mechanism - in other words, a way for traffic to get
...
-kind: KafkaProxyIngress
+kind: KafkaGateway
 metadata:
   name: myclusterip
 spec:
-  # reference to the proxy that will use this ingress.
+  # reference to the proxy that will use this gateway.
   proxyRef:
     kind: KafkaProxy  # if present must be Gateway, otherwise defaulted
     group: kroxylicious.io # if present must be proxy.kroxylicious, otherwise defaulted=
```
robobario, line 123: incorrect comments, will tack onto this PR

k-wall (approved):

k-wall:
I know it was my idea, but I've had chance to reflect.  There's two things I want to call out.

First the level of abstraction is different.   For Kroxylicious proper, the Gateway exposes levers to control the node identification strategy and lets the user control precisely the advertised names and ports.   At the Operator API level, we are much more opinionated.  The API is in terms of clusterIPs, loadBalancers and OpenShift Routes.     Things like node identification strategy and (sometimes) the ability to control the advertised disappear.   They are implementational details which are a concern to the Operator.

Second, in the Operator API, Gateways are defined at the Proxy level, then associated with the VirtualClusters that need to use them.  That's a different model than the one we have at the level of Kroxylicious proper.

Of those two concerns, it is the first one that gives me most pause for thought.

We could use the same term (gateway) and use the documentation to explain the differences.   However, maybe users would struggle to grok the difference.

Maybe sticking with ProxyIngress better.

k-wall (changes requested):
see above.

robobario:
Okay, I think the argument that the Gateway behaviours are an implementation detail is strong and having them both named Gateway could create confusion. I'm happy to stick with `KafkaProxyIngress`.

SamBarker:
I think @k-wall is right Ingress and Gateway are different concepts lets not try to conflate them

PR #58, opened 2025-03-13
Rename api design

k-wall:
Reflecting the fact this is the way we have chose.

### DIFF
--- a/proposals/kroxylicious-operator-api-v1alpha.md
+++ b/proposals/kroxylicious-operator-api-v1alpha.md
@@ -1,7 +1,7 @@
-# Kroxylicious Operator API - Alternative Proposal.
+# Kroxylicious Operator API - v1alpha
 
 <!-- TOC -->
-* [Kroxylicious Operator API - Alternative Proposal.](#kroxylicious-operator-api---alternative-proposal)
+* [Kroxylicious Operator API - v1alpha](#kroxylicious-operator-api---v1alpha)
 * [Background](#background)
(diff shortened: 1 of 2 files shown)


tombentley (approved):
The name of the template files suggests we once thought of these documents as having numeric ids. It that worth keeping? If not the chronological evolution of decisions will be harder to pick out for our future selves.

proposals/kroxylicious-operator-api-v1alpha.md
```diff
@@ -1,7 +1,7 @@
-# Kroxylicious Operator API - Alternative Proposal.
+# Kroxylicious Operator API - v1alpha
```
SamBarker, line 1: Do we intend to reflect the actually implemented API back to this repo? 
If so should this be `v1alpha1` 
if not should this be `v1alpha skeleton`  or `v1alpha proposal`?

SamBarker (approved):
This is certainly less confusing than before, though I'm still wondering if its quite the right title.

robobario (approved):
LGTM and I'm +1 to numbering the proposals

proposals/kroxylicious-operator-api-v1alpha.md
k-wall: > Do we intend to reflect the actually implemented API back to this repo?

I don't think we should do that.  I think our CRDs should become the source of truth.  

I think updating the proposal to assist our development is fine.

PR #61, opened 2025-03-17
Rename KafkaClusterRef to KafkaService and refactor responsibilities

### LINKED ISSUE #59: [API Design] No way to express TLS ciphersuite or protocol restrictions if referencing a Kafka CR
https://kroxylicious.slack.com/archives/C04V1K6EAKZ/p1741818868349069

@tombentley said:

> a KafkaClusterRef has got spec.tls.protocols and spec.tls.cipherSuites, but it's also proposed for a VirtualKafkaCluster 's spec.targetCluster.clusterRef  to support referencing a Strimzi Kafka CR. In this case, how can the user specify the allowed/denied TLS protocols and cipher suites?

robobario:
Closes #59

1. Rename KafkaClusterRef custom resource to KafkaService
2. KafkaService will be the place that either contains bootstrapServers, or references other CRs like a Strimzi Kafka
3. `targetCluster` object on VirtualKafkaCluster is replaced with `targetKafkaServiceRef` as it's responsibilities have been moved to KafkaService

This enables us to configure the TLS options for both inline bootstrapServers, and kafka connections derived from other resources, like a reference to strimzi Kafka Custom Resource. Also when we declare the target cluster topology it will apply to both mechanisms (may be useful at first with Strimzi, though we should be able to discover the nodeIds).

`KafkaClusterRef` was also a painful name as we have adopted `*Ref` as our property naming convention, pushing us towards a `clusterRefRef` property. We think KafkaService is a better description.


robobario:
was merged in by hand

proposals/kroxylicious-operator-api-alternative.md
```diff
@@ -231,20 +231,8 @@ spec:
...
-  targetCluster:
-    clusterRef:
-      group: strimzi.io|kroxylicious.io # must be strimzi.io or kroxylicious.io
-      kind: Kafka|KafkaClusterRef  # must be Kafka (belonging to group strimzi.io) or TargetClusterRef (belonging to group kroxylicious.io)
-      name: my-cluster
-      listenerName: listener # name of strimzi listener, permitted only for Kafka (group strimzi.io).  (Optional)
-    tls:
-      # Optional - client auth certificate
-      # secret provided by the Developer.
-      certificateRef:
```
robobario, line 243: are we happy to have the client cert specified by the kafka service ref? If we have it on the virtual cluster then different virtual clusters can have different certificates for the same kafka service. Else the user would have to create two KafkaService's with the same details except for the client certs.

Or we could support some combination of the two, allowing VCluster to override the client certs

k-wall (approved):
LGTM.. but please update the TOC

proposals/001-kroxylicious-operator-api-v1alpha.md
```diff
@@ -35,7 +35,7 @@ has prompted him to think about an alternative API design.
 * KafkaProxyIngress CR - Defines a way to access a KafkaProxy
 * VirtualKafkaCluster CR - a virtual cluster
 * KafkaProtocolFilter CR - a filter definition
-* KafkaClusterRef CR - a reference to a kafka cluster
+* KafkaService CR - description of a target kafka service
```
SamBarker, line 38:
```suggestion
* KafkaService CR - describes how the proxy should connect to a target Kafka cluster
```

proposals/001-kroxylicious-operator-api-v1alpha.md
```diff
@@ -289,23 +277,28 @@ spec:
...
-KafkaClusterRef points to a Kafka cluster. It might be a Kafka cluster stood up on the same Kubernetes cluster, a cluster
+KafkaService points to a Kafka cluster. It might be a Kafka cluster stood up on the same Kubernetes cluster, a cluster
 running on a remote Kafka cluster, it might be service running on bare metal, or a Kafka service of a Cloud Provider.
 
-The KafkaClusterRef CR may the responsibility of a Developer or the Infrastructure admin.
+The KafkaService CR may the responsibility of a Developer or the Infrastructure admin.
 
-The KafkaClusterRef is a spec only resource.  It may be referenced by many VirtualCluster belonging to the same Proxy, or
+The KafkaService is a spec only resource.  It may be referenced by many VirtualCluster belonging to the same Proxy, or
 VirtualClusters belonging to different proxies.
```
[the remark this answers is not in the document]
robobario: I guess we're just being specific that there's no hierarchy here. Virtual clusters are associated with a single KafkaProxy. KafkaService's are not associated with anything in particular.

proposals/001-kroxylicious-operator-api-v1alpha.md
```diff
@@ -231,20 +231,8 @@ spec:
...
-      listenerName: listener # name of strimzi listener, permitted only for Kafka (group strimzi.io).  (Optional)
-    tls:
-      # Optional - client auth certificate
-      # secret provided by the Developer.
-      certificateRef:
-        kind: Secret # if present must be Secret, otherwise defaulted to Secret
-        group: ""  # if present must be "", otherwise defaulted to ""
-        name: clientcert
-        namespace: # namespace of the secret, if omitted assumes namespace of this resource
+  kafkaServiceRef: # required
```
tombentley, line 234: I think it can help people understand the resource if we include the role that the referred-to thing plays in the name of the property. I.e. `targetKafkaServiceRef` says a lot more to someone unfamiliar with the API than `kafkaServiceRef`.

proposals/001-kroxylicious-operator-api-v1alpha.md
```diff
@@ -289,23 +277,28 @@ spec:
     selectorConfig: {}
 ```
 
-## KafkaClusterRef
+## KafkaService
 
-KafkaClusterRef points to a Kafka cluster. It might be a Kafka cluster stood up on the same Kubernetes cluster, a cluster
-running on a remote Kafka cluster, it might be service running on bare metal, or a Kafka service of a Cloud Provider.
+KafkaService points to a Kafka cluster. It might be a Kafka cluster stood up on the same Kubernetes cluster, a Kafka cluster
+running on a remote Kubernetes cluster, it might be service running on bare metal, or a Kafka service of a Cloud Provider.
```
tombentley, line 283:
```suggestion
KafkaService represents, and encapsulates connection information for, a Kafka cluster. It might be a Kafka cluster stood up on the same Kubernetes cluster, a Kafka cluster
running on a remote Kubernetes cluster, it might be service running on bare metal, or a Kafka service of a Cloud Provider.
```

proposals/001-kroxylicious-operator-api-v1alpha.md
robobario: done

robobario:
updated diagram too

proposals/001-kroxylicious-operator-api-v1alpha.md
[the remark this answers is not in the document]
SamBarker: Ok, but wouldn't stopping the sentence after 
>referenced by many VirtualCluster.

Do the same? 

The following is a bit of mouthful but I think makes the point clearer.

```suggestion
The KafkaService is a spec only resource. It is global in scope and thus can be referenced by multiple VirtualCluster definitions regardless of the KafkaProxy instance hosting them.
```

robobario:
the Upstream specified by Kafka CR example needs to be updated

proposals/001-kroxylicious-operator-api-v1alpha.md
```diff
@@ -364,9 +356,8 @@ spec:
   ingress:
   - name: myclusterip
 
-  targetCluster:
-    clusterRef:
-      name: mytargetcluster
+  targetKafkaServiceRef:
```
robobario, line 359: would `proxiedKafkaServiceRef` be more easily understood by users?

tombentley: The comment above is `# Points to the cluster being proxied`. It's interesting that such an explanation is necessary, and that that's the language we use there.

SamBarker: I dunno, I can see arguments both ways really. 

For me `target` encapsulates what it actually and I think fairly obvious what it means. My other hesitation is `proxiedKafkaService` is the result of applying the CRs so feels like a more natural name for a thing describing how to connect to Kafka via the proxy. 

But I can certainly see `proxiedKafkaServiceRef` as being easier to understand when just reading a CR without any context. 

That said the terminology we use in the proxy is `upstream` so to be consistent it should be `upstreamKafkaServerRef`

tombentley: I'm fine with merging as-is.

> That said the terminology we use in the proxy is upstream so to be consistent it should be upstreamKafkaServerRef

Would be more compelling if we actually _liked_ that terminology. But we've questioned it in the past, so...
Also that terminology is usually used in a slightly different context: where we have two connections and want to distinguish them with a single qualifier and it's desirable for the one qualifier to be "the opposite" of the other. In contexts like this one we don't use the term "downstream cluster".

tombentley (approved):

proposals/001-kroxylicious-operator-api-v1alpha.md
robobario: issue to carry on https://github.com/kroxylicious/design/issues/63

PR #65, opened 2025-04-15
Proposal for triggering a proxy re-configuration

SamBarker:


proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+[^1]: For simplicity CRC32 as that's already supported by the JDk. However, the algorithm used is immaterial.
+## Advantages
+Maintains the same logical division between reconcilers thus avoiding the need for the aggregating reconciler to watch all resources involved and thus the need for it to understand the details of the full configuration model and its kubernetes representation. 
+
+## Limitations
+1. The blast radius of any config change is still the whole proxy process. I.e. a malformed config file still prevents the whole proxy from starting.
+This is largely mitigated by the fact the config file is in the kubernetes case is programmatically generated, and we can validate that before applying it.
+2. Any change to the config model requires the entire proxy process to restart. I.e. renewing the certificate of one virtual cluster requires the all connections to the hosting proxy instance to be disconnected and reconnect.
```
tombentley, line 57: Two things worth calling out:

1. this mechanism is oblivious to any changes in the proxy configuration that doesn't manifest in any of the referenced resources. For example if the operator was inserting some boiler plate configuration (e.g. setting up metrics, for example) and this changed between one operator version and the next. In this example we'd expect that the `image` would change, so a rollout would happen. But I think it's worth mentioning all the same.

2. this mechanism requires that the operator actually `get` all the resources (in order to obtain the `metadata.generation` and `metadata.uid`). It means the operator can't just pass references to Secrets (for example) to be mounted into the pod. Instead it must have access rights to read those resources itself. This should be fine for Secrets, because the operator necessarily needs to be able to create Deployments, which means it inherently has access to Secrets anyway. But I suppose it could be a problem for CSI mounted secret stores.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Rejected Alternatives
+1. Have the Kafka Proxy reconciler (KPR) watch all resources involved for changes and trigger on any update. This has a variety of issues
+   1. We conflate concerns between reconcilers as we end up with two (or more) reconcilers watching the same resource.
+   2. The KPR can, and likely will, be triggered on invalid or intermediate states thus making for vastly more complicated error handling and recovery
+   3. It becomes much harder for operators to work out what triggered a particular change to the running processes.
+2. The proxy process monitors the config file(s) for changes 
+   1. What set of files should the proxy monitor? The set can grow and shrink during parsing as filers are added and removed
+   2. How does the proxy know when a file is fully updated? User saves changes to the live config file before looking up the correct config syntax.
```
tombentley, line 66: There's no pathway for reporting errors with the configuration back to the operator.

tombentley (commented):
A few nits.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,72 @@
...
+* [Rejected Alternatives](#rejected-alternatives)
+<!-- TOC -->
+# Problem Statement
+There are a wide array of Kubernetes resources which are used to create the configuration used by any given Kroxylicious proxy process. These resources are consumed and generated by multiple independent reconcilers (in Java Operator SDK terms). Which leads to a tangled web of resources, reconcilers and triggers. 
+
+We split the operator into multiple reconcilers so that there was a single writer for each of our "intermediate" resources, and thus it is much easier to reason about what changed and why. However, that means that we do not have a single global view of when a proxies configuration needs re-generated.
+
+# Proposed Solution
+
+The Kroxylicious operator is built up from what we refer to as "leaf" and "aggregating" reconcilers.
```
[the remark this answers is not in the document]
SamBarker: Does the table not do that? Or do you have a better way of doing it?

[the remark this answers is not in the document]
tombentley: The table does do that, but it happens a bit later on when reading the document. In the mean time the reader might not be really be understanding what's being said.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,72 @@
+# Proposal 002 - Automatically re-load proxy pods on Kubernetes config change 
+
+<!-- TOC -->
+* [Proposal 002 - Automatically re-load proxy pods on Kubernetes config change](#proposal-002---automatically-re-load-proxy-pods-on-kubernetes-config-change-)
```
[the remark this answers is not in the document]
SamBarker: Yeah I was caught between `re-loading configuration` and `restarting pods` and realising one doesn't restart in kube.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
SamBarker: On 1. in the comment. Isn't the config map holding the actual proxy config one of the dependent resources we would generate a checksum for? And thus something like adding metrics would be caught.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,75 @@
...
+2. Any change to the config model requires the entire proxy process to restart. I.e. renewing the certificate of one virtual cluster requires the all connections to the hosting proxy instance to be disconnected and reconnect.
+3. The operator requires permissions to actually get all the resources (in order to obtain the `metadata.generation` and `metadata.uid`). While it can just query for the specific metadata the Kubernetes RBAC model works at verb/resource level not to fields within resources. 
+
+# Rejected Alternatives
+1. Have the Kafka Proxy reconciler (KPR) watch all resources involved for changes and trigger on any update. This has a variety of issues
+   1. We conflate concerns between reconcilers as we end up with two (or more) reconcilers watching the same resource.
+   2. The KPR can, and likely will, be triggered on invalid or intermediate states thus making for vastly more complicated error handling and recovery
+   3. It becomes much harder for operators to work out what triggered a particular change to the running processes.
+2. The proxy process monitors the config file(s) for changes
```
k-wall, line 65: I think *eventually* we will give the Proxy the ability to watch files containing things like TLS material and passwords and have those reloaded on the fly.   As you allude to, there's a fair amount of work in doing that.   There's the mechanism of watching the files, having to reload the contents and apply the result to the model and get things like SSL contexts to refresh (both Netty and JDK flavours).  Should we break existing TCP connections and force the clients to reconnect?   That's a lot of work.

I am trying to say, I think this proposal is right one for now.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,75 @@
...
+   1. What set of files should the proxy monitor? The set can grow and shrink during parsing as filers are added and removed
+   2. How does the proxy know when a file is fully updated? User saves changes to the live config file before looking up the correct config syntax.
+   3. We add complexity to the proxy, extra threads for file watches.
+   4. There is no mechanism for the proxy to report errors back to the operator. e.g. conflicting port allocations. 
+3. Follow the model pioneered by HTTPD but copied by many other services which support vhosting of a `conf.d` directory and have one file per virtual cluster.
+   1. Increases the complexity of managing bare metal clusters 
+   2. Makes crafting valid config harder (identifying for port collisions across a directory of file)
+   3. Increases complexity within the proxy
+   4. No mechanism to report problems encountered at the proxy level back to the operator.
+   5. Lack of developer bandwidth to context switch to the proxy
```
k-wall, line 75: I'm not sure what this bullet means.

k-wall (approved):

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
SamBarker: Yes, fully agree. The proxy should be re-loading much more automatically and in a less drop everything and restart fashion but our time is better spent else where for now.

SamBarker: But I'm very glad to hear you agree!

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,75 @@
...
+3. The operator requires permissions to actually get all the resources (in order to obtain the `metadata.generation` and `metadata.uid`). While it can just query for the specific metadata the Kubernetes RBAC model works at verb/resource level not to fields within resources. 
+
+# Rejected Alternatives
+1. Have the Kafka Proxy reconciler (KPR) watch all resources involved for changes and trigger on any update. This has a variety of issues
+   1. We conflate concerns between reconcilers as we end up with two (or more) reconcilers watching the same resource.
+   2. The KPR can, and likely will, be triggered on invalid or intermediate states thus making for vastly more complicated error handling and recovery
+   3. It becomes much harder for operators to work out what triggered a particular change to the running processes.
+2. The proxy process monitors the config file(s) for changes 
+   1. What set of files should the proxy monitor? The set can grow and shrink during parsing as filers are added and removed
+   2. How does the proxy know when a file is fully updated? User saves changes to the live config file before looking up the correct config syntax.
```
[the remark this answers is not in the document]
SamBarker: Its not really a use case I value either, but it was meant to be indicative of the types of timing problems that exist. 

I guess the other example for bare metal would be replacing the certificate files and updating the ${ENV_VAR} or similar holding the password for that file. Re-loading the cert file without co-ordination just fails.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
SamBarker: I was trying to say we are all focused on the operator and we don't have the capacity to go re-work proxy config loading as otherwise this model is really a big part of the answer. 

Actually maybe its not, this addresses the orthogonal concern of blast radius, tweaking on VKC causing a stampeding heard as all connections restart. It doesn't really address the co-ordination problem

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,76 @@
...
+* [Rejected Alternatives](#rejected-alternatives)
+<!-- TOC -->
+# Problem Statement
+There are a wide array of Kubernetes resources which are used to create the configuration used by any given Kroxylicious proxy process. These resources are consumed and generated by multiple independent reconcilers (in Java Operator SDK terms). Which leads to a tangled web of resources, reconcilers and triggers. 
+
+We decided to split the operator into multiple reconcilers so that there was a single writer for each of our "intermediate" resources (KafkaService, VirtualKafkaCluster, etc), and thus it is much easier to reason about what changed and why. However, that means that we do not have a single global view of when a proxy's configuration needs to be re-generated.
+
+# Proposed Solution
+
+The Kroxylicious operator is built up from what we refer to as "leaf" (such as the Kafka Service Reconciler) and "aggregating" (with the Kafka Proxy Reconciler being the ultimate aggregator) reconcilers.
```
robobario, line 19: Feels like the `leaf` and `aggegator` terms could use a definition immediately. Also feels a bit funky having the `leaf` naming when VKC has children. `component reconciler` vs `aggregating reconciler`? Tough one.

robobario: or `intermediate` as you've named it in the problem statement?

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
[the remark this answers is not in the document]
k-wall: > Re-loading the cert file without co-ordination just fails.

I was wondering if the semantics would be the SSLContext would only be swapped for the TLS server (or client) once a new one has been successfully created. So if the key material updates first then password later, the updating thread(?) would ignore the error in the intervening period.  The old SSLContext would continue to be used during this time.

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
k-wall: I like aggegator and intermediate

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
```diff
@@ -0,0 +1,77 @@
...
+The Kafka Service Reconciler, KSR, will: 
+1. Watch for Kafka Service CRs to appear
+2. Parse the KafkaService
+   1. Follow any references to external secretes
+   2. For each valid secret
+      1. Compute a checksum of secret UID & Generation
+3. Annotate the KafkaService with `kroxylicious.io/referent-checksum = <computedChecksum>`  
+
+
+[^1]: For simplicity CRC32 as that's already supported by the JDk. However, the algorithm used is immaterial.
```
tombentley, line 51:
```suggestion
[^1]: For simplicity CRC32 as that's already supported by the JDK. However, the algorithm used is immaterial.
```

SHA1 is also already supported by the JDK, but I'm fine with CRC32 for now.

tombentley (approved):
A few very nitty nits, but I think this is good enough for now.

SamBarker:
/me stares at IntelliJ why are my commits not signed...

proposals/002-Auttomaticly-reload-on-Kubernetes-config-change.md
SamBarker: I don't think the table is exhaustive and it shouldn't be. I can add a config map to table. I'll need to go read up on CSI secretes as I don't have any context for writing it up outside of your comment.

tombentley: > I don't think the table is exhaustive and it shouldn't be.

Well it would certainly aid understanding if it covered all the resources we expect to initially be using this mechanism.

> I can add a config map to table. 

Thank you!

SamBarker: I think both the config map reference and CSI issue are noted and resolved so I'm going to merge.

PR #67, opened 2025-05-20
Metrics Improvements

k-wall:
This proposal suggests some improvement to address some short-comings in the existing metrics exposed by the Proxy.


proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,161 @@
...
+
+The new "message_size" differ from the old "payload_size" in several key ways:
+1. New "message_size" will be used to track both **opaque** and **decoded** messages ("payload_size" tracked only decoded messages).
+2. Downstream "message size" will track [message size](https://kafka.apache.org/protocol#:~:text=DESCRIPTION-,message_size,-The%20message_size%20field)
+   as messages come in from or leave for the client.  Upstream "message size" will track message size as messages go to or arrive from the broker.
+3. The design of the metric allows it to be used to track messages that originate within a filter.
+4. The design of metric allow a message-size histogram to be added.  This would allow the collection of statistical information
+   on, say, the size of the produce request or the fetch response.  If added, we would make this configurable.
+
+In Micrometer, `Distribution` type metrics actually comprise three Prometheus metrics `_sum`, `_count` and `_max`.  This means that
```
k-wall, line 72: I'm in two minds about this.  On the one hand, this approach saves an additional counter.  On the other hand, it feels wrong to have the user rely on something with the words `message_size` `kroxylicious_downstream_kafka_rpc_message_size_bytes_count`.   It is counter intuitive.  I'd prefer an explicit counter.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,161 @@
...
+### Deprecate `kroxylicious_inbound*`/`kroxylicious_payload_size_bytes` replace with new `kroxylicious_(downstream|upstream)_kafka_rpc_message_size_bytes` metrics
+
+This proposal suggests that we deprecate the `kroxylicious_inbound*`/`kroxylicious_payload_size_bytes` metrics and replace
+with two new **distribution** (aka histogram) metrics `kroxylicious_downstream_kafka_rpc_message_size_bytes` and `kroxylicious_upstream_kafka_rpc_message_size_bytes`.
+
+The new "message_size" differ from the old "payload_size" in several key ways:
+1. New "message_size" will be used to track both **opaque** and **decoded** messages ("payload_size" tracked only decoded messages).
+2. Downstream "message size" will track [message size](https://kafka.apache.org/protocol#:~:text=DESCRIPTION-,message_size,-The%20message_size%20field)
+   as messages come in from or leave for the client.  Upstream "message size" will track message size as messages go to or arrive from the broker.
+3. The design of the metric allows it to be used to track messages that originate within a filter.
```
SamBarker, line 68: Just to make sure I'm understanding this point. The metric will allow us to count all messages being sent to the broker by the proxy, be they a message from a client or invented by a filter.

Rather than allow us to identify which filter the message came from.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,161 @@
...
+* filters have the ability to change the size of the RPCs as they traverse the proxy (record encryption)
+* filters have the ability to change the flow of messages within the Kafka system. Specifically, they can short-circuit
+  requests by producing their own responses (the broker never sees the request) and introduce their own requests.
+
+## Proposal
+
+### Deprecate `kroxylicious_inbound*`/`kroxylicious_payload_size_bytes` replace with new `kroxylicious_(downstream|upstream)_kafka_rpc_message_size_bytes` metrics
+
+This proposal suggests that we deprecate the `kroxylicious_inbound*`/`kroxylicious_payload_size_bytes` metrics and replace
+with two new **distribution** (aka histogram) metrics `kroxylicious_downstream_kafka_rpc_message_size_bytes` and `kroxylicious_upstream_kafka_rpc_message_size_bytes`.
```
SamBarker, line 62: aren't `kafka_rpc` and `message` synonyms in this context? Does it help users to repeat them?

SamBarker, line 62: To support:

```
Specifically, they can short-circuit requests by producing their own responses (the broker never sees the request) and introduce their own requests.
```
should we add a `source`label e.g. `source=client` or `source=filterNameA` (I think this is possible to do given all additional messages will go via `sendRequest` ?

I can see that being the next question users of the metrics have. "I'm seeing a lot of messages which filter is generating them?"

Though we quickly head to metric explosion as that is per filter instance per virtual cluster...

SamBarker: Oh right you add `originating_filter` later.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,161 @@
...
+for boostrap, the label will be absent.
+
+**How this supports the use-cases?**
+
+This should aid problem diagnosis. If a single upstream broker of a cluster has an issue, it will be obvious from the metrics (and alerts)
+which broker is down.
+
+(Note: that proposal is not suggesting adding `node_id` to the `kroxylicious_downstream_connections_attempts` or
+`kroxylicious_downstream_errors`. This is because at the point these metrics are emitted, the upstream target has not yet
+been identified.  Emitting the metric early has advantages)
```
SamBarker, line 142: >Emitting the metric early has advantages

Could you elaborate a bit on what those advantages are?  

Where are you expecting to track this? in `io.kroxylicious.proxy.internal.ProxyChannelStateMachine#onClientException` ?

SamBarker (commented):
Thanks for thinking this through @k-wall I really appreciate it.

I think I'm largely in agreement with this. I feel like there is one missing metric however to cover `backpressure`, the channel state machine is aware when a given channel is stalled and I think that would be quite important to surface to proxy oeprators.

The  other thought I kept having when reading this (and generally when thinking about our metrics) is that the number explodes quickly. However I'm coming round to the opinion that is a metrics consumer issue, the proxy should surface everything it knows and the scrape config should be used to limit the number of metrics.

proposals/003-metric-improvenents.md
SamBarker: Hmm `originating_filter` or `source` get complicated / nuanced as `sendRequest` is passed to the remaining filters in the chain. So filterA generate a request of X bytes but filter Y changes to be X+1000 bytes. Maybe thats ok? The fact that the message came from filter X is probably more important the strict accuracy of the size metric.

k-wall: Yes, you are right.  The Kafka protocol documentation uses the word message.  I'll use the term kafka_message.

k-wall: Firstly, it is only kroxylicious_upstream_kafka_rpc_message that will instrument async requests and their responses so only that metric will have the possibility if the originating_filter label. Async requests never traverse the downstream side of the proxy.

In the case where filter A sends an async request using sendrequest, and filter B makes the request bigger, the metrics proposed here won't capture the original size.  The proposal is capturing the encoded size, just before it is put to the wire.  'originating_filter' will record Filter A.

SamBarker:
> I'll expand the proposal to include it. Do you think this would be an essential for 0.13.0 or something we could defer?

I was just thinking of a back pressure gauge so very cheap todo in the context of this proposal. The state machine already captures all the transitions so it should be straight forward (though might be hard to test).

I think we need the proposal as a whole in 0.13, but I could live without back pressure if it came to it

proposals/003-metric-improvenents.md
k-wall: @SamBarker any opinion on this?  I'm veering towards adding counter metrics kroxylicious_(up|down)stream_kafka_message_count which the same set of labels as I think clean API wins rather than relying on the fact that distribution (histogram) metric actually comprises a _count metric too.

k-wall:
> The other thought I kept having when reading this (and generally when thinking about our metrics) is that the number explodes quickly. However I'm coming round to the opinion that is a metrics consumer issue, the proxy should surface everything it knows and the scrape config should be used to limit the number of metrics.

Yes, I'm in the same place.   That said, if someone want to add metric configuration that allowed you to control which apikeys emitted metrics and which didn't (allow & deny list I suppose), I wouldn't be against it.   I am not going to propose that now though.  I think scrape configuration is good enough.

proposals/003-metric-improvenents.md
SamBarker: Urgh it's a hard one. As you say the metrics are public API, and consistency and clarity are important.

In isolation I'd definitely say we should have a separate counter but we are in the context of Prometheus metrics with their own conventions. 

I'm not sure that it would be helpful to the end users to have duplicate counters...

🤷

proposals/003-metric-improvenents.md
ShubhamRwt: Iam also confused with this sentence `Emitting the metric early has advantages`

ShubhamRwt:
The proposal looks good to me.............From a user perspective it makes the metrics more clear.

proposals/003-metric-improvenents.md
k-wall: I am wrong.   When I wrote this I believed the that the kroxylicious_downstream_connections_attempts as soon as the connection arrives.  I see now that it is not the case.  TLS is negotiated, we've identified the virtual cluster and the broker node id is know.  So I know think that `kroxylicious_downstream_connections_attempts` should have the same set of labels as `kroxylicious_upstream_connections_attempts` (i.e. add node_id).

This investigation made me realise that in the downstream TLS negotiation fails, or we fail to identify the virtualcluster, we don't actually have a metric for that.  The state machine hasn't been created so we can't tick `kroxylicious_downstream_errors`.   I think initPlainChannel and initTlsChannel need to tick that metric (the virtual_cluster won't be known). Or maybe a separate metric?

SamBarker: They really should be incorporated into the state machine but that's a lot of thread pulling. 

Let's just increment the same counter without the node_id or virtual cluster labels.

proposals/003-metric-improvenents.md
ShubhamRwt: I think having another counter would be better to avoid confusions since `message_size` means a completely different thing then `message_count`

SamBarker (approved):
LGTM

proposals/003-metric-improvenents.md
k-wall: Yes - exactly

proposals/003-metric-improvenents.md
k-wall: Fixed.

proposals/003-metric-improvenents.md
k-wall: I've added this to the proposal.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+  * [Summary](#summary)
+    * [New counter `kroxylicious_downstream_kafka_message_total`](#new-counter-kroxyliciousdownstreamkafkamessagetotal)
+    * [New counter `kroxylicious_upstream_kafka_message_total`](#new-counter-kroxyliciousupstreamkafkamessagetotal)
+    * [New distribution `kroxylicious_downstream_kafka_message_size_bytes`](#new-distribution-kroxyliciousdownstreamkafkamessagesizebytes)
+    * [New distribution `kroxylicious_upstream_kafka_message_size_bytes`](#new-distribution-kroxyliciousupstreamkafkamessagesizebytes)
+    * [New distribution `kroxylicious_message_process_time`](#new-distribution-kroxyliciousmessageprocesstime)
+    * [New gauge metrics `kroxylicious_(up|down)stream_backpressure_state`](#new-gauge-metrics-kroxyliciousupdownstreambackpressurestate)
+    * [New distribution `kroxylicious_(up|down)stream_backpressure_time`](#new-distribution-kroxyliciousupdownstreambackpressuretime)
+    * [Add label `node_id` and `virtual_cluster` to `kroxylicious_(down|up)stream_connections_attempts` and `kroxylicious_(down|up)stream_errors`](#add-label-nodeid-and-virtualcluster-to-kroxyliciousdownupstreamconnectionsattempts-and-kroxyliciousdownupstreamerrors)
```
ShubhamRwt, line 27: Are we going to have the `nodeID` only for the `kroxylicious_(down|up)stream_connections_attempts` and `kroxylicious_(down|up)stream_errors` ? Reading the proposal further I think we are going to add these labels to other metrics also. Reading the sentence it felt like that we are adding the labels to only these specific metrics

ShubhamRwt (approved):
LGTM, overall

proposals/003-metric-improvenents.md
k-wall: > Are we going to have the nodeID only for the kroxylicious_(down|up)stream_connections_attempts and kroxylicious_(down|up)stream_errors ?

no.  `node_id` will be present on all the new metrics where it is appropriate (that's every one except the backpressure metric).

>  Reading the proposal further I think we are going to add these labels to other metrics also. 

Yup.

> Reading the sentence it felt like that we are adding the labels to only these specific metrics

no, I spilt up the summary thus:

- new metrics
- changes to existing metrics that we are keeping
- metrics we are deprecating

If that's not clear I can improve.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+* *2 - there are no quantised buckets configured for this metric. it gives the user count/sum/max values only. it would be
+  hard to use this metric to understand something about produce requests or fetch responses.
+* *1 / *2 - we have contradictory meanings for the `flowing` label.  in *2, requests flow `UPSTREAM`, responses flow `DOWNSTREAM`.
+  In *1 requests flow `DOWNSTREAM` which makes no sense.
+* *3 - downstream errors does not record account for errors such as TLS negotiation errors or failures to determine virtual cluster or broker.  
+
+### General short-comings/weaknesses
+
+* The metrics endpoint populates `HELP` for each metric, but this is populated with nothing more than the metric's name.
+  We ought to be giving a clear description of the metric.
```
tombentley, line 66: I think it's also fair to say that some of the names are less than clear.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+  requests by producing their own responses (the broker never sees the request) and introduce their own requests.
+* filters can perform asynchronous work.  When asynchronous work occurs, the proxy stops reading the downstream or upstream.
+
+## Proposal
+
+## Summary
+
+The following metrics will be added.  They will each be described in more detail later in this section.
+
+| Metric                                           | Type         | Labels                                                                                | Description                                                                             |
```
tombentley, line 92: Do we really need to prefix every metric with `kroxylicious`? It's long, difficult to type and difficult (for some) to pronounce.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+
+The requests sent by filter themselves, the start time for a request will be time filter called `#sendRequest`. The end time
+will be the time the network write at the proxy upstream completes.  For responses, the start time is time the bytes containing
+the message arrived from the network. The end time will be the time the response reaches the originating filter (i.e. the future completes).
+
+### New gauges `kroxylicious_(up|down)stream_backpressure_state`
+
+Reports the backpressure state for upstream and downstream.  The use-case for metric is to allow the user to
+understand when the proxy must pause reading owing whilst asynchronous work completes (for example, in record validation
+requesting a schema or in record encryption, generating a new DEK).
```
SamBarker, line 194: I'm not sure we are thinking about the same backpressure systems. I was thinking about the `channelWritabilityChanged` changed events or their ProxyChannelStateMachine equivalents as driving the gauge. That gives us insight into when we get back pressure at a TCP level. Your description is operating at a different level, that of the proxy itself.

Maybe the two can be conflated into a single measure. 

When the proxy pauses to load a Key from the KMS for record encryption which gauge do you expect to increment? `downstream` on the produce path  and `upstream` on the fetch path?

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Proposal
+
+## Summary
+
+The following metrics will be added.  They will each be described in more detail later in this section.
+
+| Metric                                           | Type         | Labels                                                                                | Description                                                                             |
+|--------------------------------------------------|--------------|---------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------|
+| kroxylicious_downstream_kafka_message_total      | Counter      | virtual_cluster, node_id†, flowing, api_key, api_version, opaque                      | Incremented by one every time a request comes from/response goes to the downstream.     |
```
tombentley, line 94: Why should we have `flowing` as a label? I think people are actually interested in requests and responses, which are fundamentally different things. Having it as a label means to get when you want you need to filter as well as aggregate.

To me this should be two separate metrics:

* `client_to_proxy_request_total`
* `proxy_to_client_response_total`

These names are self-explanatory, and shorter.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+by the proxy decoding and encoding messages and any processing time incurred by the filter.
+
+The requests sent by filter themselves, the start time for a request will be time filter called `#sendRequest`. The end time
+will be the time the network write at the proxy upstream completes.  For responses, the start time is time the bytes containing
+the message arrived from the network. The end time will be the time the response reaches the originating filter (i.e. the future completes).
+
+### New gauges `kroxylicious_(up|down)stream_backpressure_state`
+
+Reports the backpressure state for upstream and downstream.  The use-case for metric is to allow the user to
+understand when the proxy must pause reading owing whilst asynchronous work completes (for example, in record validation
```
SamBarker, line 193: I'm not sure what `owing` is doing in the sentence. I presume its a hang over from a previous draft.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+
+The existing label `virtualCluster` will be deprecated.  A new label `virtual_cluster` will be added.  It will carry
+the virtual cluster's name.
+
+### Use `kroxylicious_downstream_errors` use to record other downstream errors
+
+Issues should as downstream TLS negotiating errors or failure to resolve virtual cluster result in the connection being
+closed, however, there is no metric counting those errors. `kroxylicious_downstream_connections_attempts` and
+`kroxylicious_downstream_errors` should be incremented for these case too.  `virtual_cluster` and `node_id` won't be known
+so these labels should be omitted.
```
SamBarker, line 216:
```suggestion
Issues such as downstream TLS negotiating errors or failure to resolve virtual cluster result in the connection being
closed, however, there is no metric counting those errors. `kroxylicious_downstream_connections_attempts` and
`kroxylicious_downstream_errors` should be incremented for these case too.  `virtual_cluster` and `node_id` won't be known
so these labels should be omitted.
```

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+## Proposal
+
+## Summary
+
+The following metrics will be added.  They will each be described in more detail later in this section.
+
+| Metric                                           | Type         | Labels                                                                                | Description                                                                             |
+|--------------------------------------------------|--------------|---------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------|
+| kroxylicious_downstream_kafka_message_total      | Counter      | virtual_cluster, node_id†, flowing, api_key, api_version, opaque                      | Incremented by one every time a request comes from/response goes to the downstream.     |
+| kroxylicious_upstream_kafka_message_total        | Counter      | virtual_cluster, node_id†, flowing, api_key, api_version, opaque, originating_filter† | Incremented by one every time a request goes to/response comes from the upstream.       |
```
tombentley, line 95: I'm sceptical about sometimes including labels. I think it forces people to have to write more complicated post-processing to extract the measurement they actually want. Here when `originating_filter` is present you're measuring a completely different thing. I don't think people are often going to want an aggregate which included the filter-originated request with the client-originated ones. To me it makes much more sense to embed the originator in the metric name.

* `proxy_to_server_request_total`
* `filter_to_server_request_total`
* `server_to_proxy_response_total`
* `server_to_filter_response_total`

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+| kroxylicious_downstream_kafka_message_size_bytes | Distribution | virtual_cluster, node_id†, flowing, api_key, api_version  opaque                      | Records the message size of every message coming from/response going to the downstream. |
+| kroxylicious_upstream_kafka_message_size_bytes   | Distribution | virtual_cluster, node_id†, flowing, api_key, api_version, opaque, originating_filter† | Records the message size of every message coming from/response going to the upstream.   |
+| kroxylicious_message_process_time                | Distribution | virtual_cluster, node_id†, flowing, api_key, api_version, opaque, originating_filter† | Records the time taken for the message to traverse the proxy                            |
```
tombentley, line 98: * `client_to_server_request_transit_time`
* `filter_to_server_request_transit_time`
* `server_to_client_response_transit_time`
* `server_to_filter_to_response_transit_time`

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+
+(† signifies labels that will be omitted in some circumstances. The circumstances are explained below).
+
+### New counter `kroxylicious_downstream_kafka_message_total`
+
+`kroxylicious_downstream_kafka_message_total` is incremented every time a message passes to/from the downstream.
+It is discriminated by `virtual_cluster`, `node_id`, `flowing`, `opaque`, api_key`, and `api_version` labels.
+
+`node_id` will only be present if for requests going to or responses coming from a broker.  For bootstrap interactions,
+`node_id` will be absent.
```
tombentley, line 131: It being absent is not very self-describing. Perhaps we could use `node_id=bootstrap` in this case.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+| kroxylicious_downstream_kafka_message_size_bytes | Distribution | virtual_cluster, node_id†, flowing, api_key, api_version  opaque                      | Records the message size of every message coming from/response going to the downstream. |
+| kroxylicious_upstream_kafka_message_size_bytes   | Distribution | virtual_cluster, node_id†, flowing, api_key, api_version, opaque, originating_filter† | Records the message size of every message coming from/response going to the upstream.   |
+| kroxylicious_message_process_time                | Distribution | virtual_cluster, node_id†, flowing, api_key, api_version, opaque, originating_filter† | Records the time taken for the message to traverse the proxy                            |
+| kroxylicious_downstream_backpressure_state       | Gauge        | virtual_cluster, node_id†                                                             | Records the backpressure state from the downstream                                      |
```
tombentley, line 99:
```suggestion
| kroxylicious_downstream_backpressure_state       | Gauge        | virtual_cluster, node_id†                                                             | Records the backpressure state being applied to the downstream                                      |
```

But more importantly, backpressure is a per-connection thing, but this is a Gauge, and there is no label for the client's identity, so I'm confused.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+### New counter `kroxylicious_upstream_kafka_message_total`
+
+`kroxylicious_upstream_kafka_message_total` is incremented every time a message passes to/from the upstream.
+It is discriminated by `virtual_cluster`, `node_id`, `flowing`, `opaque`, `api_key`, `api_version` and `originating_filter` labels.
+
+`node_id` will only be present if for requests going to or responses coming from a broker.  For bootstrap interactions,
+`node_id` will be absent.
+
+`originating_filter` is only used for requests generated by the filters themselves (`sendRequest` API)
+and their responses.  It will be populated with the filter's name. Otherwise `originating_filter` will be absent. `opaque`
```
tombentley, line 150: I know in the API we use the term opaque. But we should remember that people operating the proxy are not going to know that terminology. Using `decoded` would be more self-explanatory.

proposals/003-metric-improvenents.md
SamBarker: Downstream TLS is an awkward case as I'm not sure we have the appropriate call backs or events to track it specifically but I think errors from downstream TLS and the [net filter](https://github.com/kroxylicious/kroxylicious/blob/79dd894bf4b69a184e5cc0dc72570ef8cb6a367d/kroxylicious-runtime/src/main/java/io/kroxylicious/proxy/internal/KafkaProxyInitializer.java#L281-L284) bubble up as exceptions and are caught and fed into the existing `kroxylicious_downstream_errors` counter via the ProxyChannelStateMachine. 

Issues where the netFilter selects something we can't connect to should also be fed into an error counter already.

proposals/003-metric-improvenents.md
```diff
@@ -0,0 +1,262 @@
...
+the message arrived from the network. The end time will be the time the response reaches the originating filter (i.e. the future completes).
+
+### New gauges `kroxylicious_(up|down)stream_backpressure_state`
+
+Reports the backpressure state for upstream and downstream.  The use-case for metric is to allow the user to
+understand when the proxy must pause reading owing whilst asynchronous work completes (for example, in record validation
+requesting a schema or in record encryption, generating a new DEK).
+
+This gauge will use values `0` and `1` only.  `1` will signify that the proxy has stopped reading the channel whilst
+it awaits the completion of asynchronous work.  `0` signifies that there is no pending asynchronous.
```
tombentley, line 197: As mentioned previously there is no "the channel" for this metric.

SamBarker (approved):
The implementation schedule sounds right to me.

Thanks Keith

proposals/003-metric-improvenents.md
SamBarker: >Firstly, it is only kroxylicious_upstream_kafka_rpc_message that will instrument async requests and their responses so only that metric will have the possibility if the originating_filter label. Async requests never traverse the downstream side of the proxy.

(only clarifying to make sure everyone is on the same page re streams and directions)

I don't believe there is **any** traversal or filter chains that are specific to `upstream` or `downstream`. A message from a client traverses the filter chain from the `downstream` to `upstream` ends , a message from the broker traverses from `upstream` to `downstream` but its the same chain.  If we have a filter chain of `A,B,C` a client message goes `A -> B -> C` and a broker message goes `C -> B -> A`. Client messages are typically requests and broker messages are typically responses. So if `B` uses send request in response to a client message the request will go via `C` and the response will return to `B` via `C`. So if 
>Async requests never traverse the downstream side of the proxy.
means that an async request never gets sent to the client then yes we agree, but if you mean it doesn't reach filters on the downstream side of the proxy then we disagree. 

I think of the proxy as having downstream and upstream sides (I find calling them client and server highly problematic), but there are not separate downstream and upstream filter chains or actions.

proposals/003-metric-improvenents.md
k-wall: Yes, i think this is better.  I'll make a change.

proposals/003-metric-improvenents.md
k-wall: I like filter_to_server

k-wall: I like the idea if separated metrics for requests sent async by the filter, however filter_to_server not too clear.  A user writing a filter that intercepts requests might expect f2s to tick.

I think we need a need that communicates that these metrics relate to the filter asynchronous request.

filter_async_to_server is naff.

(Thinking...)

proposals/003-metric-improvenents.md
k-wall: Having a client_id, principal or channel as a label would threaten a cardinality explosion.

The use case for this is to get a broad understanding of how often the proxy is toggling autoread.  You are right, it won't be sufficient to identify root cause but it should be sufficient to tell an investigation if they are on the right track.  I'd expect a user with an auto read problem to turn to tracing or filter specific metrics to delve deeper.

proposals/003-metric-improvenents.md
k-wall: > I'm not sure we are thinking about the same backpressure systems. I was thinking about the channelWritabilityChanged changed events or their ProxyChannelStateMachine equivalents as driving the gauge. That gives us insight into when we get back pressure at a TCP level. Your description is operating at a different level, that of the proxy itself.

Ah, we are indeed talking about different notations.   I was thinking about the toggling of autoread that we when a filter request/response filter is not yet complete.   You are talking about the socket level.  I think both are useful.  Gut feel is that conflating them is not the right thing to do - they are at different levels of the abstraction.   Your comment makes me realise that I've hijacked the term back pressure for my use-case.  autoread case needs a better name.

Your comment made be think about https://github.com/kroxylicious/kroxylicious/issues/2208.  I wondered if Netty itself exposed metrics around channel writability and maybe autoread.  From a very quick look at the [meters](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyMeters.java), it appears no. 

My second thought was we ought to be looking at what other projects that use Netty do with metrics and talking to some of the Netty folks and picking their brains before decide what it makes sense to measure.

I've come around to the view that we should push pause on the `back pressure` metrics and leave those for future proposal. 

In the meanwhile, if someone has a performance issue `kroxylicious_message_process_time` should help us narrow the problem space.

SamBarker: Let's defer back pressure

proposals/003-metric-improvenents.md
k-wall: > think errors from downstream TLS and the [net filter](https://github.com/kroxylicious/kroxylicious/blob/79dd894bf4b69a184e5cc0dc72570ef8cb6a367d/kroxylicious-runtime/src/main/java/io/kroxylicious/proxy/internal/KafkaProxyInitializer.java#L281-L284) bubble up as exceptions and are caught and fed into the existing kroxylicious_downstream_errors counter via the ProxyChannelStateMachine

I don't see that behaviour.  A test has been written to show that's not the case.
https://github.com/kroxylicious/kroxylicious/pull/2230

proposals/003-metric-improvenents.md
k-wall: I've added an Out of Scope section and marked both TCP/IP backpressure and Auto Read state as deferred.

proposals/003-metric-improvenents.md
k-wall: I decided I could live with _filter_to and _to_filter.  The description will make it clear how they are used.

proposals/003-metric-improvenents.md
k-wall: We made backpressure out of scope.

proposals/003-metric-improvenents.md
SamBarker: Thanks Keith! Looking at #2230 I realised the error in my thinking 🤔 The initialiser is handling this before the state machine kicks in.

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Proposal
+
+### Summary
+
+The following metrics will be added.  They will each be described in more detail later in this section.
+
+| Metric                                              | Type         | Labels                                                      | Description                                                                                                                        |
+|-----------------------------------------------------|--------------|-------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
+| kroxylicious_client_to_proxy_request_total          | Counter      | virtual_cluster, node_id, api_key, api_version, decoded     | Incremented by one every time a **request** arrives at the proxy from the downstream (client).                                     |
```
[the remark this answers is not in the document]
SamBarker: I'm 100% on board if everyone is happy with the prefix, I was floating the thought in case someone else thought it important.

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+| kroxylicious_upstream_connections_total   | Counter | virtual_cluster, node_id | virtualCluster    | Incremented by one every time a request comes from/response goes to the downstream. |
+| kroxylicious_upstream_error_total         | Counter | virtual_cluster, node_id | virtualCluster    | Records the time taken for the message to traverse the proxy                        |
+
+The labels are defined as follows:
+
+| Label           | Definition                                                                                                                                                                |
+|-----------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| virtual_cluster | Name of virtual cluster.  Label omitted if the virtual cluster has not been established‡.                                                                                 |
+| node_id         | Numeric node_id of the broker being communicated with. the value `bootstrap` is used for bootstrapping interactions. Label omitted if the node has not been established‡. |
```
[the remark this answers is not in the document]
SamBarker: Because as I said in the original comment we don't currently virtualise the node id, so it could be considered misleading.

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+number of bytes traversing the proxy.
+
+In the case where a filter causes a request (or request) to expand, this will be evident from the metrics. `kroxylicious_client_to_proxy_request_size_bytes`
+will record the size when the request arrived and `kroxylicious_proxy_to_broker_request_size_bytes` will record its new
+encoded size as it leaves.  The same thing will happen from responses.   This will allow the effect of adding filters
+such as the Record Encryption (where produce request sizes will change) to be understood.
+
+### New distributions `kroxylicious_client_to_broker_request_transit_time` and `kroxylicious_broker_to_client_response_transit_time`
+
+These metric records the length of time a message (request or response) has taken to transit the proxy.
```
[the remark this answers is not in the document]
SamBarker: No not called out here, called out in the docs for our metrics.

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+* *1 / *2 - we have contradictory meanings for the `flowing` label.  in *2, requests flow `UPSTREAM`, responses flow `DOWNSTREAM`.
+  In *1 requests flow `DOWNSTREAM` which makes no sense.
+* *3 - downstream errors does not record account for errors such as TLS negotiation errors or failures to determine virtual cluster or broker.  
+
+### General short-comings/weaknesses
+
+* The metrics endpoint populates `HELP` for each metric, but this is populated with nothing more than the metric's name.
+  We ought to be giving a clear description of the metric.
+* Some of the label names don't follow Prometheus naming conventions.  Prometheus recommends snake case whereas we have used camel case in some cases.
+* Some of the metrics names aren't too clear.  For instance, the word payload has defined meaning in the [Kafka Protocol Guide](https://kafka.apache.org/protocol).
```
robobario, line 67: I can't spot `payload` mentioned in the protocol doc linked, there are a couple of mentions in the kafka doc though:

```
design.html
565:    Compaction also allows for deletes. A message with a key and a null payload will be treated as a delete from the log. Such a record is sometimes referred to as a <i>tombstone</i>. This delete marker will cause any prior message with that key to be removed (as would any new
```

and a few usages of payload in the kafka clients code, so yeah it's overloaded

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+The following metrics wil have changes to their labels.
+
+| Metric                                    | Type    | Labels                   | Deprecated Labels | Description                                                                         |
+|-------------------------------------------|---------|--------------------------|-------------------|-------------------------------------------------------------------------------------|
+| kroxylicious_downstream_connections_total | Counter | virtual_cluster, node_id | virtualCluster    | Incremented by one every time a request comes from/response goes to the downstream. |
+| kroxylicious_downstream_error_total       | Counter | virtual_cluster, node_id | virtualCluster    | Incremented by one every time a request goes to/response comes from the upstream.   |
+| kroxylicious_upstream_connections_total   | Counter | virtual_cluster, node_id | virtualCluster    | Incremented by one every time a request comes from/response goes to the downstream. |
+| kroxylicious_upstream_error_total         | Counter | virtual_cluster, node_id | virtualCluster    | Records the time taken for the message to traverse the proxy                        |
```
robobario, line 124: the descriptions of these look wrong.

* the `*error_total` descriptions don't mention errors. 
* the `*connections_total` descriptions talk about requests, but I guess it's supposed to count established connections

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+
+### New counters `kroxylicious_<A>_to_<B>_request_total` and `kroxylicious_<A>_to_<B>_response_total`
+
+Counts the number of request/responses flowing between point A and point B.
+
+Use cases that were previously served by the deprecated metrics `kroxylicious_inbound_downstream_messages` and
+`kroxylicious_inbound_downstream_decoded_messages` can now use these metric instead.
+
+In the case where a filter **short-circuits** a request and provides its own response, `client_to_proxy_request_total` will tick
+but `kroxylicious_proxy_to_broker_request_total` will not.
```
robobario, line 157: I wonder if we should consider a specific counter, or `short_circuit` label on `kroxylicious_proxy_to_client_response_total`. Something users can look at to convince themselves this accounts for the discrepancy in and out.

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+| kroxylicious_broker_to_filter_response_total        | Counter      | virtual_cluster, node_id, api_key, api_version, filter_name | Incremented by one every time a **response** for a **filter-initiated request** arrives at the proxy from the upstream (broker).   |
+|                                                     |              |                                                             |                                                                                                                                    |
+| kroxylicious_filter_to_broker_request_size_bytes    | Distribution | virtual_cluster, node_id, api_key, api_version, filter_name | Records the message size of every **filter-initiated request** goes from the proxy to the upstream (broker).                       |
+| kroxylicious_broker_to_fiter_response_size_bytes    | Distribution | virtual_cluster, node_id, api_key, api_version, filter_name | Records the message size of every **response** for a **filter-initiated request** arrives at the proxy from the upstream (broker). |
```
robobario, line 111:
```suggestion
| kroxylicious_broker_to_filter_response_size_bytes    | Distribution | virtual_cluster, node_id, api_key, api_version, filter_name | Records the message size of every **response** for a **filter-initiated request** arrives at the proxy from the upstream (broker). |
```

proposals/003-metric-improvements.md
ShubhamRwt: `the *connections_total descriptions talk about requests, but I guess it's supposed to count established connections` 

Iam not sure but doesn't this increment even if the connection fails?(Based of my understanding)

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Summary
+
+The following metrics will be added.  They will each be described in more detail later in this section.
+
+| Metric                                              | Type         | Labels                                                      | Description                                                                                                                        |
+|-----------------------------------------------------|--------------|-------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
+| kroxylicious_client_to_proxy_request_total          | Counter      | virtual_cluster, node_id, api_key, api_version, decoded     | Incremented by one every time a **request** arrives at the proxy from the downstream (client).                                     |
+| kroxylicious_proxy_to_broker_request_total          | Counter      | virtual_cluster, node_id, api_key, api_version, decoded     | Incremented by one every time a **request**† goes from the proxy to the upstream (broker).                                         |
```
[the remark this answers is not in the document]
ShubhamRwt: +1

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+| api_key         | API key of the kafka message e.g. PRODUCE                                                                                                                                 |                                                                                                                                               |
+| api_version     | Numeric API version of the message.                                                                                                                                       |                                                                                                                                               |
+| decoded         | Flag indicating if the proxy decoded the message or not (1 = decoded, 0 = not decoded)                                                                                    |                                                                                                                                               |
```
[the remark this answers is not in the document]
k-wall: I was being influenced by https://www.robustperception.io/booleans-logic-and-math/

proposals/003-metric-improvements.md
k-wall: > 565:    Compaction also allows for deletes. A message with a key and a null payload will be treated...

ha - that's wrong.   They actually mean `value`.

proposals/003-metric-improvements.md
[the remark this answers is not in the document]
tombentley: Sgtm

proposals/003-metric-improvements.md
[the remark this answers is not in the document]
k-wall: > What about requests which don't complete? I.e. they fail with an exception hmm maybe they will be caught when we close the channel.

yeah - I can imagine there are some gaps.  We are not thinking about filters that fail (or filter that close the channel) at the moment.

> I think we should document this as a elapsed time rather than active processing time. I.e if a request stalls while waiting for a KEK to refresh.

we do impose a timeout on async filter work already (20s) io/kroxylicious/proxy/internal/KafkaProxyFrontendHandler.java:626

I was imagining ticking the metric as the upstream/downstream write finishes.  I'm not sure how I'd use metrics to capture a growing elapsed time.


> Should make sure we have the Pause time metric available and call it out here.

By pause time, do you mean the auto-read thing?

proposals/003-metric-improvements.md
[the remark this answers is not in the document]
k-wall: @ShubhamRwt @SamBarker WDYT?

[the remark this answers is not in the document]
SamBarker: `kroxylicious_client_connections_total` I feel like I'm losing this fight, but I think `client_connections` is ambiguous is it clients connecting to the proxy or the proxy connecting to the upstream broker? `downstream` and `upstream` are ambiguous.

proposals/003-metric-improvements.md
k-wall: I've recorded this in the out-of-scope section.

proposals/003-metric-improvements.md
k-wall: Yes, there is a typo (request vs connection) and it is under specified.

It is a `connection_total` so I guess the intent was for it to count all connections, even those that are short lived due to error.  @SamBarker was that the intent?

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+Records the sizes of request/responses flowing between point A and point B.
+
+Use cases that were previously served by the deprecated metric `kroxylicious_payload_size_bytes` can now use this metric
+instead. The new metric records the size of both **opaque** and **decoded** messages allowing the user to understand the
+number of bytes traversing the proxy.
+
+In the case where a filter causes a request (or request) to expand, this will be evident from the metrics. `kroxylicious_client_to_proxy_request_size_bytes`
+will record the size when the request arrived and `kroxylicious_proxy_to_broker_request_size_bytes` will record its new
+encoded size as it leaves.  The same thing will happen from responses.   This will allow the effect of adding filters
+such as the Record Encryption (where produce request sizes will change) to be understood.
```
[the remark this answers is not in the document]
k-wall: I've noted this in the proposal.

proposals/003-metric-improvements.md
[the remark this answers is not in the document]
SamBarker: >I was imagining ticking the metric as the upstream/downstream write finishes. I'm not sure how I'd use metrics to capture a growing elapsed time.

Yes, that is going to capture elapsed time isn't it? 

1. Request arrives
2. timer starts
3. reaches record encryption filter
4. KEK not found in cache - cache load triggered
5. request waits for KEK to load
6. kek loads
7. request encrypted
8. request passes the rest of the filter chain
9. request written to socket
10. timer stopped

The time tracked is the time that elapsed between the request arriving and leaving, but its not the time that the proxy actually spent working on the request.

proposals/003-metric-improvements.md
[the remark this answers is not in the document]
k-wall: You are right, I was conflating the two things in my mind.  I can't find any best practice for the domain of boolean labels, so I think you are right.  string true and false will more clear.

proposals/003-metric-improvements.md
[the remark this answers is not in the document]
k-wall: I could live with those suggested by @robobario

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,312 @@
...
+| kroxylicious_payload_size_bytes                  | Distribution | flowing, virtualCluster, ApiKey, ApiVersion | Incremented with RPC's frame size for every _decoded_ RPC from either the _downstream or upstream_. (*2)            |
+| kroxylicious_downstream_connections              | Counter      | flowing, virtualCluster                     | Incremented by one each time a connection arrives from the downstream                                               |
+| kroxylicious_downstream_errors                   | Counter      | flowing, virtualCluster                     | Incremented by one each time a connection fails owning to a downstream error (dropped connection) (*3)              |
+| kroxylicious_upstream_connections_attempts       | Counter      | flowing, virtualCluster                     | Incremented by one each time a connection attempt to made to the upstream.                                          |
+| kroxylicious_upstream_errors                     | Counter      | flowing, virtualCluster                     | Incremented by one each time a connection attempt fails owning to a upstream error (tls negotiation etc)            |
```
robobario, line 53:
```suggestion
| kroxylicious_inbound_downstream_message          | Counter      | flowing, virtualCluster                     | Incremented by one every time every inbound RPC arriving _from the downstream only_. (*1)                           |
| kroxylicious_inbound_downstream_decoded_messages | Counter      | flowing, virtualCluster                     | Incremented by one every time every inbound RPC arriving _from the downstream that the proxy needs to decode_. (*1) |
| kroxylicious_payload_size_bytes                  | Distribution | flowing, virtualCluster, ApiKey, ApiVersion | Incremented with RPC's frame size for every _decoded_ RPC from either the _downstream or upstream_. (*2)            |
| kroxylicious_downstream_connections              | Counter      | virtualCluster                              | Incremented by one each time a connection arrives from the downstream                                               |
| kroxylicious_downstream_errors                   | Counter      | virtualCluster                              | Incremented by one each time a connection fails owning to a downstream error (dropped connection) (*3)              |
| kroxylicious_upstream_connections                | Counter      | virtualCluster                              | Incremented by one each time a connection is established to the upstream and the proxy is ready to forward data.    |
| kroxylicious_upstream_connection_attempts        | Counter      | virtualCluster                              | Incremented by one each time a connection attempt to made to the upstream.                                          |
| kroxylicious_upstream_connection_failures        | Counter      | virtualCluster                              | Incremented by one each time a connection attempt terminates exceptionally.                                         |
| kroxylicious_upstream_errors                     | Counter      | virtualCluster                              | Incremented by one each time an upstream connection terminates exceptionally, whether it's connecting or not        |
```

A couple are were missing

proposals/003-metric-improvements.md
```diff
@@ -0,0 +1,312 @@
...
+| decoded         | Flag indicating if the proxy decoded the message or not (true = decoded, false = not decoded)                                                                               |                                                                                                                                               |
+| filter_name     | Name of the filter that initiated the request                                                                                                                               |                                                                                                                                               |
+
+**#3** This happens only in the case of connection to gateways using "SNI Identifies Node" where the connection fails TLS negotiation or the SNI name is unrecognised.
+
+The following metrics will be deprecated.
```
robobario, line 158: I think this list needs to be expanded to list all existing metrics

PR #71, opened 2025-06-13
Authentication API

tombentley:


proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+    void clientAuthenticationFailure(LoginException e);
+    
+}
+```
+
+### Runtime changes
+
+1. Change the `kroxylicious-runtime` module so a `Subject` gets instantiated on initial connection.
+2. On successful completion of a TLS handshake, associate a `javax.security.auth.x500.X500Principal` with that `Subject`.
+3. On construction of the filter chain fire an internal message down the pipeline.
```
k-wall, line 70: The goal of the proposal said:

> Enable `Filters` to access a client's identity using a single, consistent API, irrespective of what authentication mechanism is implemented, TLS or SASL,

I see how point 2 satisfies the TLS identity.  What catching the SASL identity and adding a subject for it?

k-wall:
With the audit use-case in mind, I can't help but think about IP address.  IP address might be from network stack or the Proxy Protocol.

I'm not trying to extend the scope of this piece of work, but I wonder how we'd model it in case it alters direction.  @rgodfrey expressed an idea to model IP address as a Principal.  If we had that, the IP address could be added to to the Subject.

Alternatively, we could have method(s) on the FilterContext to retrieve the IP.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+                return null;
+            }
+            else {
+                return null; // TODO return an error reponse: Errors.TOPIC_AUTHORIZATION_FAILED
+            }
+        }
+
+        private static boolean authorized(Subject subject) {
+            // TODO this is not abstracted from the type of Principal
+            //   We really want a model like Kafka's org.apache.kafka.server.authorizer.Authorizer
```
k-wall, line 102: I think what you are proposing is good enough.  We can document which Principals the core of Kroxylicious will use.   That will be good enough for 99% of use-cases.

k-wall:
Nice work, @tombentley. A couple of questions/comments left.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+```java
+package io.kroxylicious.proxy.filter;
+
+import javax.security.auth.Subject;
+import javax.security.auth.login.LoginException;
+
+// This is a new interface in the api module
+interface ClientSubjectAwareFilter extends Filter {
+    // Called when a client authenticates, or reauthenticates
+    void onClientAuthentication(Subject clientSubject, FilterContext context);
```
SamBarker, line 47: Presumably this method will be invoked multiple times for a given client session. i.e. once when a TLS principle is supplied, then again when a SASL auth request completes.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+    // Called when a client fails authentication, or reauthentication
+    default void onClientAuthenticationFailure(LoginException exception, FilterContext context) { 
+    }
```
SamBarker, line 50: Is it a implementations job when overriding this to invoke `FilterContext.clientAuthenticationFailure`? or are you expecting the runtime to just send the same failure to each filter in the chain?

I'm wondering how we could/would support a filter noticing that a `onClientAuthenticationFailure` had happened and generating a new temporary token. Can it then make a out of band request to authenticate with that new token? When that completes does it call `clientAuthenticationSuccess`? Does it make sense in this scenario for the proxy to have told every other filter in the chain about the same failed authentication event?

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+
+* by implementing `ClientSubjectAwareFilter` and reusing the `clientSubject`, it's possible to add a username principal to any existing `X500Principal` added to the object by the runtime.
+* `org.apache.kafka.common.security.plain.PlainLoginModule` (and probably other Kafka `LoginModule` implementations) don't work with how JASS `LoginModules` were originally designed to support stackable modules. For example it adds the `username` as a _credential_ of the `Subject`, rather than adding it as _principal_, and `login` is hard coded to return `true`.
+
+
+### Design choices
+
+* Use JAAS because:
+    - the Kafka broker and clients already use JAAS
+        - we'd rather avoid adding a dependency on those not-publicly-supportewd Kafka classes in kroxylicious-api and kroxylicious-runtime, but
```
SamBarker, line 202: Typo:
```suggestion
        - we'd rather avoid adding a dependency on those not-publicly-supported Kafka classes in kroxylicious-api and kroxylicious-runtime, but
```

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+        - we recognise that 3rd part Filter authors might want to make that choice
+    - we have no appetite to build-out our own API, nor to pick up a dependency on a 3rd party framework
+* Use `Subject` to convey identity information, rather than `Principal` because 
+    - this is more sympathetic with the conceptpaul model of JAAS.
+    - it allows attaching multiple `Principals` to client `Subjects`, and to be consumed by `ClientSubjectAwareFilter` (e.g. know the client TLS certificate DN _and_ the SASL SCRAM-SHA username).
+
```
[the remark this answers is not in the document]
SamBarker: That's an interesting suggestion @k-wall but that ties us into the Java Security manager which is [going away](https://openjdk.org/jeps/486) and I think has nasty implications for performance.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+## Compatibility
+
+This change would be backwards compatible for Filter developers and proxy users (i.e. all existing proxy configurations files would still be valid).
+
+## Rejected alternatives
+
+* Why not just add `public Principal clientPrincipal();` to `FilterContext`?
+    - It doesn't support multiple principals.
+* OK, so why not just add `public Subject clientSubject();` to `FilterContext`?
+    - It makes the `Subject` a property of the connection/Netty channel. The public API presented here allows a `Filter` to change the `Principals` presented to subsequent `Filters` in the chain, enabling Filters to implement use cases like impersonation.
```
SamBarker, line 222: This assertion kinda makes me feel like we should be defining some standard or expected principals. Such as a KafkaPrincipal to represent the principal which has actually been authenticated by the broker.

SamBarker (commented):
I like the core of the proposal but I think we need to get into few more of the details around what principals are actually expected to be available and how multiple `ClientSubjectAwareFilter`s interact.

proposals/003-authentication-api.md
robobario, line 47: Would it be too complicated to make this method async, in case they need to do some async work on authentication?

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+```
+
+Add the following two methods to `FilterContext` to allow SASL terminating `Filter` implementations to propagate their authentication outcomes to upstream filters:
+
+```java
+public interface FilterContext {
+
+    void clientAuthenticationSuccess(Subject subject);
```
robobario, line 59: Would this synchronously invoke all the other filters in the chain with their onClientAuthentication*?

I'm wondering if it would help the Framework control the order-of-events if we make it something you can tack on to the result builder instead like:

`context.requestFilterResultBuilder().withClientAuthenticationSuccess(subject).forwardRequest(headers, response)`

This might make the users testing life easier, instead of having to check if some void method was invoked by the filter, they can assert the properties of the result object.

Also, it might help head off bugs (features?) like repeated calls to the auth methods, calling both success/fail methods, calling them at unexpected times from other threads etc.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
tombentley: I think the problem with that idea is there in general there can be two subjects in play. The subject we obtain by authenticating the client, and the subject we obtain by authenticating the broker (when mutual authentication mechanisms are used). 

I'm also not aware of a lot of prior art that uses Subject#current which people might want to reuse.

proposals/003-authentication-api.md
tombentley: Yes, that was the idea.

tombentley: And I guess _again_ on reauthentication, but I've not thought overly hard about that yet.

tombentley: It's got a filter contents, so I guess it can run something async, and complete on the future internally. Since the runtime doesn't require a result from this method I don't see why we should leap through hoops to allow a filter to prevent continuation of the processing by the runtime while it does something which only it (the plugin) can be waiting for the result of.

proposals/003-authentication-api.md
tombentley: Yes, you'd need a plugin implementation to do SASL sniffing. But it wouldn't be in the runtime.

tombentley: It would be nice if we had a mechanism at proxy start up to validate the non-emptiness of intersection of set of principal types emitted by plugins and the set of principal types required by `ClientSubjectAware` plugins.

proposals/003-authentication-api.md
robobario: If my filter is interested in one but not the other, how do I discriminate? In your ExampleIdentityConsumingFilter the second call with SASL creds would be installed in the field blowing away the X500 principal. Does the filter have to inspect the principles looking at the Principal types to figure out what's relevant, like "i've received an X500 principal so I know this is from TLS"?

I'm wondering if we should expose something like an enum declaring the source like, `CLIENT_PROXY_TLS_HANDSHAKE`, `SASL_PASSTHROUGH_INTERCEPT`, `SASL_TERMINATION`. Or describe the layer it's happening at, `TLS`, `KAFKA_PROTOCOL_SASL`? To make it clearer to plugins what they are intercepting.

Maybe it could go on that context object? `ClientAuthenticationContext#layer()`?

proposals/003-authentication-api.md
robobario: ah so the user says I'm interested in SASL identities, so I'll install a SASL sniffer early on in the chain?

proposals/003-authentication-api.md
tombentley: > If my filter is interested in one but not the other, how do I discriminate? In your ExampleIdentityConsumingFilter the second call with SASL creds would be installed in the field blowing away the X500 principal. Does the filter have to inspect the principles looking at the Principal types to figure out what's relevant, like "i've received an X500 principal so I know this is from TLS"?

I don't think there's much we can do from an API PoV about that. Different impls would have different requirements about how to handle multiple principals. It's up to them to decide whether to have a field for the subject, or one(s) for the principals, and to decide how to handle reauth.

> I'm wondering if we should expose something like an enum declaring the source

SubjectAware filters may definitely care about what the _mechanism_ is. But I don't really want to enumerate it statically, and especially not the source. That's because it violates the principle of composition. A plugin that wants to use identity shouldn't care what provided that identity. And a plugin that provides identity shouldn't care what plugins might be interested. Anything else is spooky action at a distance.

So I was kindof assuming that TLS would use a X500Principal and SASL would use *some other type*. But that's not good enough because it leaves it to the person cofiguring the chain to know that all the things are speaking the same language. 

So I'm coming round to declaring a `record SaslPrincipal(String mechanism, String name) { }`, where mechanism would be like `plain`, `oauthbearer` or whatever. Together with the JDK's `X500Principal` those would function as a lowest common denominator which plugins could assume. We could add `Set<Class<? extends Principal>> requiredPrincipalClasses()` to plugins to get the startup validation that all the things had mechanisms in common.  Nothing would prevent plugins adding their own principal types if they wanted (in addition to, or instead of).

I'm also wondering if we should have our own immutable `Subject`-like class, just because mutable things are pretty nasty.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,223 @@
...
+* If clients are authenticating using the `SaslAuthenticate` KRPC the only way a `Filter` can know about that is by intercepting those frames.
+    - For "SASL passthrough", a `Filter` implementation could try to _infer_ the client's identity by watching for a successful `SaslAuthenticateResponse` 
+      returned by the broker. This is a technique that each identity-dependent filter in the chain could use, but each needs to reimplement the logic.
+    - For "SASL termination", a `Filter` implementation could handle the `SaslAuthenticateRequest` frames itself 
+        - If it makes its own `SaslAuthenticateRequests` (with different credentials, and/or a different mechanism) to the server then subsequent filters in the chain need to use the "SASL passthrough" technique to be aware of the broker-facing identity, and there's no way for them to know the original identity.
+        - If it doesn't (so that there is no `SaslAuthenticate` intersection with the broker) then subsequent filters in the chain have no way of knowing the client's identity.
+
```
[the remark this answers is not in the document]
k-wall: I think it worthwhile to make it explicit in the proposal that v0 is out of scope.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
k-wall: Ok, happy to drop the idea.  
 
If the user's using a library that expects Subject#callAs to have been established subject on the thread, they can do it themselves.

proposals/003-authentication-api.md
k-wall: If we provide a SaslSniffingFilter as part of this work (I think we should). We need to decide what sort of Principal it will add.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+package io.kroxylicious.proxy.authentication;
+
+import javax.security.auth.Subject;
+import javax.security.auth.login.LoginException;
+
+// Allows a plugin to opt-in to being aware of client-facing authentication outcomes
+interface ClientSubjectAware {
+    
+    /**
+     * Called when a client authenticates, or reauthenticates, with the proxy.
```
k-wall, line 85: So I'm using Kafka without TLS and without SASL (i.e. unauthenticated), will this method be called?  I think the answer should be yes (with an empty subject). I think this preferable to the filter having to assume the case where it has received no subject as the unauthenticated use-case.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+        Configuration config = new ConfigFile(URI.create("client-jaas.login.config"));
+
+        // TODO the callback handler to use depends on the context in the jaas config
+        //   so how do we know which handler to instantiate?
+        //   Kafka's SaslChannelBuilder basically does its own parsing of the jaas configuration to figure it out.
+        CallbackHandler callbackHandler = new PlainServerCallbackHandler();
+
+        try {
+            LoginContext loginContext = new LoginContext("client_auth_stack", subject, callbackHandler, config);
+            loginContext.login();
```
k-wall, line 269: aside: I _suspect_ this call may block, depending on the JAAS configuration.   I know this is just example code, but it probably ought to use its own TPE.  Adding a comment will be sufficient for the proposal.  If the code ever made it is into documentation, we'd want a async approach..

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+    Subject proxySubject();
+}
+```
+
+### Proposed API for initiating a SASL exchange with the broker 
+
+There are two mutually exclusive cases to consider for SASL initiation:
+
+1. That a SASL initiator plugin should perform SASL authentication _eagerly_, as soon as the underlying connection is ready. 
+This would be driven by the runtime following a successful TCP handshake on the client-to-proxy connection, and after `ClientSubjectAware` plugins have initially been called with any `X500Principal`, but before any SASL exchange on the client side.
```
k-wall, line 432: It would need to occur after the ApiVersions exchange, right?

k-wall (approved):
Very thorough work, @tombentley, well done.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+* When the proxy configured to require TLS client certificates from clients and validates these against a set of trusted signing certificates (CA certificate) it is performing **client mutual TLS authentication** ("client mTLS").
+* A **TLS server certificate** is a TLS certificate for the server-side of a TLS Handshake. As above, there could be two of these for a given connection through a proxy.
+* When the proxy is configured to use a TLS client certificate when making a TLS connection to a server, we will use the term **server mutual TLS authentication_** ("server mTLS").
+
+Now let's talk about SASL. In the following the word "component" is generalising over filters, other plugins, and the proxy as a whole:
+
+* a component which forwards a client's `SaslAuthenticate` requests to the server, and conveys the responses back to the client, is performing **SASL Passthrough**.
+* SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that a proxy's principal is the same as the broker's principal for all clients connecting through that proxy.
```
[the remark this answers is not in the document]
tombentley: Did you have a suggestion for how to improve the wording? Or are you disagreeing with this definition of identity preserving?

proposals/003-authentication-api.md
tombentley: On the one hand it feels right to announce the status of a connection as early as possible, and to all the plugins (thus allowing any one of them the opportunity to prevent the connection), but on the other hand that results in filters needing tricky logic to know when the connection is fully authenticated (because of the possible combinations for authentication configuration). In other words, I was thinking it would be like this:

* no TLS and no SASL: Called once, 
    * by the runtime without any principal. This would also be the case if there was SASL, but without there being a plugin handling the SASL (either terminator, or passing through).
* TLS, but no SASL: Called once,
    * by the runtime with an `X500Principal`. This would also be the case if there was SASL, but without there being a plugin handling the SASL (either terminator, or passing through).
* no TLS, but with SASL: Called twice, 
    * first by the runtime without any principal, 
    * a second time (via the new API) with an additional principal for the SASL authn.
being a plugin handling the SASL (either terminator, or passing through).
* TLS + SASL: Called twice, 
    * first by the runtime with an `X500Principal`, 
    * a second time (via the new API) with an additional principal for the SASL authn.


But I don't think this is good enough. I think we should provide the following guarantees:
* that there is a clear event which gets called for SASL, whether or not SASL auth is done, and whether or not a terminator or passthrough filter is present.
* either that this method is always called at least twice, or split this method into two (one for TLS and one for SASL), and guarantee that they're both always called at least once.

The wording for that second guarantee is itself quite tricky, because I don't think there's anything _in the protocol itself_ which forces SASL authentication to happen immediately after `ApiVersions`. We need to be careful not to make promises in our API which could is invalidated by changes in the protocol.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+```java
+package io.kroxylicious.proxy.filter;
+
+public interface FilterContext {
+
+    // ... existing methods ...
+
+   
+    /**
+     * Allows a filter (typically one which implements {@link SaslAuthenticateRequestFilter})
```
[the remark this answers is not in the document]
tombentley: Implementing `SaslAuthenticateResponseFilter` would not be needed in the case of SASL termination. In any case, I don't think we need the Javadoc to be completely prescriptive with respect to which requests a filter that uses this method needs to implement (which is why I only said 'typically').

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+     * @param subject The authenticated subject.
+     */
+    void clientAuthenticationSuccess(Subject subject);
+
+    /**
+     * Allows a filter (typically one which implements {@link SaslAuthenticateRequestFilter})
+     * to announce a failed authentication outcome to subsequent plugins.
+     * @param exception An exception describing the authentication failure. 
+     */
+    void clientAuthenticationFailure(LoginException exception);
```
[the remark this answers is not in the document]
tombentley: I can see there could be use cases for something like that. For example banning or rate limiting connection attempts after a small number of failed authentications with a particular user name. But:
* that sort of thing is out of scope of this proposal and might be better handled in the proxy runtime itself
* providing a principal might not be possible in all cases, and certainly isn't something we can retrofit into Java's `LoginException`.
* do we want an API that permits _unauthenticated_ `Principal` instances floating about? That means that you can only trust a `Principal` instance if you know which method it came from. If we were really serious about using type safety in that way we'd have our own `Principal` types and not let plugins instantiate them directly, but allow only the runtime to do so, via `clientAuthenticationSuccess`.

proposals/003-authentication-api.md
tombentley: You're correct that we can't rule out Callbackhandlers from blocking. The method return type is `CompletionStage<RequestFilterResult>` for that reason, but I can add a comment.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+
+and adding
+
+```java
+
+public interface FilterFactoryContext {
+
+  // ... existing methods ...
+
+  TlsCredentials tlsCredentials(Certificate certificate, PrivateKey key, Certificate[] intermediateCertificates);
```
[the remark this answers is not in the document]
tombentley: Added some words.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,516 @@
...
+
+```java
+package io.kroxylicious.proxy.authentication;
+
+import javax.security.auth.Subject;
+import javax.security.auth.login.LoginException;
+
+// Allows a plugin to opt-in to being aware of client-facing authentication outcomes
+interface ClientSubjectAware {
+
```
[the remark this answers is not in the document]
tombentley: I created a branch exploring this idea: https://github.com/tombentley/kproxy/tree/proposal-authn-api. I don't especially like the pattern of `FilterFactories` having to describe themselves through the `FilterFactoryContext` API. I guess we could achieve the same thing through annotations.

proposals/003-authentication-api.md
k-wall: Yes, I went down this train of thought too.  At first I thought about triggering the event on the first non-SASL RPC following the (optional) ApiVersions.  Like you, I got worried about the edge cases.  I retreated to a position where the event would only be sent after positive authentication events (TLS and SASL Authenticate).  I think this would be good enough for most production users (who will use Kafka in an authenticated way).  I think use cases where a filter is having to deal with authenticated and unauthenticated use-cases might be small.  Such a filter can always implement RequestFilter to surmise a channel is unauthenticated if need be.

SamBarker (commented):
I've run out of brain *damn jet lag*.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,521 @@
...
+It is not proposed to allow dynamic selection of a set of trust anchors. Those should remain under the control of the person configuring the proxy.
+
+
+#### An Example: TLS-to-TLS identity mapping
+
+This shows how a plugin could choose a TLS client certificate for the broker connection based on the connected Kafka client's TLS identity.
+
+```java
+
+class ExampleMTlsFilter implements ClientSubjectAware, ServerCredentialSupplier {
```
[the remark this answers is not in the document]
tombentley: I think what you're saying here is something like: "This class is not a `Filter`, so it's not clear to me when these methods get invoked". I was wondering about that yesterday (also from the PoV of where in the configuration file this plugin should be declared). It doesn't actually care about the protocol directly, so there's no need for it to be a `Filter` (as we've defined it until now).

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,521 @@
...
+* A **TLS server certificate** is a TLS certificate for the server-side of a TLS Handshake. As above, there could be two of these for a given connection through a proxy.
+* When the proxy is configured to use a TLS client certificate when making a TLS connection to a server, we will use the term **server mutual TLS authentication_** ("server mTLS").
+
+Now let's talk about SASL. In the following the word "component" is generalising over filters, other plugins, and the proxy as a whole:
+
+* a component which forwards a client's `SaslAuthenticate` requests to the server, and conveys the responses back to the client, is performing **SASL Passthrough**.
+* SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that a proxy's principal is the same as the broker's principal for all clients connecting through that proxy.
+* a component performing SASL Passthrough and looking at the requests and responses to infer the client's principal is performing **SASL Passthrough Sniffing**. Note that this technique does not work with all SASL mechanisms.
```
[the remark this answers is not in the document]
tombentley: I suppose you put a sniffing filter in front of a terminating filter. But it's not really clear to me what the point would be, because the sniffer would see the same principal as the terminator would produce. 

I suppose you might do something like map the SASL/PLAIN username to a different principal, so that any filters between the sniffer saw the sniffers principal, and anything after the terminator saw the terminators. But we'd need a realistic use case to justify the API catering to this possibility. And in the absence of the API modelling that kind of thing, I don't think changing the definition here improves the readability of the proposal.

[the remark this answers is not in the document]
tombentley: That's not to say we shouldn't one day implement a 1st party `SaslSniffer` filter (as @k-wall suggests). But since that's a user of this API, and not part of the API being defined in this proposal, I see that as out of scope right now.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,521 @@
...
+* a component that responds to a client's `SaslAuthenticate` requests _itself_, without forwarding those requests to the server, is performing **SASL Termination**.
+* a component that injects its own `SaslAuthenticate` requests into a SASL exchange with the server, is performing **SASL Initiation**.
```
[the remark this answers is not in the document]
tombentley: 'Identity mutating' is strongly suggestive (to me at least) that there is an object somewhere that I can mutate. That's not how this API works though, so I think it's misleading and unhelpful to introduce that terminology.

The way I think of it, 'identity preserving' is a property of the combined proxy + broker system. It depends on the proxy and brokers being configured such that they have the same (or at least a common) domain of principals, and that those principals have the same name.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,521 @@
...
+For this purpose we will add the following new interface to the new package `io.kroxylicious.proxy.authentication` in the `kroxylicious-api` module:
+
+```java
+package io.kroxylicious.proxy.authentication;
+
+import javax.security.auth.Subject;
+import javax.security.auth.login.LoginException;
+
+// Allows a plugin to opt-in to being aware of client-facing authentication outcomes
+interface ClientSubjectAware {
```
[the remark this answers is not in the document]
tombentley: > I suspect the answer is probably not, we are probably better off defining a auth callback interface and an other for routers and another for anything else in the session life cycle we find it useful for plugins to know about.

This is definitely where I am right now. This API would work with the `Router` proposal (i.e. I envisage a `Router` implementation being able to implement `ClientSubjectAware` just  like a `Filter` can under this proposal).

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,521 @@
...
+    
+    /**
+     * Called when a client authenticates, or reauthenticates, with the proxy.
+     */
+    void onClientAuthentication(Subject clientSubject, ClientAuthenticationContext context);
+    
+    /**
+     * Called when a client fails authentication, or reauthentication, with the proxy.
+     */
+    default void onClientAuthenticationFailure(LoginException exception, ClientAuthenticationContext context) {
```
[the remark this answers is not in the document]
tombentley: The `clientSubject` could be null. Consider oauth, where we've been passed some opaque token, but our interaction with the identity server fails: We never found out a meaningful user name.

We could always add this later, when someone actually has a use for it, by adding it to the `context`.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,521 @@
...
+### Proposed API for learning about client authentication outcomes
+
+Plugin implementations require an API through which to learn about client authentication outcomes.
+
+For this purpose we will add the following new interface to the new package `io.kroxylicious.proxy.authentication` in the `kroxylicious-api` module:
+
+```java
+package io.kroxylicious.proxy.authentication;
+
+import javax.security.auth.Subject;
```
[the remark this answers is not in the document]
tombentley: `Subject` has methods for getting hold of private and public credentials, so in theory we could add the `X509Certificate` as the public credential of the client's `Subject`. But that doesn't mean it's a great API. I don't love that `Subject` is mutable, nor that it's not defined what the types of Principal and credential are actually used (making for a not very type-safe API).

Yesterday I was [playing with an alternative API](https://github.com/tombentley/kproxy/tree/proposal-authn-api) which didn't use `Subject`, just `X500Principal` and a `SaslPrincipal`. That wouldn't work for your use case though.

I'm beginning to think it would be cleaner to have two interfaces: 

```java
interface ClientTlsAware {
    void onClientTlsAuthentication(
                                   @Nullable X500Principal clientPrincipal,
                                   @Nullable X509Certificate clientCertificate,
                                   ClientTlsContext context);
}
interface ClientSaslSubjectAware {
    void onClientSaslAuthentication(@Nullable SaslPrincipal clientPrincipal,
                                    ClientSaslContext context)
}
```

That frees us from providing a callback for the TLS auth failure case at all (where we haven't instantiated any filters at that point), and allows the different `context` objects to have different APIs.

proposals/003-authentication-api.md
robobario: have discussed this with Tom and I think I'm worrying about some unlikely edge cases like:
1. Filters unexpectedly invoking the methods after returning a result
2. Filters unexpectedly invoking the methods multiple times etc

that can just be handled with good javadoc on the FilterContext methods describing how/when we expect the methods to be called. It's quite deep work/knowledge required to be working with auth anyway.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
SamBarker: > I suppose you put a sniffing filter in front of a terminating filter. But it's not really clear to me what the point would be, because the sniffer would see the same principal as the terminator would produce.
> 
> I suppose you might do something like map the SASL/PLAIN username to a different principal, so that any filters between the sniffer saw the sniffers principal, and anything after the terminator saw the terminators. But we'd need a realistic use case to justify the API catering to this possibility. And in the absence of the API modelling that kind of thing, I don't think changing the definition here improves the readability of the proposal.

Isn't an audit log a reasonable use case? Downstream sniffer detects the supplied principal, upstream sniffer detects the terminated principal both get written to the log. 

> That's not to say we shouldn't one day implement a 1st party SaslSniffer filter (as @k-wall suggests). But since that's a user of this API, and not part of the API being defined in this proposal, I see that as out of scope right now.

Yeah, a first party SaslSniffer is a good idea but is a consumer of this API, so might be of illustrative use in scope but isn't a goal of this document.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
SamBarker: Hmm, that makes me think my mental model is off. In my head there is always a `clientSubject` as there is always a client. It contains a set of principals from various sources. So in your example of a failed token mapping we would have a `clientSubject` with an empty set (as we talked somewhere about only having auth'd/valid principals). 

I'm fully prepared to be wrong, but I'd like to understand where the fault in my model is.

I feel like there is state machine lurking in this proposal that would benefit from being explicitly outlined.

kidpollo:
Sorry for the delayed response. I cooccur with @dmariassy in that we would like to have the full cert available at connection and filtering time. For refference this is the interface for the connection in our prototype.

```
public interface ClientConnectionContext {

    /**
     * Gets the client's X.509 certificate if the client connection uses mTLS.
     *
     * @return the client certificate, or empty if the client is not using mTLS
     */
    Optional<X509Certificate> clientCertificate();

    /**
     * Gets the client's host address.
     *
     * @return the client host address
     */
    String clientHost();

    /**
     * Gets the client's port number.
     *
     * @return the client port number
     */
    int clientPort();

    /**
     * Gets the SNI hostname provided by the client during TLS handshake.
     *
     * @return the SNI hostname, or empty if not provided
     */
    Optional<String> sniHostname();

    /**
     * Gets the name of the client software (from ApiVersions request).
     *
     * @return the client software name, or empty if not available
     */
    Optional<String> clientSoftwareName();

    /**
     * Gets the version of the client software (from ApiVersions request).
     *
     * @return the client software version, or empty if not available
     */
    Optional<String> clientSoftwareVersion();
}
```

proposals/003-authentication-api.md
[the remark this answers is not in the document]
SamBarker: I'm not intending to suggest a different meaning to `identify preserving` I agree its a property I was just looking to define the a term to cover the other actions.

 > The way I think of it, 'identity preserving' is a property of the combined proxy + broker system. It depends on the proxy and brokers being configured such that they have the same (or at least a common) domain of principals, and that those principals have the same name.


Thats quite an expanded definition, it might be consistent with the rest of the proposal but until this comment we haven't really talked about the broker, and I don't really follow the need to expand the scope to the broker.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,522 @@
...
+## Terminology
+
+Let's define some terminology to make discussion easier. For the purposes of this proposal we consider TLS and SASL as the only authentication mechanisms we care about.
+
+Let's first define terms for TLS:
+
+* A **mutual authentication mechanism** is one which proves the identity of a server to a client.
+* When the proxy is configured to accept TLS connections from clients it is performing **TLS Termination**, which does not imply mutual authentication.
+* A **TLS client certificate** is a TLS certificate for the client-side of a TLS Handshake. For a given client and server pairing a proxy might have _two_ of these the Kafka client's TLS client certificate, and the proxy's own TLS client certificate for its connection to the server.
+* When the proxy configured to require TLS client certificates from clients and validates these against a set of trusted signing certificates (CA certificate) it is performing **client mutual TLS authentication** ("client mTLS").
```
gunnarmorling, line 15:
```suggestion
* When the proxy is configured to require TLS client certificates from clients and validates these against a set of trusted signing certificates (CA certificate) it is performing **client mutual TLS authentication** ("client mTLS").
```

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,522 @@
...
+
+## Terminology
+
+Let's define some terminology to make discussion easier. For the purposes of this proposal we consider TLS and SASL as the only authentication mechanisms we care about.
+
+Let's first define terms for TLS:
+
+* A **mutual authentication mechanism** is one which proves the identity of a server to a client.
+* When the proxy is configured to accept TLS connections from clients it is performing **TLS Termination**, which does not imply mutual authentication.
+* A **TLS client certificate** is a TLS certificate for the client-side of a TLS Handshake. For a given client and server pairing a proxy might have _two_ of these the Kafka client's TLS client certificate, and the proxy's own TLS client certificate for its connection to the server.
```
gunnarmorling, line 14:
```suggestion
* A **TLS client certificate** is a TLS certificate for the client-side of a TLS Handshake. For a given client and server pairing, a proxy might have _two_ of these: the Kafka client's TLS client certificate, and the proxy's own TLS client certificate for its connection to the server.
```

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,522 @@
...
+package io.kroxylicious.proxy.authentication;
+
+public interface ClientAuthenticationContext {
+
+    /** 
+     * The subject that the proxy presented to the client.
+     * This may be null of the authentication mechanism does not support 
+     * mutual authentication.
+     */
+    Subject proxySubject();
```
gunnarmorling, line 113: What would be use cases for this?

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,522 @@
...
+     * @param exception An exception describing the authentication failure. 
+     */
+    void clientAuthenticationFailure(LoginException exception);
+    
+}
+```
+
+It's worth noting that `LoginException` has a number of subclasses in `javax.security.auth.login`.
+
+This API allows SASL-terminating plugin to announce its authentication outcomes to later filters in the filter chain which have implemented `ClientSubjectAware`.
```
gunnarmorling, line 203:
```suggestion
This API allows a SASL-terminating plugin to announce its authentication outcomes to later filters in the filter chain which have implemented `ClientSubjectAware`.
```

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,522 @@
...
+
+By implementing `ClientSubjectAware` and reusing the `clientSubject`, we're adding a principal to any existing `X500Principal` of the subject.
+
+
+### Proposed API for selecting TLS client certificates for server connections
+
+Initiating a connection to a broker is currently entirely the responsilibity of the runtime, using the `NetFilter` interface. 
+(note that `NetFilter` is **not** part of the `kroxylicious-api` module.)
+This means that the TLS client certificate used is currently always the same, and cannot depend on the client subject.
+Adding this new API will allow TLS client certificates for conections to servers to depend on client identity (as learned about using `ClientSubjectAware`).
```
gunnarmorling, line 301:
```suggestion
Adding this new API will allow TLS client certificates for connections to servers to depend on client identity (as learned about using `ClientSubjectAware`).
```

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,522 @@
...
+```
+
+These are the mirror image of the equivalent client methods, and calling them would similarly result in `ServerSubjectAware`-implememting plugins getting informed of the server's subject.
+
+
+### Use cases
+
+* SASL-to-SASL identity mapping (in combination with a SASL terminator)
+* SASL-to-SASL identity preservation with in-proxy authorization (in combination with a SASL terminator, and an authorizer).
+* Audit logging
```
gunnarmorling, line 483: Adding another one which is my motivation for joining this discussion:

```suggestion
* Audit logging
* Projections in fetch responses based on client identity, i.e. only returning those fields of the records on a given topic a client is authorized for
```

gunnarmorling (commented):
Did a light-touch pass with a few minor suggestion. Exciting work!

proposals/003-authentication-api.md
tombentley: I'd call this "principal aware data masking", and yes, absolutely this is a use case.

proposals/003-authentication-api.md
tombentley: Partly this is just about completeness: If an authentication mechanism provides mutual authentication then it feels like something which we _ought_ to expose to plugins. 

It's not terribly useful if you're deploying a proxy in front of a single cluster. But if you had a proxy deployed as a sidecar alongside a Kafka client application then it becomes a bit more interesting. There could be uses for it if/when we added the proposed Router API. For example, if you could use it to route data about people to a cluster that's located in a particular jurisdiction, based on their nationality (e.g. compliance with data processing laws). I'll admit I'm hand waving a bit here, but it seems harmless to add this, or at least have a design that's compatible with adding it in the future, if use cases emerge.

tombentley:
Thank you for the feedback @gunnarmorling @k-wall @SamBarker @robobario @dmariassy @kidpollo. I have bad news: I've basically rewritten this proposal based on experience of trying to implement the proposed APIs. So I'd be very grateful if you could take another look. (@dmariassy @kidpollo you get your wish for the full `X509Certificate`).

The good news is that https://github.com/tombentley/kproxy/tree/proposal-authn-api has a partial implementation including:

* SASL PLAIN Terminator 
* SASL Inspector (I decided that Sniffer was probably not the ideal choice of word in an enterprise context)
* SASL PLAIN Initiator

I'm intending to keep working on that branch. I'd like to get an implementation of the `ServerTlsCredentialSupplier` working in the integration tests, just to show how these APIs can be composed for various different use cases.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+package io.kroxylicious.proxy.tls;
+
+import java.security.cert.X509Certificate;
+import java.util.Optional;
+
+public interface ClientTlsContext {
+    /**
+     * @return The TLS server certificate that the proxy presented to the client during TLS handshake.
+     */
+    X509Certificate proxyServerCertificate();
```
k-wall, line 101: Is a single certificate sufficient for the use-cases?  What if the server certificate is using intermediates?  I wonder if returning an ordered list containing the whole chain would be a better API.  There would be a guarantee that the chain would always have at least one certification (the self signed case).

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+
+    /**
+     * @return the client's certificate, or empty if no TLS client certificate was presented during TLS handshake.
+     */
+    Optional<X509Certificate> clientCertificate();
```
k-wall, line 106: ditto, the client may have sent a list of certificates when responding to server's certificate request message.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+    /**
+     * The name of the SASL mechanism used.
+     * @return The name of the SASL mechanism used.
+     */
+    String mechanismName();
+
+    /**
+     * Returns the principal returned by the server.
+     * @return the principal returned by the server,
+     * or empty if the SASL mechanism used does not support mutual authentication.
```
k-wall, line 360: I didn't realise SASL mutual auth was a thing.  What mechanisms support it?

k-wall:
@tombentley I like the API revised proposal.  It seems very thorough.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+All that's required is an API for exposing appropriate details to `Filters`.
+
+The following method will be added to the existing `FilterContext` interface:
+
+```java
+    /**
+     * @return The TLS context for the client connection, or empty if the client connection is not TLS.
+     */
+    Optional<ClientTlsContext> clientTlsContext();
+```
```
SamBarker, line 87: Which client? The client connecting to the proxy or the proxy acting as a "client" of the broker.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+     * <p>Callers should not:</p>
+     * <ul>
+     *   <li>assume any particular type of principal.
+     *   <li>assume the principal does not change during the lifetime of a filter (due to reauthentication)
+     * </ul>
+     *
+     * @return The authenticated principal for the client connection, or empty if the client
+     * has not successfully authenticated.
+     */
+    Optional<? extends Principal> clientPrincipal();
```
SamBarker, line 236: Why `principal` and not `Subject`?

I know its not in this draft but I feel we've talked a lot about there being several possible principals in the session.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+     * or it may be a {@link SaslPrincipal} if client identity is SASL based.</p>
+     *
+     * <p>Callers should not:</p>
+     * <ul>
+     *   <li>assume any particular type of principal.
+     *   <li>assume the principal does not change during the lifetime of a filter (due to reauthentication)
+     * </ul>
+     *
+     * @return The authenticated principal for the client connection, or empty if the client
+     * has not successfully authenticated.
```
SamBarker, line 234: What does this return when a client presents a client TLS cert and then something performs SASL initiation? As presumably the person configuring the proxy would have configured `SASL` as the principal type?

On that note I guess a plugin wanting to use the client TLS cert to initiate SASL would need to use the `clientTlsContext` rather than the principal. 

The timing dependent nature of this leaves me a feeling a bit concerned. I need to mull it further.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+     * Returns the authenticated principal for the server connection, or empty if the server
+     * has not successfully authenticated, or if the server authentication was not mutual.
+     * The concrete type of principal returned depends on the proxy configuration.
+     * For example,
+     * it may be a {@link javax.security.auth.x500.X500Principal} if server identity is TLS-based,
+     * or it may be a {@link SaslPrincipal} is client identity is SASL based.
+     * @return The authenticated principal for the server connection, or empty if the server
+     * has not successfully authenticated.
+     */
+    Optional<? extends Principal> serverPrincipal();
```
SamBarker, line 382: The proxy is both a client and a server! I think `serverPrincipal` is really ambiguous.

`downstreamPrincipal` and `upstreamPrincipal` seem much clearer to me or even `receivedDownstreamPrincipal` and ` sentUpstreamPrincipal`.

SamBarker (commented):
Sorry review is WIP but I've run out of time.

proposals/003-authentication-api.md
robobario: to expand, we had been discussing Role-based Access as being something useful to support, so some way of mapping the clientPrincipal to some rolePrincipals which Filters can consume.

@SamBarker I think there's value in having some way to identify this singular `clientPrincipal` that directly relates to the singular principal transmitted by the client via TLS or SASL that the user has configured as their VirtualCluster#principalType. Authorization Filters may be interested in a derived set of principals (from that originator) to test roles, while Audit filters may need to pluck out the originator principal for logging.

Maybe we could have our own Subject record that embeds that concept like:

```
record ClientSubject(Principal clientPrincipal, Set<Principal> derivedPrincipals)
```
which would contain the users desired clientPrincipals and then principals deriverd from that clientPrincipal.

proposals/003-authentication-api.md
robobario: Agree on the ambiguity, but I think we should continue to shy away from upstream/downstream. It feels like the language and what it implies is cultural/individual and has caused confusion in the past.

Maybe the pain is that we've named the entity talking to the proxy `client`. If we have multiple things acting as clients, calling one of them 'the client' is confusing.

What if we introduced a new term for the client like `user`? And the upstream server is the `target`. And the proxy is the `proxy`. Then we'd have:

```
userClientPrincipal
proxyClientPrincipal
proxyServerPrincipal
targetServerPrincipal
```

robobario: note, in this context I think `upstreamServer` would also be pretty unambiguous, it's just once we start having pervasive `upstream/downstream` language it gets harder to understand imo.

proposals/003-authentication-api.md
robobario: maybe we could allow filters to specify their own principalType in the filterDefinition? By default they'd receive the virtual cluster principalType, but could override it. 

You would have to amend the startup validations, since a VirtualCluster with principalType=SASL containing a filter that has specified principalType=X500 would need to ensure mTLS is enabled and SASL components are in the filter chain. Plus you'd want to validate that a filterfactory that consumes X509Principals can't declare it wants SASL principalType.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+
+* Configuring a virtual cluster with `principalType: SASL` without a SASL terminator or SASL inspector in the virtual cluster's `filters` (because where is the SASL principal going to come from?)
+* Configuring a virtual cluster with `principalType: X500` in a virtual cluster not configured for client mTLS (because where is the TLS principal going to come from?)
+* Configuring multiple SASL terminators and/or SASL inspectors in the `filters` of a virtual cluster (because there should be a single producer of client identity). Similarly for SASL initiators (becausde there should be a single producer of server identity).
+
+To provide this kind of fail-safe, the proxy runtime needs to know which filters are SASL inspectors, terminators or initiators, and what sort of identity information a filter consumes.
+At proxy start up time, the runtime only knows about `FilterFactories`, not about any `Filter` instances or their types, 
+The `FilterFactory` service interface doesn't provide a way for the runtime to know what kind of filters it may create.
+Therefore we will introduce the following runtime-retained annotation types to be applied to `FilterFactory` implementations:
```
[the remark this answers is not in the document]
robobario: seems like a reasonable restriction to put on Filter Authors. If you want to make a Filter that can work with the framework's clientPrincipal concept you must annotate it so we can check it drops into the Filter chain nicely.

proposals/003-authentication-api.md
robobario: closing this as outdated by the latest changes

proposals/003-authentication-api.md
robobario: marking resolved, specific principal types are now proposed in the latest update.

proposals/003-authentication-api.md
robobario: The latest update has specific Principal types defined. Marking resolved.

proposals/003-authentication-api.md
robobario: These sentences about how/when initiation occurs have dropped out. Closing as outdated.

proposals/003-authentication-api.md
robobario: Closing as outdated since the conversation about principals/subjects is ongoing in fresher threads and the event based flow has been replaced with FilterContext methods

proposals/003-authentication-api.md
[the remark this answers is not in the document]
robobario: resolving as outdated as the eventing style is replaced by proposal to use FilterContext methods

proposals/003-authentication-api.md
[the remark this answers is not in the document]
robobario: Resolving, the latest proposal gives access to the X509Certificate via the FilterContext. Moving away from evented style.

robobario:
Thanks for the work @tombentley feels like its nearly there. 

I've gone and resolved a stack of outdated/addressed old threads, please have a futher look at your open threads and see if there's more outdated content there. Since we've got so many threads I think it's worth trying to keep only alive/unaddressed threads alive and aim to resolve them all.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
tombentley: If you still want a term for this then "Identity substituting" would be my preference. I don't think I really _needed_ this in the proposal as it stands, but I can add it if you really want it there so we all have a consistent vocab. Let me know.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
tombentley: Resolving because we got rid of this interface.

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,650 @@
...
+     */
+    String authorizationId();
+
+    /**
+     * The server identity that the proxy presented to the client using SASL authentication.
+     * @return the proxy's identity with the client. This will be null
+     * if the proxy did not supply an identity because the SASL mechanism used
+     * does not support mutual authentication.
+     */
+    Optional<String> proxyServerId();
```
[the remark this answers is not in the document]
robobario: Got it. 

What I'm interested in, like can we document it to say `if you annotate your Filter with ClientPrincipalConsumer, then clientPrincipal() should be non-empty for all messages except X,Y,Z`. Is the guarantee different if you're in front of the SASL filter or after it in the chain? 

Maybe we should require the consumers to come after the producer in the chain here, `Configuring a virtual cluster with principalType: SASL without a SASL terminator or SASL inspector in the virtual cluster's filters (because where is the SASL principal going to come from?)`, just to reduce the different orderings we have to consider.

Optional emptiness implying the initial state and failed state starts looking a bit dodgy too.

proposals/003-authentication-api.md
robobario: :+1: I'm happy enough to start with `Optional<? extends Principal> clientPrincipal();`, slight preference for putting it in a little wrapper record for future extension vs having to continue pumping up the `FilterContext` with more methods in future.

proposals/003-authentication-api.md
robobario: hmm, maybe `upstream` and `downstream` language to describe the entities is fine in this networking context, IIRC the confusion in the past was more about how we were using it in documentation to describe the direction of flow of messages through filters.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
tombentley: Gonna close this one, because the current version doesn't use Subject at all.

proposals/003-authentication-api.md
tombentley: This sentence has gone.

proposals/003-authentication-api.md
tombentley: Reworded the javadoc

proposals/003-authentication-api.md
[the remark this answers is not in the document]
tombentley: I think there would be a lot of ifs-and-buts in any language which tries to tie down the contract more exactly. 
It depends on:
* Whether the transport is TCP or TLS and whether `VirtualCluster.principalType` is `TLS` or `SASL` and 
* If the transport is TLS and `VirtualCluster.principalType` is `TLS` and  then a filter would always see non-empty.
* If `VirtualCluster.principalType` is `SASL`, whether there's any SASL terminator or inspector in the filter chain
* If `VirtualCluster.principalType` is `SASL` and there's a SASL terminator or inspector in the filter chain, a filter intercepting ApiVersions or Sasl* could see empty initially.
* After seeing non-empty it's still possible to see empty later on. This has legit use cases too. Consider SASL using OAUTHBEARER where the terminator is coded to timeout if the oath server doesn't respond (thus making a non-empty clientPrincipal empty), but the client retries a few ms later and the terminator gets a response (transition from empty to non-empty). 

It's much simpler to think of this method as returning the current state, and accept that success and failed are not terminal states in the state machine, but trust that while in the failed state protected requests won't get propagated.

> Maybe we should require the consumers to come after the producer in the chain here,

I started off thinking like that too. But it makes it impossible to write a audit logging filter which traces the SASL exchange and sees the outcome of a separate SASL terminator. That's because it would need to be in front of the terminator to see the SASL requests (because a terminator doesn't propagate those requests), but it want to consume the outcome of the terminator. You could argue that such audit logging should be built into terminators and inspectors, but that comes with its own costs.

proposals/003-authentication-api.md
tombentley: @robobario which is the worse poison: more methods, or more methods and more types to contain them and _still more_ methods to access the wrapper types? I don't think either is obviously better than the other, tbh. I did the wrapper interfaces for TLS and SASL because:

* The accessor for the type can usefully be Optional, to signify the absence of that layer (and independently of using an Optional return type to represent the absence of mutual authentication info, for example).
* We would be able to reuse those wrapper interfaces for other plugins like `Router`, without having multiple bloated `${Plugin}Context` types.

There are plenty of other ways to do it which aren't obviously better or worse. We have to make _a_ decision. I'm saying the bike shed should be blue.

k-wall:
@tombentley I reviewed your branch yesterday. I think it looks good. Thank you for the effort.  Are you thinking to land all of it in one go, or chunk it?

proposals/003-authentication-api.md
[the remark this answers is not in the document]
robobario: Okay, so the contract would be brittle/complicated and not something we could easily stand behind. Would it still be worth a less formal note like "If your Filter handles SASL client principals, in some configurations it will observe ApiVersions and Sasl exchange messages before the principal is available"?

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,647 @@
...
+
+* Configuring a virtual cluster with `principalType: SASL` without a SASL terminator or SASL inspector in the virtual cluster's `filters` (because where is the SASL principal going to come from?)
+* Configuring a virtual cluster with `principalType: X500` in a virtual cluster not configured for client mTLS (because where is the TLS principal going to come from?)
+* Configuring multiple SASL terminators and/or SASL inspectors in the `filters` of a virtual cluster (because there should be a single producer of client identity). Similarly for SASL initiators (becausde there should be a single producer of server identity).
+
+To provide this kind of fail-safe, the proxy runtime needs to know which filters are SASL inspectors, terminators or initiators, and what sort of identity information a filter consumes.
+At proxy start up time, the runtime only knows about `FilterFactories`, not about any `Filter` instances or their types, 
+The `FilterFactory` service interface doesn't provide a way for the runtime to know what kind of filters it may create.
+Therefore we will introduce the following runtime-retained annotation types to be applied to `FilterFactory` implementations:
```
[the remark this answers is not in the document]
tombentley: > if it's dropped in a chain with no authentication, no principal is logged (or we come up with some UnauthenticatedPrincipal).

It could substitute its own thing (`principal=bob` versus `anonymousPrincipal=true`) for the empty case. 

> we come up with some UnauthenticatedPrincipal

I think that runs counter to what a principal is supposed to be. It opens up some weird corner cases:
* What does it mean to have a `Subject` with both `UnauthenticatedPrincipal` and `SaslPrincipal("bob")`? 
* What's the difference between a Subject with no principals and one with just `UnauthenticatedPrincipal`?

(I know I've kept `Subject` out of this API, but I think it's pretty likely we'll end up building Subjects when we do authZ, so I think it's worth thinking about how Subjects with multiple principals would work).

> Could we, in future, have some way for them to signal through the enum that they are capable of working even without a Principal, like `@ClientPrincipalConsumer(value=Principal.class, requiresPrincipal=false)`.

It depends how you expressive you want to make it. We _could_ support things like `@ClientPrincipalConsumer(allOf={SaslPrincipal.class, X500Principal.class})`, and `@ClientPrincipalConsumer(anyOf={SaslPrincipal.class, X500Principal.class})` and `@ClientPrincipalConsumer(oneOf={SaslPrincipal.class, X500Principal.class})`. 

I don't really have a good feel if this is over engineering or not. I know orgs sometimes have very niche requirements. But I think devs generally prefer a simple model they can get their heads around easily.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
robobario: marking resolved as it's not about the design, more something that should go into the PR review when we're adding the interface

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,647 @@
...
+
+```java
+package io.kroxylicious.proxy.authentication;
+
+interface TlsCredentials {
+  /* Intentionally empty: implemented and accessed only in the runtime */
+}
+```
+
+// TODO Why not use the JDK's `X500PrivateCredential`?
```
kidpollo, line 533: seems X500PrivateCredential contains the cert so this feels fine, not sure what is the benefit

kidpollo:
Looks good to me! great job

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,647 @@
...
+
+### Protections for those configuring a virtual cluster
```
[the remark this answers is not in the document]
robobario: as I see on your javadoc comment you are saying that @ClientPrincipalConsumer has that function already

proposals/003-authentication-api.md
k-wall: Ok, we can change the API later if the need arises.

proposals/003-authentication-api.md
k-wall: I didn't have a use-case in mind.  I was being influenced by the fact that the SslSession object makes the full chain available.  As I said above, let keep the API as is.  If more requirements come to light, we can handle those then.

proposals/003-authentication-api.md
k-wall: I've actually grown to like upstream and downstream terminology.

proposals/003-authentication-api.md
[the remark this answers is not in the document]
k-wall: > TBH I think we should require the annotations, and throw from relevant methods of FilterContext is a filter uses those methods if its FilterFactory didn't have the annotations.

I see the point you are making.  I wonder how you'd keep the implementation of the enforcement simple.  Won't we need to hold references to the factory that created the FilterContext - or at least something about its annotations?

proposals/003-authentication-api.md
k-wall: Why do we still have distinct interfaces for ClientSaslContext and ServerSaslContext.  I think we can have a single SaslContext used for both the client (inspector) and server (initiator) use cases.  There is no reason I can think of to hide the `authorizationId` that the initiator has used.

SamBarker (approved):
I've be working with a different mental model for what `serverPrincipal` represented, which has only come out chatting with @k-wall and re-reading the comment threads. 

To me I think there are two principals, maybe three now I think about what is currently `serverPrincipal`. 

1. `clientPrincipal` aka `downstreamPrincipal` -> the principal/identity the *client* sent to the proxy and is thus acting as.
2.`effectivePrincipal` (what I thought `serverPrincipal` was) -> the principal/identity the *proxy* sent to the broker and thus is acting as
3. `serverPrincipal` -> the identity the *broker* sent to the proxy (which is why mutual auth matters here).

When running in pass through mode `clientPrincipal` and `effectivePrincipal` are the same identity. There are however still two logical identities. 

If we mandated a `SaslInspector` at the upstream tail of the filter chain we could ensure that the `effectivePrincipal` is always populated and potentially allows us to make `serverSaslAuthenticationSuccess` & `serverSaslAuthenticationFailure ` a private API. 

I think whats outlined in this PR will work, so I'm +0.5, but I am contending that the separatedPrincipals I mentioned above make life easier for plug-in authors.

k-wall:
For a thought experiment, we were discussing a use-case where you'd want to use the proxy to change the SASL mechanism. How would you do that with this model?

Say an organisation want to improve its security posture by mandating SCRAM-SHA. It has a large legacy of clients connecting using SASL_PLAIN. A big bang change is impractical.  It needs to phase the introduction of SCRAM, so it deploys the proxy with filter(s) capable of rewriting SASL exchanges that use PLAIN to using SCRAM.  The legacy clients using PLAIN keep working until they can be upgraded.  For clients using SCRAM, the filter would be effectively a no-op.

How would you do this with this API?   Could you implement a filter that is both a SASL Terminator and SASL Initiator?
I think the answer is yes.  You wire the filter to incept the SaslHandshake and SaslAuthenticate.  Any interactions that use SCRAM go straight through unchanged.  The filter catches the PLAIN handshake/authentication attempts and initiates its own SCRAM authentication with the Broker.  The filter returns a SaslAuthenticate response to the client with outcome based on outcome of the initiated SCRAM authentication.  The filter wouldn't need the `@ServerSaslProducer` annotation as it wouldn't need to call serverSaslAuthenticationSuccess.

If there were other filters in the chain that need to know the resulting client principal, you could an a SaslInterceptor filter to chain (at position 2).  It would hear both the passed through SCRAM SASL interactions and the initiated SCRAM SASL ones.  It would be annotated `@ClientSaslProducer`.

robobario (approved):
Thanks Tom, I'm +1

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,647 @@
...
+
+    /**
+     * Returns the client's authorizationId that resulted from the SASL exchange.
+     * @return the client's authorizationId.
+     */
+    String authorizationId();
+}
+```
+
+### APIs for using client principals in a generic way
```
robobario, line 212: Continuing @SamBarker's comment here, as I can't thread off the top-level comment

> 1. clientPrincipal aka downstreamPrincipal -> the principal/identity the client sent to the proxy and is thus acting as.
> 2.effectivePrincipal (what I thought serverPrincipal was) -> the principal/identity the proxy sent to the broker and thus is acting as

to make sure I've got it:

The `ClientPrincipal` in the proposal is a different concept again from what you've listed @SamBarker, it's a _chosen_ (by the proxy operator) principal for the downstream client. 

For example when the proxy terminate mTLS and the user says their cluster has principalType TLS, it will be the x500 certificates, regardless of whether they've had to initiate SASL with the upstream. In this case it would be the same as your `downstreamPrincipal` but could be different from the `effectivePrincipal`.

When the proxy terminates mTLS and the user says their cluster has principalType SASL which they implement with an initiator plugin, the ClientPrincipal will be the SASL principle after initiation has succeeded. In this case it is the same as your `effectivePrincipal`, but different from the raw `downstreamPrincipal`.

Similar with a SASL swapping scenario, we could have an `effectivePrincipal` different from the raw `downstreamPrincipal` and the `clientPrincipal` could be either of them depending how/where we intercept the principal.

I think we could add in the "raw" principals received and sent by the proxy later if required, it would just be a bit of an ugly clash with the chosen naming.

> I think whats outlined in this PR will work, so I'm +0.5, but I am contending that the separatedPrincipals I mentioned above make life easier for plug-in authors.

I think the proposed ClientPrincipal concept makes life easier for a class of Filters because the Filters don't have to know when to pick between the downstream principal and the proxyClient/effective principal for generic cases like authn, it's done by configuration. Also, by having principalType in terms of `SSL` and `SASL` we can introduce these extra protections at startup. I was thinking on whether the user could configure `principalType: downstream|effective`, but then you can only validate that there is some downstream auth set up, or some upstream auth mechanism. Maybe for the SASL swapping case it would be clearer to specify an upstream or downstream principal, rather than "SASL" :thinking: .

Some usages for the separate principals might be audit, I'm interested in logging out the raw downstream principal alongside the effective principle. As proposed I don't think we could get at the `downstream` SASL principal if we swapped for a different SASL principal.

proposals/003-authentication-api.md
SamBarker: Happy to defer yes

tombentley:
@sorooshme-happening thanks for chiming in!

This proposal covers authentication, but not authorization. 

To "hide" authentication you'd need a filter acting as a SASL terminator. That would handle the `SaslHandshake` and `SaslAuthenticate` requests locally in the proxy, so they won't propagate to the broker. How that appears to the broker depends on how the broker is configured. 
* If it's configured with `security.protocol=PLAIN` then the broker would see the proxy as an `Anonymous` client, which means you can't really do much authorization in the broker either.
* If it's configured with `security.protocol=SSL` then the broker will use the TLS client certificate presented by the proxy to construct its `KafkaPrincipal`. This proposal adds a mechanism for a proxy plugin to select the certificate it presents (based on the TLS connection it has with the Kafka client, or the SASL authentication with the client), so you could propagate some kind of client identity to the broker.
* If you want to configure the broker with `security.protocol=SASL_PLAINTEXT` or `SASL_SSL` then the proxy needs to perform some kind of SASL interaction with the broker. In this case there are two options. 
    * Use a SASL Initiator as well as a SASL Terminator. This decouples the client-proxy and the proxy-broker SASL exchanges. So for example you could, for example do SASL `PLAIN` with the Kafka client but `SCRAM-SHA-512` with the broker.
    * Use a SASL Inspector instead of a SASL Terminator. This means the Kafka client's `SaslHandshake` and `SaslAuthenticate` requests propagate to the broker, so the proxy and the broker will have a common idea of who the Kafka client is. (This case is not hiding authentication, of course, so it's not what you're looking for, but I mention it for completeness).

It's worth spelling out that this proposal only covers **the API** between filters and the proxy runtime. It doesn't mean the Kroxylicious project will itself provide any filters that act as SASL Terminators, SASL Inspector or SASL Initiators. There are some basic implementations of these things in the tests which could serve as inspiration. I think it's likely the project would provide implementations of these things in the fullness of time. But right now I don't want to hold back the API work.

So back to the "not authorization" part... I intend to make a proposal concerning authorization once I've implemented this proposal. Right now the idea is:

* Introduce a `SubjectBuilder` plugin interface, which allows to construct a `Subject` from the `ClientTlsContext` and/or `ClientSaslContext`. This interface should also allow looking up Principals from remote systems (e.g. `RolePrincipals` from LDAP/Active Directory).
* Have some kind of API for Filters to be able to ask "Is the current subject allowed to perform $ACTION on $RESOURCE?" (e.g. "connect to target cluster bar", or "read from topic foo")
* Actions and resources should be extensible, so someone writing a Filter can define their own types of resource and enumerate the possible actions it supports. 
* Allow end users of the proxy to write their authorization policies in a single place (such as the proxy configuration file).
* All this would have been relatively easy if Java 17 hadn't deprecated (and Java 21 removed) most of the authorization half of JAAS along with the `SecurityManager`. 

All of the above is subject to change, and I don't know yet whether it's really going to be possible to do all those things. But if you have specific use cases for authorization that you're willing to share it would great if you could open an Issue describing them.

tombentley:
@ujjwalkalia did you have any feedback on the proposal? It would great to confirm that this meets your needs.

ujjwalkalia:
> @ujjwalkalia did you have any feedback on the proposal? It would great to confirm that this meets your needs.

Hey @tombentley , yes the proposal satisfies our requirement of making the mTLS based principal available to the filter context.

tombentley:
Replaced by multiple other, more narrow PRs, #72, #73, #74, #75

proposals/003-authentication-api.md
```diff
@@ -0,0 +1,647 @@
...
+
+The APIs presented so far are sufficient to write plugins which:
+
+* Propagate SASL, letting a client's SASL identity reach the server unchanged
+* Initiate SASL, injecting mechanism-specific credentials prior to letting client-originated requests reach the server. 
+  The selection of the server-facing SASL credentials to use could be based on the client's identity 
+  (e.g. SASL termination and SASL initiation in the same virtual cluster)
+
+What's missing is an API where the server-facing TLS client certificate is chosen based on the client's identity. 
+For this purpose we will add a new plugin interface, `ServerTlsCredentialSupplierFactory`.
```
kidpollo, line 393: @robobario I was just looking at 0.14.0 and did not find a way/plugin to propagate TLS credentials to upstream. Having the ClientTlsContext in the Filter context is great! We can add our custom filtering logic that way however we would not be able to then propagate same creds upstream so that the brokers can use those creds for ACLs enforcement, throttling etc.

Is this api still being considered?

tombentley: @kidpollo that part of this proposal got factored into https://github.com/kroxylicious/design/pull/75. I'd still like to implement that, but I've been working on other things recently, so I don't have an implementation. I'd like there to be at least a partial implementation before we commit to that proposal.

PR #72, opened 2025-07-25
004-terminology-for-authentication

tombentley:
It's been suggested to break https://github.com/kroxylicious/design/pull/71/ into separate proposals to simplify review etc. This PR adds a "proposal" which is really just defining some common terminology for the project. Other proposals will refer to this one, so the terms are all defined in one place.

### DIFF
--- a/proposals/004-terminology-for-authentication.md
+++ b/proposals/004-terminology-for-authentication.md
@@ -11,7 +11,7 @@ In the following, the word "component" is generalising over filters, other plugi
 
 ### Mutual authentication
 
-A **mutual authentication mechanism** is one which proves the identity of a server to a client.
+Usually 'authentication' is the process by which a client proves their identity to a server. A **mutual authentication mechanism** is one which proves the identity of each party to the other.
 
 ### TLS Termination
 
@@ -40,11 +40,11 @@ a component which forwards a client's `SaslAuthenticate` requests to the server,
 
 ### Identity preserving
 
-SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that, for all client principals in the virtual cluster, each of those principals will have the same name as the corresponding client principal in the broker.
+SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that, for all client identities in the virtual cluster, each of those identities will have the same name as the corresponding client identity in the broker.
 
 ### SASL Passthrough Inspection
 
-a component performing SASL Passthrough and looking at the requests and responses to infer the client's principal is performing **SASL Passthrough Inspection**. 
+A component performing SASL Passthrough and looking at the requests and responses to infer the client's identity is performing **SASL Passthrough Inspection**. 
 Note that this technique does not work with all SASL mechanisms.
 
 ### SASL Termination
--- a/proposals/004-terminology-for-authentication.md
+++ b/proposals/004-terminology-for-authentication.md
@@ -36,16 +36,13 @@ When the proxy is configured to use a TLS client certificate when making a TLS c
 
 ### SASL Passthrough
 
-a component which forwards a client's `SaslAuthenticate` requests to the server, and conveys the responses back to the client, is performing **SASL Passthrough**.
-
-### Identity preserving
-
-SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that, for all client identities in the virtual cluster, each of those identities will have the same name as the corresponding client identity in the broker.
+A component which forwards a client's `SaslAuthenticate` requests to the server, and conveys the responses back to the client, is performing **SASL Passthrough**.
 
 ### SASL Passthrough Inspection
 
 A component performing SASL Passthrough and looking at the requests and responses to infer the client's identity is performing **SASL Passthrough Inspection**. 
-Note that this technique does not work with all SASL mechanisms.
+Note that this technique does not work with all SASL mechanisms. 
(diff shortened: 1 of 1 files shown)


proposals/004-terminology-for-authentication.md
```diff
@@ -0,0 +1,115 @@
...
+It is the first of a number of proposals which will use this terminology.
+
+For the purposes of this proposal we consider TLS and SASL as the only authentication mechanisms we care about.
+In the following, the word "component" is generalising over filters, other plugins, and a proxy virtual cluster as a whole.
+
+## Definitions
+
+### Mutual authentication
+
+A **mutual authentication mechanism** is one which proves the identity of a server to a client.
```
robobario, line 14: Is it missing a bit like, and proves the identity of a client to a server? Think theres a 'mutual' aspect missing.

proposals/004-terminology-for-authentication.md
```diff
@@ -0,0 +1,115 @@
...
+
+When the proxy is configured to use a TLS client certificate when making a TLS connection to a server, we will use the term **server mutual TLS authentication_** ("server mTLS").
+
+### SASL Passthrough
+
+a component which forwards a client's `SaslAuthenticate` requests to the server, and conveys the responses back to the client, is performing **SASL Passthrough**.
+
+### Identity preserving
+
+SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that, for all client principals in the virtual cluster, each of those principals will have the same name as the corresponding client principal in the broker.
```
robobario, line 43: Worth defining `principal`? Like, when/where is a principal established? How does it differ from credentials or tokens transmitted from the client?

tombentley: The reason I commented it out is because JAAS doesn't really feature in any of the other proposals now. That's intentional, because I don't think we want to lean on JAAS -- certainly not at the API level -- for authz. And if we're not leaning on it for authz then there's not really any need to use `Principal` in the API AFAICS. The same information will be available from the `ClientTlsContext` and/or `ClientSaslContext`, just not packaged as a `Principal`. If we're no longer letting SASL terminators or inspectors propagate a `Principal`, then any use of `Principals` can be a Filter-local decision. 

In case anyone asks the question "Why _not_ lean on JAAS for authz", here are some reasons:

* [JEP 411](https://openjdk.org/jeps/411) deprecates the `SecurityManager` and a host of other Authz stuff such as `Policy` from Java 17. [JEP 486](https://openjdk.org/jeps/486) removes those APIs from Java 24. This means there's nothing in the JRE which will actually evaluate an access control decision. 
* We _could_ make use of `Principal `, `Subject` and `Permission` in our APIs and write the access control evaluation ourselves, but I'm right I'm doubtful if that would be the right thing to do. Would it scale to the number of resources we're likely to encounter? We'd also need a replacement for `Policy`, i.e. the representation/configuration of who can do what to what.
* We should consider how to let authz be pluggable. If it's going to be pluggable then why adopt a representation of principal in our API between runtime and filter that uses JAAS types, if the implementation isn't JAAS?
* There are other interesting authz technologies to consider, such a Zanzibar, which has an open source implementation in the form of [SpiceDb](https://github.com/authzed/spicedb).

For these reasons I think it's wise to just pass on JAAS, at least for what we're trying to do right now. 

I'll invent some words next week to define Principal in a JAAS-independant way, or remove talk of principals entirely.

robobario (approved):
LGTM

proposals/004-terminology-for-authentication.md
```diff
@@ -0,0 +1,115 @@
...
+
+When the proxy is configured to use a TLS client certificate when making a TLS connection to a server, we will use the term **server mutual TLS authentication_** ("server mTLS").
+
+### SASL Passthrough
+
+a component which forwards a client's `SaslAuthenticate` requests to the server, and conveys the responses back to the client, is performing **SASL Passthrough**.
+
+### Identity preserving
+
+SASL Passthrough is one way to for a proxy to be **identity preserving**, which means that, for all client identities in the virtual cluster, each of those identities will have the same name as the corresponding client identity in the broker.
```
SamBarker, line 43: Is it worth defining what we/Kafka mean by identity? 

We can't actually ensure that we offer this definition. In the face of opaque tokens we have no way of knowing what the identity in the proxy is or if is the same as the broker holds. So is there a practical difference between `SASL Passthrough` and `Identity Preserving`?

SamBarker (approved):
LGTM 

I think starting with terms is a great place to begin.

k-wall (approved):

proposals/004-terminology-for-authentication.md
```diff
@@ -0,0 +1,115 @@
...
+
+### SASL Passthrough Inspection
+
+A component performing SASL Passthrough and looking at the requests and responses to infer the client's identity is performing **SASL Passthrough Inspection**. 
+Note that this technique does not work with all SASL mechanisms.
```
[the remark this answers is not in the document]
tombentley: Clarified with an extra sentence (but without reference to any specific mechanism).

proposals/004-terminology-for-authentication.md
tombentley: I dropped "identity preserving" from the doc. I don't think we really use this term in any of the other documents, so there's no benefit from trying to define it.

FTR I guess there are two orthogonal ideas going on. The definition I wrote required the proxy to have inferred some kind of identity (i.e. assumed passthrough inspection). But the wider idea is that when we're passing through the proxy is not substituting any credentials of its own. I.e. there is only one "domain of users".

PR #75, opened 2025-07-25
plugin api to select tls credentials for server connection

tombentley:
It's been suggested to break https://github.com/kroxylicious/design/pull/71 into separate proposals to simplify review etc. This PR adds a proposal for a plugin API for runtime selection of the TLS certificate to be used on the proxy-to-server connection.

### DIFF
--- a/proposals/011-plugin-api-to-select-tls-credentials-for-server-connection.md
+++ b/proposals/011-plugin-api-to-select-tls-credentials-for-server-connection.md
@@ -1,7 +1,7 @@
 # Plugin API for selecting TLS client credentials for proxy-to-server connection
 
 This proposes a new plugin API for allowing the selection of the TLS client credentials to be used for the connection between the proxy and a Kafka server.
-It makes use of [proposal-004](proposal-004) for the terminology used.
+It makes use of [proposal-004](./004-terminology-for-authentication.md) for the terminology used.
 
 ## Current situation
 


proposals/007-plugin-api-to-select-tls-credentials-for-server-connection.md
```diff
@@ -0,0 +1,178 @@
...
+
+
+## Affected/not affected projects
+
+The `kroxylicous` repo.
+
+## Compatibility
+
+This change would be backwards compatible for `Filter` developers and proxy users (i.e. all existing proxy configurations files would still be valid).
+
```
robobario, line 177: Could add

```
# Future Work

* Enable using identity from a SASL Terminator in upstream TLS certificate selection.
```

Maybe this is a good spot to explain why it doesn't work initially.

proposals/007-plugin-api-to-select-tls-credentials-for-server-connection.md
```diff
@@ -0,0 +1,178 @@
...
+## Proposal
+
+
+### API for selecting target cluster TLS credentials
+
+We will add a new plugin interface, `ServerTlsCredentialSupplierFactory`.
+It will use the usual Kroxylicious plugin mechanism, leveraging `java.util.Service`-based discovery.
+However, this plugin is not the same thing as a `FilterFactory`.
+Rather, an implementation class will be defined on the `TargetCluster` configuration object, and instantiated once for each target cluster.
+The TargetCluster's `tls` object will gain a `tlsCredentialSupplier` property, supporting `type` and `config` properties (similarly to how filters are configured).
```
robobario, line 30: Sidebar: not for this proposal. Wonder where we would put this in the operator API. In operator land KafkaService has the upstream tls details, but we can reuse a KafkaService across virtual kafka clusters. Would we put it on KafkaService, so users need multiple KafkaServices if they want different selection logic per virtual cluster. Or put it on VirtualCluster, with the downside that we spread out the upstream TLS settings across two CRs.

proposals/007-plugin-api-to-select-tls-credentials-for-server-connection.md
```diff
@@ -0,0 +1,178 @@
...
+
+```java
+package io.kroxylicious.proxy.authentication;
+
+interface TlsCredentials {
+  /* Intentionally empty: implemented and accessed only in the runtime */
+}
+```
+
+// TODO Why not use the JDK's `X500PrivateCredential`?
```
robobario, line 167: In the end did you choose not to do this because of how in-flux JAAS looked?

tombentley: Although `X500PrivateCredential` has a cert and private key, it doesn't directly model intermediate certificates. Instead is has an `alias` property, which refers to an entry in some `KeyStore` that the instantiator and user of the `X500PrivateCredential` instance have to mutually have a reference to (which is left to the application to sort out).

So using it would involve some costs:
* extra hoop jumping in terms of passing around a KeyStore
* and also any credentials needed to access the entry within the key store. 
 
Meanwhile the only benefit I can see right now is "It's a JDK type". 

So over all I don't think it's worth it.

proposals/007-plugin-api-to-select-tls-credentials-for-server-connection.md
```diff
@@ -0,0 +1,178 @@
...
+
+         /**
+          * Creates some TLS credentials for the given parameters.
+          * @param key The key corresponding to the given client certificate.
+          * @param certificateChain The client certificate corresponding to the given {@code key}, plus any intermediate certificates forming the certificate chain up to (but not including) the TLS certificate trusted by the peer.
+          * @return The TLS credentials instance.
+          * @see ServerTlsCredentialSupplier.Context#tlsCredentials(PrivateKey, Certificate[])
+          */
+         TlsCredentials tlsCredentials(PrivateKey key,
+                                       Certificate[] certificateChain);
```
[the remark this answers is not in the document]
tombentley: Good spot. Fixed.

proposals/007-plugin-api-to-select-tls-credentials-for-server-connection.md
```diff
@@ -0,0 +1,180 @@
...
+    interface Context {
+
+        /**
+         * An executor backed by the single Thread responsible for dispatching
+         * work to a ServerTlsCredentialSupplier instance for a channel.
+         * It is safe to mutate ServerTlsCredentialSupplier members from this executor.
+         * @return executor
+         * @throws IllegalStateException if the factory is not bound to a channel yet.
+         */
+        ScheduledExecutorService filterDispatchExecutor();
```
SamBarker, line 66: Should we call this `pluginDispatchExecutor`?

proposals/007-plugin-api-to-select-tls-credentials-for-server-connection.md
```diff
@@ -0,0 +1,180 @@
...
+import java.security.PrivateKey;
+import java.security.cert.Certificate;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+import io.kroxylicious.proxy.authentication.ClientSaslContext;
+
+/**
+ * Implemented by a {@link io.kroxylicious.proxy.filter.Filter} that provides
+ * the credentials for the TLS connection between the proxy and the Kafka server.
```
SamBarker, line 107:
```suggestion
 * Implemented by a class annotated with {@link io.kroxylicious.proxy.plugin.Plugin} that provides
 * the credentials for the TLS connection between the proxy and the Kafka server.
```

I don't think the intention is for filters to implement this, is it?

SamBarker (approved):
A few references to Filter that I thought we were trying to avoid, using plugin instead, but otherwise LGTM

kidpollo (approved):
LGTM

robobario (approved):
LGTM, thanks @tombentley

kidpollo:
Sample implementation https://github.com/kidpollo/kroxylicious-tls-demo

PR #77, opened 2025-08-15
007 - Add topic name lookup facility

robobario:
This proposal outlines a new feature for the Kroxylicious filter framework: a mechanism to resolve topic names from their corresponding topic IDs. This will enable filters to implement business logic based on human-readable topic names, rather than the underlying Kafka implementation detail of topic IDs.

### DIFF
--- a/proposals/007-topic-name-lookup-facility.md
+++ b/proposals/007-topic-name-lookup-facility.md
@@ -1,7 +1,7 @@
 # Proposal 007: Resolve Topic Names from Topic IDs in the Filter Framework
 
 <!-- TOC -->
-* [Proposal 008: Resolve Topic Names from Topic IDs in the Filter Framework](#proposal-008-resolve-topic-names-from-topic-ids-in-the-filter-framework)
+* [Proposal 007: Resolve Topic Names from Topic IDs in the Filter Framework](#proposal-007-resolve-topic-names-from-topic-ids-in-the-filter-framework)
   * [Summary](#summary)
   * [Current Situation](#current-situation)
   * [Motivation](#motivation)


proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+ * @return A CompletionStage that will complete with an unmodifiable Map<Uuid, String>,
+ * mapping each topic UUID to its topic name. The stage will complete
+ * exceptionally if any UUID in the set cannot be resolved.
+ */
+CompletionStage<Map<Uuid, String>> getTopicNames(Set<Uuid> topicUuids);
+```
+
+This method will internally send a `Metadata` request to the upstream Kafka cluster to fetch the required topic information.
+
+The operation will fail atomically to ensure filters do not operate on a partial or incomplete mapping of IDs to names.
```
robobario, line 52: Not sure about this one, should we instead describe which ones we couldn't map in the result.

robobario:
Here's a rough prototype https://github.com/kroxylicious/kroxylicious/pull/2583

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+
+### 2. Composability
+
+A core value of Kroxylicious is filter composability. Other filters with virtualization capabilities (e.g., a multitenancy filter) may need to alter the topic names presented to downstream filters.
+
+To support this, the underlying `Metadata` request generated by this new method will flow through the standard filter chain. This allows any filter to intercept and mutate the `MetadataResponse`, ensuring seamless composition with existing filters like `Multitenant` without requiring any changes to them.
+
+### 3. Performance and Caching
+
+To avoid overwhelming the upstream cluster with frequent `Metadata` requests, the responses from this lookup should be cached. Equally we do not wish to impede client and other internal requests, we
```
k-wall, line 62: How long will the cache be valid for?  What effect would a client's metadata request have on the cache?

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+
+I propose:
+1. adding a new method to the `FilterContext` interface that allows filters to look up topic names for a given set of topic IDs.
+2. This will drive a request to the upstream to learn the topic names. 
+3. We will cache the topic names at the edge of the proxy/upstream per channel.
+4. Existing Filters that mutate MetadataResponse will compose with this API without modification.
+
+### 1. API Addition
+
+The following method will be added to the `FilterContext`:
```
k-wall, line 36: It seems odd to me that you can go from topic uuid to name but not the other way round.  Is there something that makes you confident that the queries will always go that way?

proposals/008-topic-name-lookup-facility.md
k-wall: I _wonder_ if we should start off with a solution that doesn't cache, and add caching later if there is user demand for it.  Users can always cache the results themselves, if they wish.

robobario: yep, very possible to implement it without caching first.

Mm for cache invalidation we should probably also react to RPCs that fail due to non-existent topic, like produce/consume.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+
+```java
+/**
+ * Asynchronously resolves a set of topic UUIDs to their corresponding topic names.
```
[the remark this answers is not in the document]
robobario: The former, we would deal in immutable maps.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+
+This method will internally send a `Metadata` request to the upstream Kafka cluster to fetch the required topic information.
+
+The operation will fail atomically to ensure filters do not operate on a partial or incomplete mapping of IDs to names.
+
+### 2. Composability
+
+A core value of Kroxylicious is filter composability. Other filters with virtualization capabilities (e.g., a multitenancy filter) may need to alter the topic names presented to downstream filters.
+
+To support this, the underlying `Metadata` request generated by this new method will flow through the standard filter chain. This allows any filter to intercept and mutate the `MetadataResponse`, ensuring seamless composition with existing filters like `Multitenant` without requiring any changes to them.
```
robobario, line 58: A central thing on my mind is whether using the Filter API is the right direction. It feels like a natural way to access the data that is available via the Kafka APIs and means that intervening virtualization filters don't have to do anything special, they mutate kafka RPCs as they would anyway. 

But, if you wanted to generalize the mechanism to something outside of Kafka's knowledge and say "physical kafka resources have some external metadata that we want to retrieve from within the filter chain" this falls apart.

With topic selectors on the near horizon as well, I wonder if we will be building quite a different API to run selector requests/responses through the filter chain and if it's worth considering a unified API for retrieving "features" or "metadata" of physical kafka entities.

Maybe that's fine to have one way for the built-in metadata and another for external. 

The alternative Tom came up with in the past was something like a ResourceRequest API, I guess addressing just this, interfaces for querying aspects of the physical resources.

robobario: As discussed in call, having two different mechanisms which a Filter would have to implement to mutate the topic metadata would be error prone. Easy to overlook or have the mechanisms diverge/differ.

tombentley: I think we should stick to propagating a `Metadata` request though the pipeline to the broker. 

The `ResourceRequest` API still works, but the key point is that it's not querying for data that's held in Kafka, so what it returns can't be inconsistent with what's observable to the filter via Metadata. Of course it's still possible for the external data source to be inconsistent wrt topic existence. But topic deletion is eventually consistent from the PoV of a runtime client API caller, because clients have a metadata cache, so that doesn't seem as bad compared with getting inconsistent results via different API calls _at the same time_.

proposals/008-topic-name-lookup-facility.md
tombentley: I agree that we should see how it behaves without caching before adding a cache. 

However, one thing to consider is a chain of `N` filters which sequentially handle the same response `R` containing a UUID:

1. Filter `N` handles `R`, sees unknown UUID `x`, requests metadata, gets metadata response forwards response `R` to the client
2. Filter `N-1` handles `R`, sees unknown UUID `x`, requests metadata, gets metadata response forwards response `R` to the client
3. Filter `N-2` handles `R`, sees unknown UUID `x`, requests metadata, gets metadata response forwards response `R` to the client
4. ...You get the idea

To avoid this I think we could safely cache the Metadata response at least until we actually send the response to the client. 

However, it's also quite possible the client also doesn't know `x` (more likely for larger clusters, where there's a higher probability that it found out `x` via a different connection). So slightly better would be to make use of `R.correlation_id` and keep the cached metadata response until we've handled request `R.correlation_id+k`, for some small `k`. Clients can pipeline requests, so while I think `k=1` is (I think) perfectly safe, it might be beneficial to make k=5 (IIRC the default max inflight requests), or make it even smarter by accounting for the number of inflight requests we already know about. (We're returning `R` for correlation_id `c`, the highest correlation id we've observed for a request from this client is `c+y`, so keep the metadata respose till we've handled request `c+y+1`. That should catch the case where request `c+y+1` is the client's metadata request for `x`. Might also want an upper bound on time-staleness too, just to be sure.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+
+### 3. Performance and Caching
+
+To avoid overwhelming the upstream cluster with frequent `Metadata` requests, the responses from this lookup should be cached. Equally we do not wish to impede client and other internal requests, we 
+should forward them to the upstream.
+
+I propose marking these framework-generated requests as a special class of "edge-cacheable" requests. This tells the framework that it is safe to learn the topic names and produce short-circuit responses without calling out to the upstream cluster, reducing redundant network traffic.
```
tombentley, line 65: Not sure what this means TBH.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+should forward them to the upstream.
+
+I propose marking these framework-generated requests as a special class of "edge-cacheable" requests. This tells the framework that it is safe to learn the topic names and produce short-circuit responses without calling out to the upstream cluster, reducing redundant network traffic. 
+
+The 'marking' can be implemented by tagging the RequestFrame and propagating it through the correlation manager. Then we can
+configure a caching filter to only intercept requests and responses with that tag. This would be a framework only feature, not surfaced in the Filter API.
+
+The 'edge' aspect is saying that we want to cache what the upstream said the topic names are, at the edge of the Proxy -> Upstream, but they must then traverse the Filter chain to maintain composability.
+
+We can implement a very long lived cache invalidation strategy as Kafkatopics cannot be renamed. Inconsistencies will arise if the upstream topics are deleted but we should be able to rely on the upstream failing if we forward any interactions with deleted topics.
```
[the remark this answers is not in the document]
tombentley: I, too, don't want to think about this. But we should.

Kafka already has a mechanism for deferred deletion. I don't recall if that's just soft deleting the data, with the topic deleted from the metadata immediately (such that you can immediately recreate the topic once you have the deletion response success), or whether it imposes a delay before you can recreate the topic. If the latter then knowing the timeout would at least give us an upper bound on how long to leave it before revalidating caches.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+
+### 4. Security
+
+Kafka ACLs may restrict a user's access to topic metadata. We must not introduce a security vulnerability that leaks information about topics to unauthorized users.
+
+Therefore, the initial implementation of the cache will be **per-channel**. This approach scopes the cache 1:1 with an upstream connection and its authenticated principal, ensuring that a user can only receive metadata they are authorized to see. In the future, we could explore a virtual-cluster-level cache, but this would require users to explicitly opt-in after being warned of the potential security trade-offs.
```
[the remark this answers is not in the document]
tombentley: > hard to achieve.

except in the case where the proxy is the sole arbiter of authZ.

tombentley (commented):
Thanks @robobario

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
+# Proposal 008: Resolve Topic Names from Topic IDs in the Filter Framework
+
+## Summary
+
+This proposal outlines a new feature for the Kroxylicious filter framework: a mechanism to resolve topic names from their corresponding topic IDs. This will enable filters to implement business logic based on human-readable topic names, rather than the underlying Kafka implementation detail of topic IDs.
+
```
k-wall, line 6: Might be worth referencing the KIP that gradually driving the introduction of topic-ids.

proposals/008-topic-name-lookup-facility.md
k-wall: > To avoid this I think we could safely cache the Metadata response at least until we actually send the response to the client.

I like this idea.

> So slightly better would be to make use of R.correlation_id and keep the cached metadata response until we've handled request R.correlation_id+k, for some small k

It is a clever idea but it sounds tricky.  Do we really need it?   My gut says wait until we see a use-case where the problem bites, then add the clever code.

proposals/008-topic-name-lookup-facility.md
robobario: I'm trying to express that instead of caching mappings for every Filter in the chain, which may have be exposed to different names/ids due to upstream manipulation, _something_ can cache what has been returned by the target cluster at the "edge" between the proxy and cluster.

When a Filter requests identifiers, it would traverse the filter chain as a Metadata request/response, and this component close to the target cluster could short-circuit respond if it knows all of the names/ids requested. It means multiple filters can benefit from a single cache, but has a cost since you always have to go via the Filter chain.

But I also think to minimise impact on clients should only be applied to these special out-of-band requests, downstream client Metadata requests should flow to the target cluster without being subject to this special behaviour.

robobario: Another strategy could be having the framework handle caching per FilterHandler with the caching activated only once the Filter makes it's first request for a topic name or id.

proposals/008-topic-name-lookup-facility.md
robobario: updated to include a Result class as the Map value that can describe the name XOR exception.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+---
+
+## Proposal
+
+I propose:
+1. adding a new method to the `FilterContext` interface that allows filters to look up topic names for a given set of topic IDs.
+2. This will drive a request to the upstream to learn the topic names. 
+3. We will cache the topic names at the edge of the proxy/upstream per channel.
+4. Existing Filters that mutate MetadataResponse will compose with this API without modification.
```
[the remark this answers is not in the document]
robobario: have reworded

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param topicUuids A Set of topic UUIDs to resolve.
+ * @return A CompletionStage that will complete with an unmodifiable Map<Uuid, String>,
+ * mapping each topic UUID to its topic name. The stage will complete
+ * exceptionally if any UUID in the set cannot be resolved.
+ */
+CompletionStage<Map<Uuid, String>> getTopicNames(Set<Uuid> topicUuids);
+```
+
+This method will internally send a `Metadata` request to the upstream Kafka cluster to fetch the required topic information.
```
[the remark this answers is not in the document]
robobario: I had imagined only querying the topic ids that we are immediately interested in. I guess doing them all at once would save incrementally building the cache up.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Asynchronously resolves a set of topic UUIDs to their corresponding topic names.
+ *
+ * @param topicUuids A Set of topic UUIDs to resolve.
+ * @return A CompletionStage that will complete with an unmodifiable Map<Uuid, String>,
+ * mapping each topic UUID to its topic name. Every input topicUuid must have an entry
+ * in the result map.
+ */
+CompletionStage<Map<Uuid, TopicNameResult>> getTopicNames(Set<Uuid> topicUuids);
```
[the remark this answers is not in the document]
robobario: I was thinking more of a term for the fetching/getting/sending behaviour. But with caching as a future extension I think `topicNames()` is fine so that  we don't imply too much about the mechanism.

proposals/008-topic-name-lookup-facility.md
```diff
@@ -0,0 +1,152 @@
...
+/**
+* topicName XOR exception will be non-null
+*/
+record TopicNameResult(@Nullable String topicName, @Nullable TopicMappingException exception){}
+
+/**
+ * Asynchronously resolves a set of topic UUIDs to their corresponding topic names.
+ *
+ * @param topicUuids A Set of topic UUIDs to resolve.
+ * @return A CompletionStage that will complete with an unmodifiable Map<Uuid, String>,
```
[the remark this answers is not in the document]
robobario: You mean for the stage being completed exceptionally? I can see that the existing sendRequest can complete it with a `TimeoutException`.

Or do you mean we should define some well-understood `TopicMappingException`s in the `TopicNameResult`. Maybe `UNKNOWN_TOPIC_OR_PARTITION` and `UNKNOWN_TOPIC_ID` are worth having a distinct exception type like `UnknownTopicException` as they are so predictable. Or we could have a generic `KafkaServerErrorException()` that could convey any `org.apache.kafka.common.protocol.Errors`.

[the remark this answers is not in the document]
robobario: have pushed up a description of a generic `KafkaServerErrorException`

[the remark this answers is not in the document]
robobario: hmm maybe there should also be a type defined as a catch-all for other failures of the underlying `sendRequest`.

We should probably also javadoc the same guarantee as the `sendRequest`:
```
     * <h4>Chained Computation stages</h4>
     * <p>Default and asynchronous default computation stages chained to the returned
     * {@link java.util.concurrent.CompletionStage} are guaranteed to be executed by the thread associated with the
     * connection. See {@link io.kroxylicious.proxy.filter} for more details.
```

proposals/008-topic-name-lookup-facility.md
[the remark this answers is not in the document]
robobario: caching is moved off to a future extension so resolving this for now

tombentley:
@robobario assuming https://github.com/kroxylicious/design/pull/75 doesn't get merged first, I suppose you will need to renumber the filename.

k-wall (approved):

PR #78, opened 2025-08-27
Add azure kms proposal

robobario:


proposals/007-azure-kms.md
```diff
@@ -0,0 +1,148 @@
...
+`
+
+To minimise EDEK bytes there are a couple of opportunities when encoding the EDEK:
+1. we can exclude the vaultBaseUrl and have the limitation that we work only with a single key vault
+2. we can encode just the keyName and keyVersion, and use that to decrypt
+3. all documented examples of the keyVersion (and from my experimentation too) are 128bits encoded as a hexadecimal string.
+
+So for the EDEK I think we could optimistically try to encode it as `versionByte,keyNameLength,keyName,keyVersionLength,128-bit-decoded,edek`
+and fall back to using the string like `versionByte,keyNameLength,keyName,keyVersionLength,keyVersionString,edek`. So
+we would have either a 16-byte version implying it's the bytes, or a 32 character string.
```
robobario, line 117: we could maybe save another byte by having one EDEK version for the 16byte version where we don't need to encode the size. and another EDEK version for the fallback, where we encode version as an arbitrary length string. (it's docced as 32 characters but doesn't say they have to be ascii)

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Proposed Initial Implementation
+
+1. Support all flavours of Key Vault. The APIs will be the same, just with a different vault base URI.
+2. Support RSA and HSM-RSA key types, wrapping using `RSA-OAEP-256` but emit a warning that it is not quantum-resistant.
+3. Support HSM-AES key type and AES-GCM wrapping
```
robobario, line 123: on this, I think the EDEK should encode which key type we used, so that we can use the appropriate algorithm at key unwrapping time (Which will be 1:1 with whether the key type is RSA or AES)

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,148 @@
...
+    * [TLS](#tls)
+    * [DEK Generation](#dek-generation)
+    * [Key Identifiers](#key-identifiers)
+  * [Proposed Initial Implementation](#proposed-initial-implementation)
+    * [Configuration](#configuration)
+<!-- TOC -->
+
+## Background
+
+> Azure Key Vault is a cloud service for securely storing and accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, or cryptographic keys.
```
k-wall, line 23: Like AWS, Microsoft do an on-premise version of Azure deployable to your own datacentre.

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,148 @@
...
+
+The Azure SDK then takes advantage of the fact this happens to be the base URL for various operations like wrap/unwrap.
+
+ed. for [wrap key](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/wrap-key/wrap-key?view=rest-keyvault-keys-7.4&tabs=HTTP) the url is `POST {vaultBaseUrl}/keys/{key-name}/{key-version}/wrapkey?api-version=7.4
+`
+
+To minimise EDEK bytes there are a couple of opportunities when encoding the EDEK:
+1. we can exclude the vaultBaseUrl and have the limitation that we work only with a single key vault
+2. we can encode just the keyName and keyVersion, and use that to decrypt
+3. all documented examples of the keyVersion (and from my experimentation too) are 128bits encoded as a hexadecimal string.
```
k-wall, line 113: I'd be wary of relying on the results of experimentation.  If Azure change the naming scheme we risk a user blowing up.    Unless we can get a statement from Microsoft saying it is 128bits and hex, I think we need to be safe are treat it as a string.  Annoying.  

We had the same situation in AWS, the keyid looks like a UUID but they are documented as Strings.  We treat them as Strings. Grrr.

k-wall:
Looks like it is heading in a good direction.

proposals/007-azure-kms.md
robobario: what I've got in the prototype is I try to optimistically treat it as a 16-byte hex string and fall back to arbitrary-string, so:
* if we can decode the id as hex and it produces 16 bytes, we serialize a compact EDEK with version byte 0
* else serialize a stringy EDEK with version byte 1

I also think we can use a single byte for the string length as it's docced as being 32 characters, I don't think it can blow out to more than 127 bytes even if it contained all obscure unicode characters.

proposals/007-azure-kms.md
robobario: Another thing you can discover up front is whether `unwrapKey` and `wrapKey` are supported by the key, users are able to do things like create an RSA key that doesn't support unwrap/wrap. 

And you can determine whether the key is enabled.

proposals/007-azure-kms.md
robobario: or you could come up with some scheme to do the same thing within a single message version. like have a 'versionType' field that tells you how to interpret the bytes.

but yeah we have to be very careful, I'd only want to apply this if we can guarantee fidelity. Like Azure says the object ids are case insensitive, but I'd be wary of encoding a hex string containing uppercase alpha characters as we lose the case information.

robobario:
Something new that cropped up while working on the integration tests:

The Azure key name restrictions are:
> The name must be a 1-127 character string, containing only 0-9, a-z, A-Z, and -.

Our tests so far have expected that a kek selector of `$(topicName)` will function if the topic has underscores in its name.

What should we do? Some ideas:

1. Add an expression like `$(azureNormalizedTopicName)`, implying we will convert underscores to hyphens. We then make the Azure KMS fail fast if it receives an invalid key name.
2. Add opt-in-configuration like `normalizeKeyNames: true` to the KMS config, so if the kek selector picks a key name with underscores in it, we convert them to hyphens. I've done this in the prototype, just without the opt-in feature, to get the ITs passing.
3. Rely on docs/user to know when a topic name is incompatible with key naming? Per topic keying seems like an unlikely setup in production with the cloud providers all charging roughly a dollar per key or per-key rotation.

k-wall:
Have you looked how we'll ensure that the filter operates with least privilege?  In the case of the other KMS, we've documented what privileges the filter needs and given guidance about how to achieve that.

proposals/007-azure-kms.md
k-wall: > I think we we decide to implement both schemes, maybe we implement the alternation with the versionByte?

I see you are doing that in https://github.com/kroxylicious/kroxylicious/pull/2622 already

k-wall:
> Have you looked how we'll ensure that the filter operates with least privilege? In the case of the other KMS, we've documented what privileges the filter needs and given guidance about how to achieve that. It would be good to include the approach we'll talk to least priviledge in the proposal.

@robobario did you have chance to think about this?

robobario:
> > Have you looked how we'll ensure that the filter operates with least privilege? In the case of the other KMS, we've documented what privileges the filter needs and given guidance about how to achieve that. It would be good to include the approach we'll talk to least priviledge in the proposal.
> 
> @robobario did you have chance to think about this?

Will think about how to document it, in my exploration I used:

```
    accessPolicies: [
      {
        tenantId: tenantId
        objectId: appObjectId
        permissions: {
          keys: [
            'get'
            'wrapKey'
            'unwrapKey'
          ]
        }
      }
    ]
```
as the policy to give minimum privileges, then when creating the key enabled the minimum operations the key needs:

```
keyOps: [
      'wrapKey'
      'unwrapKey'
    ]
```

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,148 @@
...
+## Proposed Initial Implementation
+
+1. Support all flavours of Key Vault. The APIs will be the same, just with a different vault base URI.
+2. Support RSA and HSM-RSA key types, wrapping using `RSA-OAEP-256` but emit a warning that it is not quantum-resistant.
+3. Support HSM-AES key type and AES-GCM wrapping
+4. Support only client credentials oauth flow with Entra using clientId + clientSecret. This supports workloads running anywhere. We could add support for other client credentials (certificates, federated certs) and Managed Identities later. Share a single auth token per Filter Definition until near expiry.
+5. Support TLS customization of the authentication client and key vault client.
+6. DEK bytes will be generated proxy-side with a SecureRandom.
+7. The Azure SDK pulls in netty/jackson/project-reactor, lets try implementing the APIs ourselves as we have for AWS
+8. Edek stores the keyName, keyVersion, edek. We attempt to minimise keyVersion size by optimistically decoding it from hex string, else store the string.
```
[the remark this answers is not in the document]
robobario: have added a more specific section within the Proposal section for the EDEK serialization formats

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,148 @@
...
+we would have either a 16-byte version implying it's the bytes, or a 32 character string.
+
+## Proposed Initial Implementation
+
+1. Support all flavours of Key Vault. The APIs will be the same, just with a different vault base URI.
+2. Support RSA and HSM-RSA key types, wrapping using `RSA-OAEP-256` but emit a warning that it is not quantum-resistant.
+3. Support HSM-AES key type and AES-GCM wrapping
+4. Support only client credentials oauth flow with Entra using clientId + clientSecret. This supports workloads running anywhere. We could add support for other client credentials (certificates, federated certs) and Managed Identities later. Share a single auth token per Filter Definition until near expiry.
+5. Support TLS customization of the authentication client and key vault client.
+6. DEK bytes will be generated proxy-side with a SecureRandom.
```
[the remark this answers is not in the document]
robobario: added this, along with a note in the Authorization section about needing `rng` permissions.

proposals/007-azure-kms.md
robobario: yep, have gone into detail in an EDEK serialization section

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,278 @@
...
+
+Another difference is that basic Key Vault cannot generate the DEK bytes for us. Managed HSM does have an operation to generate random bytes for us https://learn.microsoft.com/en-us/rest/api/keyvault/?view=rest-keyvault-keys-7.4#key-operations-managed-hsm-only
+
+### Key Identifiers
+
+When we wrap/unwrap we have to identify a key by its name and version. You can obtain the latest version by https://learn.microsoft.com/en-us/rest/api/keyvault/keys/get-key/get-key?view=rest-keyvault-keys-7.4&tabs=HTTP.
+
+The API usually returns it as a `kid` https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#object-identifiers. Like
+
+> For Vaults: https://{vault-name}.vault.azure.net/{object-type}/{object-name}/{object-version}
```
tombentley, line 114: Let's be explicit about which parts of this can vary and how. I've not read the docs, but I'm imagining that:

* `{object-name}` will obviously vary
* `{object-version}` will obviously vary
* `{object-type}` is/may always be `keys`
* `{vault-name}` ??

Let's consider too the point you made above:
> Users can also install it on-premise by negotiation.

I guess that would imply that the host name cannot be assumed to end with `.vault.azure.net`. That means that if we are going to "compress" the start of the URL it needs to be uncompressed by prefixing with the configured `keyVaultBaseUri`. But doing this comes with a commitment: That the `keyVaultBaseUri` will _never_ be changed between when any given record is encrypted and when it's decrypted. If it does change then there is a period of time during which it's ambiguous how to uncompress a compressed key id (with the old URI or the new URI).

Can we assume that keys will absolutely never migrate between instances? If this is possible then again there are potential issues for users. They would need for the key ids using the old domain name to continue to work for the duration of the life time of those records. We developers cannot put an upper bound on that.

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,278 @@
...
+different version bytes to discriminate between the formats.
+
+## Proposal
+
+1. Support all flavours of Key Vault. The APIs will be the same, just with a different vault base URI.
+2. Support RSA and HSM-RSA key types, wrapping using `RSA-OAEP-256`.
+3. Support HSM-AES key type and AES-GCM wrapping
+4. Support only client credentials oauth flow with Entra using clientId + clientSecret. This supports workloads running anywhere. We could add support for other client credentials (certificates, federated certs) and Managed Identities later. Share a single auth token per Filter Definition until near expiry.
+5. Support TLS customization of the authentication client and key vault client.
+6. If the Key Vault is Managed HSM, then we will use the API to generate random bytes. Else, DEK bytes will be generated proxy-side with a SecureRandom.
```
tombentley, line 139: Can we clarify that "the API" means the Key Vault API? (there's multiple APIs at play)
```suggestion
6. If the Key Vault is Managed HSM, then we will use the Key Vault API to generate random bytes. Else, DEK bytes will be generated within the proxy JVM using a `SecureRandom`.
```

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,278 @@
...
+7. The Azure SDK pulls in netty/jackson/project-reactor, lets try implementing the APIs ourselves as we have for AWS
+8. [Edek](#edek-serialization-scheme) stores the keyName, keyVersion, edek. We attempt to minimise keyVersion size by optimistically decoding it from hex string, else store the string.
+9. User will supply tenantId for authentication, rather than implementing a more complicated workflow to obtain it using an HTTP request to KeyVault
```
tombentley, line 142:
```suggestion
9. User will supply tenantId for authentication, rather than implementing a more complicated workflow to obtain it using an HTTP request to KeyVault. Doing something smarter and making it optional in the future will be backwards compatible.
```

proposals/007-azure-kms.md
```diff
@@ -0,0 +1,278 @@
...
+  tls:
+    ... client trust for key vault
+  entraIdentity:
+     oauthEndpointUrl: https://login.microsoftonline.com // optional defaults to this value
+     clientId:
+       passwordFile: /path/to/id
+     clientSecret:
+       passwordFile: /path/to/id
+     tenantId: "abds-1232dsaa"
+     scope: https://vault.azure.net/.default // optional, could infer from vaultBaseUri
```
tombentley, line 176: Let's require it for now. That way people will be able to do funky things if they have use cases we haven't considered. Because:
* As with the tenant id, making it optional in the future will be a backwards compatible change, 
* No one can ever be blocked by their use case not being supported, so the improvement to make it smarter is work that readily deferrable. Whereas a work item to make it configurable might not be.

tombentley (commented):
Thanks @robobario, I had a few questions, but over all this is great effort and I think it's nearly there.

k-wall:
> > > Have you looked how we'll ensure that the filter operates with least privilege? In the case of the other KMS, we've documented what privileges the filter needs and given guidance about how to achieve that. It would be good to include the approach we'll talk to least priviledge in the proposal.
> > 
> > 
> > @robobario did you have chance to think about this?
> 
> Will think about how to document it, in my exploration I used:
> 
> ```
>     accessPolicies: [
>       {
>         tenantId: tenantId
>         objectId: appObjectId
>         permissions: {
>           keys: [
>             'get'
>             'wrapKey'
>             'unwrapKey'
>           ]
>         }
>       }
>     ]
> ```
> 
> as the policy to give minimum privileges, then when creating the key enabled the minimum operations the key needs:
> 
> ```
> keyOps: [
>       'wrapKey'
>       'unwrapKey'
>     ]
> ```


I don't think we need more details in the proposal beyond saying what are the minimum permissions that proxy will need to operate.   Out of curiosity, if the KMS sniffs the key to check its type, does it need `get`.

k-wall (approved):
Really thorough job @robobario.  This doc will be a great foundation for future KMS proposals.

proposals/007-azure-kms.md
robobario: > I guess that would imply that the host name cannot be assumed to end with .vault.azure.net.

Correct, it could also differ for other national/sovereign clouds like China's Azure instance, or the US government instance.

> {vault-name} ??

This is the user-defined name for their Key Vault.

>  That means that if we are going to "compress" the start of the URL it needs to be uncompressed by prefixing with the configured keyVaultBaseUri

Correct, that's what is proposed

Some relevant capabilities:

Key Vaults can be backed up and restored within an Azure Geography and Subscription ([docs](https://learn.microsoft.com/en-us/azure/key-vault/general/backup)). Meaning that within a geo region and subscription they could clone a key/version to a new {vault-name}, potentially the objects can exist in both vaults simultaneously.

Contrasting with AWS, there is no intermediate layer similar to the named Key Vaults. In AWS you only work with Keys. You can't backup and restore keys, but AWS  it has a concept of [multi-region-keys](https://docs.aws.amazon.com/kms/latest/developerguide/multi-region-keys-overview.html) to use keys across regions. In the AWS EDEK, we encode just the key id which looks to have some limitations https://github.com/kroxylicious/kroxylicious/issues/1217.

So maybe we should record more details in the records, like the vault-name? That would let us support the user moving to a new key vault and keys, while continuing to decrypt from the old vault-name. To support a migration of existing keys to a new vault-name maybe we could do it by configuration, like the user could configure something like `alias(from=my-old-kv, to=my-new-kv)` and we'd handle the substitution.

There are only a very limited number of public clouds, so maybe we could get some compression there by using an enum to record if they are using public cloud and infer the host from that. Else, store the whole thing.

```
for cloud in $(az cloud list --query "[].name" -o tsv); do   endpoint=$(az cloud show --name $cloud --query suffixes.keyvaultDns -o tsv);   echo "$cloud: $endpoint"; done
AzureCloud: .vault.azure.net
AzureChinaCloud: .vault.azure.cn
AzureUSGovernment: .vault.usgovcloudapi.net
AzureGermanCloud: .vault.microsoftazure.de
```

robobario:
> Out of curiosity, if the KMS sniffs the key to check its type, does it need get.

Yes, I believe that's what the `get` action covers, letting us get the key metadata. We also need it to get the latest key version, because the wrap operation appears to require the keyVersion https://learn.microsoft.com/en-us/rest/api/keyvault/keys/wrap-key/wrap-key?view=rest-keyvault-keys-7.4&tabs=HTTP

I've realised this `get`, `unwrapKey`, `wrapKey` is an [outdated keyvault authorization style](https://learn.microsoft.com/en-us/azure/key-vault/general/assign-access-policy?tabs=azure-portal), and under [RBAC](https://learn.microsoft.com/en-us/azure/role-based-access-control/permissions/security#microsoftkeyvault) I think the relevant actions are:

```
Microsoft.KeyVault/vaults/keys/read
Microsoft.KeyVault/vaults/keys/wrap/action
Microsoft.KeyVault/vaults/keys/unwrap/action	
```

will try it out

proposals/009-azure-kms.md
```diff
@@ -164,31 +164,34 @@ are currently compatible with.
...
+     tls: # optional
        ... client trust configuration for oauth
 ```
 
+All configuration options are required, except `tls` configuration. Noting that `clientSecret` and `clientId` use the
+established `PasswordProvider` mechanism, so can use inline `password` or `passwordFile`.
+
 ### Authorization
 
 We want to adhere to the principal of Least Privilege.
```
k-wall, line 186: wrong one :)

```suggestion
We want to adhere to the Principle of Least Privilege.
```

proposals/009-azure-kms.md
```diff
@@ -0,0 +1,278 @@
...
+
+Another difference is that basic Key Vault cannot generate the DEK bytes for us. Managed HSM does have an operation to generate random bytes for us https://learn.microsoft.com/en-us/rest/api/keyvault/?view=rest-keyvault-keys-7.4#key-operations-managed-hsm-only
+
+### Key Identifiers
+
+When we wrap/unwrap we have to identify a key by its name and version. You can obtain the latest version by https://learn.microsoft.com/en-us/rest/api/keyvault/keys/get-key/get-key?view=rest-keyvault-keys-7.4&tabs=HTTP.
+
+The API usually returns it as a `kid` https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#object-identifiers. Like
+
+> For Vaults: https://{vault-name}.vault.azure.net/{object-type}/{object-name}/{object-version}
```
k-wall: I think option 3 is the right approach for now.  I agree, waiting for user requirements is the right way.

k-wall (approved):
LGTM

proposals/009-azure-kms.md
robobario: One more thing here, I think we should also serialize a byte for the key type (which implies the wrapping algorithm). This means at decrypt time we will have everything we need to make the unwrap request, and not need to fetch the key to get it's key type. Feels like a sensible thing to encode.

robobario: have pushed that up

PR #79, opened 2025-09-19
Authorizer

k-wall:


franvila (approved):
LGTM

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+// Inspired by org.apache.kafka.server.authorizer.Authorizer, except is asynchronous in nature.
+interface Authorizer {
+    CompletionStage<List<AuthenticationResult>> authorize(AuthorizableRequestContext context, List<Action> actions);
+}
+```
+
+```java
+// Inspired by org.apache.kafka.server.authorizer.AuthorizableRequestContext
+interface AuthorizableRequestContext {
+    String principal();
```
tombentley, line 108: I think this is a little too simplistic. Currently the proxy runtime may know any of

* A SASL authorization id
* A TLS subject DN
* A set of TLS SANs

What you're saying here is that something knows how to turn those things into a single principal string. In general there is not a single answer to that, so it needs to be pluggable.

Futhermore, there's the (relatively) easy opportunity for making this API more general allowing other information to be loaded from a remote system to augment what's known about the client (think LDAP lookup).

Moreover, we should decide explicitly whether we are following:
* the "Kafka way", where there's a single principal but this context can convey additional information like the client id, remote peer address, etc, 
* or the "JAAS way", where there is a Subject object which can hold a set of principals which can be used to model that additional information. In other words, you could have `ClientId` and `ClientAddress` as principals, and to make authorization decisions using that info.

As motivation, a use case for making an access control decision which takes into account `clientId` is: Suppose there is some team A in Example Corp which runs the internal KaaS, and some team B which runs a bunch of applications. Team B just has a single service account and uses different client ids for each of their apps. Team A find that one of Team Bs applications is causing trouble in the cluster, and need to disable it from connecting until they can get Team B to take a look. So they need to deny the (authorizedId, clientId)-pair, so that B's other (well-behaved applications) still run.

The difference between the two ways of exposing client Id in the API crops up mainly in the configuration language you propose for granting access. It's more natural to be able to say "deny CONNECT on CLUSTER to User:A and ClientId:naughty" (i.e. you just need support `and` and `or` on principal predicates in your configuration language). In contrast, exposing it via additional properties makes the language less elegant, because you're lost the regularity of the `${PrinicpalType}:${PrincipalName}` syntax.

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+```
+
+```java
+// The following types are inspire by the Kafka classes of the same name and have the same role.  However, interfaces are used
+// rather than enums to allow for extensibility (using pattern suggested by https://www.baeldung.com/java-extending-enums)
+interface AclOperation {
+    String operationName();
+}
+
+enum CoreAclOperation implements AclOperation {
```
tombentley, line 128: One thing I don't like about this is that there's no safety mechanism to ensure that the operation is compatible with the resource type.

I was playing with a pattern last week where we would represent a resource type simply as an `enum` of the operations it supports (your `ResourceType` becomes a `Class` of that `enum`). Thus `TopicResource` ends up with `READ`, `WRITE`, `DESCRIBE` etc and ClusterResource ends up with `CONNECT`, `DESCRIBE` etc, and you can construct a type safe API where asking for an authorization decision on `TopicResource.CONNECT ` is a compile time error.

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+       return patternType.matches(pattern, resourceName);  
+    }
+}
+```
+
+#### Rules File
+
+The rules file expresses a mapping between principals (user type only with exact match) and an allow-list of resources.
+
+For the initial scope, only resource rules of type TOPIC are supported.  In order to allow for future extension, the user’s configuration set the version property to 1.  This will allow future versions of the filter to introduce support for other resource types without changing the meaning of existing configurations.
```
tombentley, line 171: There's something important here which you're not calling out clearly.

1. The rules are deny by default (this is implied by your saying it's always an allow list)
2. We're only supporting Topic resource initially
3. Thus without the version are could not compatibly add a new resource type, because existing rules files would be interpretted as denying access to the newly-supported resource types.

I think it's worth spelling it out more clearly.

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+```java
+// Inspired by org.apache.kafka.server.authorizer.Authorizer, except is asynchronous in nature.
+interface Authorizer {
+    CompletionStage<List<AuthenticationResult>> authorize(AuthorizableRequestContext context, List<Action> actions);
+}
+```
+
+```java
+// Inspired by org.apache.kafka.server.authorizer.AuthorizableRequestContext
+interface AuthorizableRequestContext {
```
tombentley, line 107: You see to consider how you handle an anonymous client. In particular what is their principal.

tombentley (commented):
I think this is definitely heading in the right direction. I'm currently experimenting with a few ideas based on what you describe. I'd like to carry on doing that for a few days and then use the knowledge gained to revise this, if that's OK with you @k-wall?

proposals/010-authorizer.md
k-wall: > What you're saying here is that something knows how to turn those things into a single principal string. In general there is not a single answer to that, so it needs to be pluggable.

I don't disagree at all.  However, isn't it basically the same conversation that we had during the Authenticate API work (https://github.com/kroxylicious/design/pull/71) (whether to use Subject, Principal, Set<Principals>).  There we decided to be content with a String until such times we need something richer.  The proposal takes the approach.

We've got the prinicpalType in the rules yaml, so that gives us wiggle room to refactor the APIs and implemementation later, without breaking user-config.

proposals/010-authorizer.md
k-wall: Good point. I hadn't thought about that at all.  I suppose you might have two gateways, one attached to a SASL listener and one attached to listener using anonymous.  You might want to give the anonymous connections fewer privileges. 

I need to think about this one.

Two thoughts: 
- As things stand now, as Kafka doesn't use https://www.rfc-editor.org/rfc/rfc4505, there's no SASL negotiation, so there's nothing to actually call `io.kroxylicious.proxy.filter.FilterContext#clientSaslAuthenticationSuccess`.  `#clientSaslContext` would return an empty optional.
- Kafka uses User:ANONYMOUS is its ACL, but that feels weak.  Somebody could define a user called `ANOMYMOUS` and get the rights ascribed to ANONYMOUS.  How useful is such an attack vector? I'm not sure.
-

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+| DESCRIBE  | Topic         | ListOffset, OffsetFetch, OffsetFetchForLeaderEpoch DescribeProducers, ConsumerGroupHeartbeat, ConsumerGroupDescribe, ShareGroupHeartbeat, ShareGroupDescribe, MetaData, DescribeTopicPartitions, ConsumerGroupDescribe |
+| CONNECT   | Cluster       | SaslAuthenticate                                                                                                                                                                                                       |
+
+In general, the filter will make access decisions in the same manner as Kafka itself.  This means it will apply the same authorizer checks that Kafka enforces itself and generate error responses in the same way.  It will
+From the client's perspective, it will be impossible for it to distinguish between the proxy and kafka cluster itself.
+It will also use the same implied operation semantics as implemented by Kafka itself, such as where `ALTER` implies `DESCRIBE`, as described by 
+`org.apache.kafka.common.acl.AclOperation`.
```
tombentley, line 69: We should define what has responsibility for this. It could be:

* A part of the contract of `interface Authorizer`
* An implementation detail of the particular Authorizer implementation being proposed, so that other Authorizers are not required to respect this implication

If the former then it's natural to model the implication relation on the operation enum. If the latter then probably we're saying that the Authorizer implementation has to special case those topic operations.

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+#### Authorizer Filter
+
+Filter Configuration:
+
+```yaml
+type: AuthorizerFilter
+config:
+  authorizer: FileBasedAllowListAuthorizer
+  authorizerConfig:
+    rulesFile: /path/to/allow-list.yaml
```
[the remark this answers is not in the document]
k-wall: Yeah, I believe it will.

proposals/010-authorizer.md
```diff
@@ -0,0 +1,212 @@
...
+### Request authorization
+
+The Authorizer filter will intercept all request messages that perform an action on a resource, and all response messages that list resources.  
+
+On receipt of a request message from the downstream, the filter will make an asynchronous call to the authorizer for the resource(s) involved in the request.  If the authorization result for all resources is `ALLOWED`, the filter will forward the request to the broker.
+If the authorization result is `DENIED` for any resource in the request, the filter will produce a short circuit error response denying the request using the appropriate authorization failed error code.  The authorizer filter must not forward requests that fail authorization. 
+
+### Response resource filtering
+
+On receipt of a response message from the upstream, the Authorizing filter will filter the resources so that the downstream receives only resources that they are authorized to `DESCRIBE`.
```
robobario, line 42: Something Tom pointed out in discussion is that Metadata, DescribeGroups, DescribeCluster, ConsumerGroupDescribe, ShareGroupDescribe responses can contain authorized operations bitmasks. Will this Filter manipulate the metadata for these RPCS? Should we be generating our own mask for these resources and or-ing it with the upstream?

tombentley: Agreed, we have to implement this correctly.

proposals/010-authorizer.md
k-wall: Done.

proposals/010-authorizer.md
k-wall: I don't understand why authenticatedSubject needs to return a CompletionStage.  Won't there always be a mutable Subject, which might be empty?  Filters or plugins will add to the Subject as they learn more about the connection.

aside:  is worth separating out the new FilterContext API work from the authorizer?  That would allow the Audit Filter work to continue.

tombentley: The problem with that approach is that in introduces non-determinism into how the proxy can behave. For example different ordering of events between the subject being updated and the subject being queried by a filter (and making an authz decision based on that subject) might lead to false negative (or, worse, false positive) access control decisions:

1. A filter provides an authorized id, which initiates a request is made to AD to get some roles.
2. A filter queries the authenticated subject. Finds only a `SaslId`, and makes a decision based on that.
3. The AD response is received.

We cannot block (because netty), nor otherwise prevent a filter making progress at step 1. That's because the API for providing the SASL authorized id is not async. Pausing invocation of the rest of filters in the pipeline after the authenticating filter returns does not work, because nothing prevents _the same_ filter, having provided the id from immediately asking for the subject. 

Therefore to avoid the badness in step 2 we have to avoid actually providing the subject until we have received the response from AD. But we can't simply block the thread (because netty), so returning a CompletionStage seems to me to by the only choice we have.

tombentley:
> We said during the sync call that @tombentley was going to iterate on the Java APIs in a PR, we'd then update the proposal with the outcome of that.

See https://github.com/kroxylicious/kroxylicious/pull/2702

PR #80, opened 2025-09-19
015 - entity isolation filter

k-wall:
Proposes the introduction of a "Entity Isolation Filter" to Kroxylicious's core filters.

This filter gives each authenticated user a private space within a Kafka cluster, isolated from other users sharing the same cluster.  Isolation can be applied selectively to some entity types and not others. 
For example, it's possible to have isolation for consumer groups, but not for topics.


proposals/011-user-namespace.md
```diff
@@ -0,0 +1,69 @@
...
+
+#### Resource Name Mapper API
+
+Asynchronous interface that lets the filter map from downstream names to upstream names and vice-versa.  It also
+has the responsibility to filter the upstream view, removing resources that don't belong in the view.
+
+```java
+interface ResourceNameMapper {
+     /** Return a mapping of downstream names to upstream names. */
+	CompletionStage<Map<String, String>> mapDownstreamResourceNames(String authorizationId, ResourceType resourceType, List<String> downstreamResourceNames);
```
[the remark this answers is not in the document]
k-wall: I think it only makes sense to use this filter when a filter has been established on the channel.

proposals/011-user-namespace.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Motivation
+
+We encountered a use-case where the desire was to share topics but maintain isolated spaces for groups and transactional-ids.
+
+## Proposal
+
+The role of the Resource Isolation filter is to give the client the impression of a private kafka cluster space that is isolated from other clients sharing the cluster.  Namespacing can be applied selectively to different resource types.
+
+The filter will use a pluggable API to determine how to map the name of each resource.  Operations that retrieve lists of resources will see only those that fall within the namespace.
```
robobario: Will the Filter fail-closed if it encounters new versions of RPCs? I think we've decided this in the PR but would be good to update the design. Like the Authz Filter I think this one needs to modify ApiVersions response data to only advertise supported RPCs, and fail explosively if we recieve an RPC version that the Filter doesn't know how to handle. We should only let through RPC versions that either carry no entity information, or the Filter knows how to mutate that version.

k-wall: Added.

k-wall:
@robobario and @tombentley low priority for re-review.

proposals/015-entity-isolation.md
```diff
@@ -0,0 +1,175 @@
...
+    principalType: io.kroxylicious.proxy.authentication.User
+```
+
+## EntityNameMapper API
+
+Synchronous API that lets the filter map from downstream names to upstream names and vice versa.
+It also has the responsibility to filter the upstream view, removing resources that don't belong in the view.
+
+Asynchronous interface that lets the filter map from downstream names to upstream names and vice-versa.  It also
+has the responsibility to filter the upstream view, removing resources that don't belong in the view.
```
SamBarker: nit: The API below looks Synchronous to me, I think this is a stray paragraph.

```suggestion
```

SamBarker (approved):
LGTM thanks @k-wall

tombentley (approved):
Thanks @k-wall

PR #83, opened 2025-11-18
012 - Hot reload feature proposal

Uzziee:
This proposal is to add hot reload functionality, which will enable the app to reload any changes to virtual cluster config without the need to restart the app

PR #89, opened 2026-02-17
feat: Virtual cluster lifecycle state model

SamBarker:
## Summary

Introduces proposal 016 defining a lifecycle state model for virtual clusters. Virtual clusters are modelled as entities in the Domain-Driven Design sense — each has a persistent identity (its name) that carries through state changes. Reload is a transition the entity passes through, not a replacement of it.

The model defines five states:

- **initializing** — cluster being set up; not yet serving traffic
- **serving** — proxy has completed setup and is serving traffic
- **draining** — no new connections or requests accepted; existing in-flight requests given the opportunity to complete
- **failed** — configuration not viable; all partially-acquired resources released
- **stopped** — terminal; cluster permanently removed from configuration

## Key Design Decisions

- **Virtual clusters are entities** — each cluster's identifier is its name; the state machine tracks the entity, not a configuration instance. Renaming a cluster is a destructive operation.
- **`serving` not `accepting`** — the state name reflects that the proxy is actively serving traffic, not merely open to new connections
- **Runtime health is out of scope** — `serving` is not split into healthy/degraded; runtime health is orthogonal to lifecycle and better addressed by readiness probes and metrics. Circuit-breaking is noted as a manifestation of the same concern.
- **Draining takes the client's view of in-flight** — a request is in-flight from the moment the client sends it until it receives a response. New requests on existing connections are rejected during drain to ensure the cluster quiesces.
- **Graceful drain is best-effort** — the drain timeout is the hard backstop; aligns with Kafka ecosystem norms (brokers drain before stopping)
- **`partialInitialisationPolicy: serve-none | serve-others`** — governs proxy behaviour when cluster initialisation fails, whether on startup or reload. Default is `serve-none` (matches current behaviour).
- **`stopped` is terminal** — reload routes through `draining` → `initializing`, never through `stopped`
- **Observability is public API** — management endpoint and metrics (current state, time in state, transition count) are committed to in the proposal

## Relationship to Other Proposals

This provides a foundation for [012 - Hot Reload](https://github.com/kroxylicious/design/pull/83), which can define reload transitions on this state model. Note: the Hot Reload proposal's "Cluster modification: remove + add" framing describes internal runtime mechanics; from the lifecycle model's perspective the named entity persists through modification.

## Test plan

- [ ] Review state definitions for clarity and completeness
- [ ] Verify state transition diagram consistency with transition descriptions
- [ ] Review rejected alternatives for completeness
- [ ] Consider whether future enhancements section adequately captures known follow-up work


proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,292 @@
...
+
+2. **Shutdown is unstructured.** The proxy stops accepting connections and closes channels, but there is no formal draining phase that ensures in-flight Kafka requests complete before the connection is torn down.
+
+3. **No foundation for partial failure.** Proposals such as [012 - Hot Reload](https://github.com/kroxylicious/design/pull/83) need the ability to express "cluster-b failed to apply new configuration but cluster-a is still serving traffic." Without a lifecycle model this state is undefined and unreportable.
+
+4. **No visibility into runtime health.** A virtual cluster may have its configuration applied successfully but be unable to serve traffic because an upstream broker is unreachable or a KMS is unavailable. There is currently no way to represent or report this.
+
+## Motivation
+
+As Kroxylicious moves toward use cases where it acts as a multi-tenant gateway (multiple independent virtual clusters serving different teams or workloads), the blast radius of failures becomes critical. A configuration error affecting one tenant's cluster should not disrupt another tenant's traffic.
```
tombentley: This conflates tenancy with virtual cluster. They're not necessarily the same thing. For example, @k-wall's isolation filter is a different take on tenancy, one where different tenants are in the same VC. 

I think we should try to keep any mention of tenancy out of this proposal, because what is proposed here can stand on its own terms without needing to bring any kind of tenancy into the discussion.

The hot-reload proposal aims to make the virtual cluster the "unit of reconfiguration". That seems reasonable in the sense that a VC defines a bunch of network-level configuration, and network level stuff is a significant source of potential reconfiguration failure. So making the VC the "unit of reconfiguration" also makes the VC the "failure domain of reconfiguration". Roughly speaking, I think the VC is the largest possible failure domain that's smaller than a whole proxy instance. 

So thinking in terms of failure domain seems like a more useful was of framing the motivation, without needing to make assumptions about what the user is using virtual clusters for.

tombentley (commented):
Thanks @SamBarker I think this is useful work, but have doubts that we really need to have "healthy" as a state.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,292 @@
...
+| **healthy** | The cluster is fully operational — all runtime health checks are passing. The cluster accepts connections. A cluster may return to `degraded` if a dependency becomes unavailable. |
+| **draining** | New connections are rejected. Existing connections remain open to give in-flight requests the opportunity to complete. Connections are closed once idle or when the drain timeout is reached. |
+| **failed** | The proxy determined the configuration not to be viable. All partially-acquired resources are released on entry to this state. The proxy retains the cluster's configuration and failure reason for diagnostics and retry. |
+| **stopped** | The cluster is no longer operational. All resources have been released. This is a terminal state. |
+
+### Health Checks
+
+The transitions between `degraded` and `healthy` are driven by health checks. This proposal defines that these transitions exist and that some mechanism triggers them, but does not prescribe what constitutes a health check. The criteria for health (upstream broker connectivity, KMS availability, filter readiness, etc.) are a separate concern from the lifecycle model itself.
```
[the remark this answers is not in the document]
SamBarker: Without taking a position on the overall point your making. 

> they would seem depend on external things (like is the Kafka cluster running?) which means transitions between them are likely to depend on some kind of polling mechanism, which is going to be a delayed signal.

The delayed nature of the signal was what pushed me towards `healthy` & `degraded`, partly I'd admit because I was including notions of availability of remote resources into the picture (which your reasonably pushing back on) but if felt strange to transition to `running` if. for instance, we can't connect to the upstream broker. 

> their definition might not even be the same for all observers. For example what if some clients connect to the proxy directly but some via some network loadbalancer or similar. You can't define healthy in the same way for both these clients without the possibility of disagreement (the loadbalancer not working makes the VC appear unhealthy for some but not others). 

I was constructing healthy, or rather health in general, as from the **proxies** perspective. What does the run time think the state of play is. 

>By building in a notion of "healthy" you're making a rod for your own back however you end up defining it.

I do find that point persuasive, we are always going to get different people wanting different definitions of healthy. However I think everyone would agree if the proxy can't connect to an upstream endpoint then its `degraded` (at best). While we don't have any immediate requirements to model anything other than the basic starting, running, stopping  states I can easily imagine requirements where detecting a failed broker we might want to apply rate limits to new connections to manage a thundering heard on re-connect. 

The health of a virtual cluster feels like an important thing for proxy administrators to be aware of and understand. So I was combining it into this life cycle. I can see there is a case to make they are orthogonal concepts. 

So I'm very interested to hear from @kroxylicious/developers as to how they see the model.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. **Startup is all-or-nothing.** If one virtual cluster fails to start (e.g. port conflict, filter initialisation failure), the entire proxy process fails. Other clusters that could have started successfully are taken down with it.
+
+2. **Shutdown is unstructured.** The proxy stops accepting connections and closes channels, but there is no formal draining phase that ensures in-flight Kafka requests complete before the connection is torn down.
+
+3. **No foundation for partial failure.** Proposals such as [012 - Hot Reload](https://github.com/kroxylicious/design/pull/83) need the ability to express "cluster-b failed to apply new configuration but cluster-a is still serving traffic." Without a lifecycle model this state is undefined and unreportable.
+
+## Motivation
+
+A virtual cluster is the natural unit of independent operation — the smallest scope at which the proxy can contain a failure without affecting unrelated traffic. Today this independence is not modelled: the proxy treats all clusters as a single unit that either starts completely or fails completely.
```
tombentley: "modelled" is a bit of an odd word here (which suggests to me that there are come components/classes in the source code which directly correspond to this, but really its more like an emergent property of the system, rather than an inherent one) . I think it would be clearer to say something like "the independence is notional"

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,203 @@
...
+
+A configurable startup policy allows deployments where partial availability is preferable to no availability:
+
+```yaml
+proxy:
+  startupPolicy: fail-fast  # default — any cluster failure prevents startup
+  # startupPolicy: best-effort  # start with whatever clusters succeed
+```
+
+In best-effort mode, the proxy starts and serves traffic for clusters that initialised successfully, while reporting failed clusters via health endpoints and logs. Kubernetes readiness probes or monitoring systems can apply their own thresholds (e.g. "all clusters must be accepting" vs "at least one cluster must not be failed"). The operator would typically set this policy.
```
tombentley:
```suggestion
In best-effort mode, the proxy starts and serves traffic for clusters that initialised successfully, while reporting failed clusters via health endpoints and logs. Kubernetes readiness probes or monitoring systems can apply their own thresholds (e.g. "all clusters must be accepting" vs "at least one cluster must not be failed"). The Kubernetes operator would typically set this policy.
```

Just because operator is easily misunderstood to mean a human

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Reload without draining
+
+The current reload path requires draining connections before reinitialising (`accepting` → `draining` → `initializing`). In the future, it may be possible to skip the drain step for certain types of configuration change — for example, swapping the filter chain in place or reconnecting upstream without dropping client connections.
+
+This would introduce a direct `accepting` → `initializing` transition. The state model as proposed accommodates this without structural changes: `initializing` already represents "setting up the cluster," and its exit transitions (`accepting` on success, `failed` on failure) remain the same regardless of whether draining preceded it.
+
+Some configuration changes will likely always require draining — for example, changes to the upstream cluster identity or TLS configuration that invalidate existing connections. The optimisation is about identifying changes where draining can be safely skipped, not eliminating it.
+
+### Proxy-level lifecycle
```
tombentley: You've defined how proxy started works wrt the VM state machines. It's not really clear to me whether we need a proposal about the rest, unless you think that that state is also exposed via metrics/mgmt etc.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,203 @@
...
+**Recovery transitions:**
+- `failed` → `initializing`: a retry is requested (e.g. operator action, reload with corrected config). Since `failed` clusters have already released all resources, this is a clean start from scratch.
+
+### Proxy Startup Behaviour
+
+On startup, the proxy attempts to initialise each virtual cluster in the configuration. Clusters that succeed move to `accepting`. Clusters that fail move to `failed` with a captured reason.
+
+By default, the proxy fails to start if any cluster fails to initialise (fail-fast). This is the correct behaviour for most deployments — configuration errors should be surfaced immediately, especially in development and bare-metal environments.
+
+A configurable startup policy allows deployments where partial availability is preferable to no availability:
```
robobario: Should it be a startup policy only, or a behaviour policy that could also cover a dynamic reload failure in future? Or do you see it as a policy that's only useful when there is no dynamic reloading involved?

I'm wondering if there are any users who might want a setup where you can reload configuration (like Authz rules) but have the proxy shutdown immediately if the reload fails.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,203 @@
...
+
+The lifecycle model's job is to track what the proxy is doing with a cluster — setting it up, accepting connections, draining, or torn down. Whether the cluster can successfully serve traffic is a separate, orthogonal concern better addressed by readiness probes, health endpoints, or metrics that can evolve independently.
+
+### Runtime health model
+
+We considered defining a broader health model alongside the lifecycle — covering upstream broker reachability, KMS availability, filter readiness, and similar runtime concerns. This was ruled out of scope. Health depends on what the proxy is being used for: a proxy doing record encryption has different health criteria from one doing schema validation. The appropriate health model will vary by deployment, and may need to account for request-level routing where health is per-destination rather than per-cluster. Defining a health model prematurely would constrain future design options without providing immediate value. The lifecycle model intentionally leaves room for health to be addressed separately.
+
+## Future Enhancements
```
robobario: Do you think resilience enhancements would fit in this State Machine later on? Since it knows about `ACCEPTING` connections, is it also a place where we might add in a circuit-breaker later to prevent connections in certain circumstances, like DOS/upstream-unhealthy?

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Current Situation
+
+Today a virtual cluster is either fully registered (ports bound, accepting connections) or it does not exist. There is no intermediate or error state.
+
+This has several consequences:
+
+1. **Startup is all-or-nothing.** If one virtual cluster fails to start (e.g. port conflict, filter initialisation failure), the entire proxy process fails. Other clusters that could have started successfully are taken down with it.
+
+2. **Shutdown is unstructured.** The proxy stops accepting connections and closes channels, but there is no formal draining phase that ensures in-flight Kafka requests complete before the connection is torn down.
```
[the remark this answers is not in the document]
SamBarker: Sorry, I've got local changes that were meant to go out in sync with comment. 

Also I thought you were pushing for the opposite solution, let the protocol handle it all and skip the drain.

I'll push the changes shortly and come back to this.

proposals/016-virtual-cluster-lifecycle.md
SamBarker: Noted in the rejected alternatives under runtime health — circuit-breaking is a manifestation of the same runtime health concerns. The same reasoning that excludes runtime health from the lifecycle model excludes circuit-breaking; these are not cluster lifecycle transitions in the scope of this proposal.

proposals/016-virtual-cluster-lifecycle.md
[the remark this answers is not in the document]
SamBarker: The idempotent producer concern is real and the proposal now acknowledges it explicitly. However, this is an existing proxy limitation that predates this proposal — any proxy restart today drops all connections with the same effect. A reload mechanism building on this lifecycle model actually improves the situation by confining disruption to a single virtual cluster rather than the whole process.

Solving the fundamental problem — replicating broker-style partition leader migration to avoid session disruption entirely — is a substantial piece of work well beyond the scope of a lifecycle model, with a multitude of possible solutions that the proxy cannot currently express. The proposal notes the limitation and directs users to consider how analogous situations (broker crash, network split) are handled without the proxy, since clients must already be prepared for those. Users requiring exactly-once guarantees across reconnections should use transactional producers.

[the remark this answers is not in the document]
SamBarker: The draining section has been updated to clarify in-flight semantics and correct an earlier misstatement about idempotent producers.

On the specific concern about idempotent producers: a proxy-initiated drain is no different from any other connection loss event from the client's perspective. Idempotent producers retain their Producer ID and sequence numbers in client memory across reconnections — a connection drop does not start a new session. When the client reconnects after a drain, it retries unacknowledged requests with the same (PID, epoch, sequence) tuple, and the broker deduplicates them using its existing producer state. A new PID is only assigned on producer process restart, not on reconnection. The drain timeout is the hard backstop after which connections close, but this is protocol-compliant — Kafka clients are required to handle connection loss, and the idempotent producer's deduplication mechanism is designed to survive exactly this scenario.

References:
- [KIP-360](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=89068820) documents how the broker retains producer state and how idempotent producers can safely bump their epoch when broker state is lost — the key point being that the PID is retained client-side across reconnections.
- The [Idempotent Producer design doc](https://cwiki.apache.org/confluence/display/KAFKA/Idempotent+Producer) confirms that deduplication is based on (PID, epoch, sequence) per partition, not per connection: "The broker maintains in memory the sequence numbers it receives for each topic partition from every PID."
- [KIP-854](https://cwiki.apache.org/confluence/display/KAFKA/KIP-854+Separate+configuration+for+producer+ID+expiry) covers broker-side producer state expiry — the state is retained well beyond any reasonable drain timeout.

The lifecycle model actually improves the posture for idempotent producers compared to the status quo. Today, any configuration change requires a full process restart, dropping every connection across every virtual cluster. With per-cluster lifecycle and reload, only the affected cluster's connections drain — the blast radius shrinks from "all clusters" to "one cluster" and planned restarts become less frequent.

I think the proposal goes far enough on draining. The lifecycle model defines that draining exists as a state and its high-level semantics (reject new connections, reject new requests, allow in-flight to complete, timeout as backstop). The detailed mechanics — specific error codes for rejected requests, interaction with pipelining, TCP-level behaviour — are implementation concerns that don't change the lifecycle state model and are better resolved in the reload proposal where draining becomes a frequent operation rather than a rare shutdown event.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,204 @@
...
+
+On startup, the proxy attempts to initialise each virtual cluster in the configuration. Clusters that succeed move to `serving`. Clusters that fail move to `failed` with a captured reason.
+
+By default, the proxy serves no traffic if any cluster fails to initialise. This is the correct behaviour for most deployments — configuration errors should be surfaced immediately, especially in development and bare-metal environments.
+
+The policy is configurable and applies whenever clusters initialise, whether on first startup or during reload:
+
+```yaml
+proxy:
+  partialInitialisationPolicy: serve-none  # default — any cluster failure prevents serving traffic
```
tombentley: I'm on-board with the broad concept, but:

* I don't find this name to be very intuitive
* Is there an equivalent thing which defines how the proxy handles failures during reload? I don't think we'd really need a separate config for that, (or did you have a use case in mind?)

This really seems to be about specifying the user's tolerance for virtual clusters entering/remaining in the failed state. Currently the state machine diagram shows a retry loop, but the way you're describing it here is that loop is, in actual fact, the kubernetes pod crash loop. And therefore that control loop does not exist when the proxy is not running in kube. 

Perhaps we should borrow some ideas from how Kubernetes allows pods to fail, but apply them directly in the proxy (whether or not it happens to be running in Kube). Specifically, some number of retries, a retry delay ± a backoff. And implicitly we terminate when we run out of retries for any VC. (You can configure the number of reties as `Integer.MAX_VALUE` if you don't want to terminate). We can add some other policy than terminate later on if it's needed.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,204 @@
...
+  # partialInitialisationPolicy: serve-others  # serve clusters that initialised successfully
+```
+
+In `serve-others` mode, the proxy serves traffic for clusters that initialised successfully, while reporting failed clusters via health endpoints and logs. Kubernetes readiness probes or monitoring systems can apply their own thresholds (e.g. "all clusters must be serving" vs "at least one cluster must not be failed"). The Kubernetes operator would typically set this policy.
```
tombentley: readiness probes, but I suppose also liveness probe in a future where we have dynamic reload.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Graceful Shutdown
+
+When the proxy receives a shutdown signal:
+
+1. All `serving` clusters transition to `draining`.
+2. All `failed` clusters transition directly to `stopped`.
+3. New connections are rejected for draining clusters. New requests on existing connections are also rejected.
+4. For each existing connection, the proxy waits for in-flight requests to complete, up to a configurable drain timeout. The proxy takes the same view of in-flight as the Kafka client: a request is in-flight from the moment the client sends it until the client receives a response.
```
tombentley: The proxy cannot know when/whether the client receives a response that's been forwarded to it (at least not without inferring that fact from the receipt of another request, but that defeats the point).

In any case, I just don't see that this can really be completely solved by the proxy in its current shape. To solve it properly I think we'd need to "virtualise the brokers". That would allow the proxy to leverage Kafka's own technique for shutting down brokers cleanly: Tell the client that this broker is no longer the leader or coordinator for anything, so that all the clients update their metadata to find the new leader/coordinator. Doing that would require a proper cluster-of-proxies, which is not something we have today.

So I think what you're describing here is not perfect, but it's as good are we're likely to get for now, and we shouldn't block this proposal while we figure out a more perfect solution.

tombentley (commented):
Thanks @SamBarker, I think this is looking pretty good. I had a question is about how the configuration API for the termination behaviour.

proposals/016-virtual-cluster-lifecycle.md
SamBarker: I can see where your coming from. I had hoped that `initialisation` would be read as VC initialisation rather than proxy startup. 

How about restructuring as:

```yaml
proxy:
  onVirtualClusterFailure:
    serve: none          # default
    # serve: remaining   # keep serving healthy clusters
```

`onVirtualClusterFailure` avoids ambiguity with target clusters. The values describe the proxy-level consequence rather than the individual cluster's fate.

I think retry is orthogonal to this policy. `serve` answers "what about the rest of the proxy?" while retry answers "what about the failed cluster?" They compose independently. By making `onVirtualClusterFailure` a structured block rather than a flat string, new keys can be added alongside `serve` in future proposals without breaking existing configs. For example:

```yaml
proxy:
  onVirtualClusterFailure:
    serve: remaining
    # future extensions — new keys, existing configs unaffected:
    # retry:
    #   maxRetries: 3
    #   backoff: 5s
    # escalate:
    #   url: http://example.slack.com/
    #   messageFormat: "%s failed with %s"
    #   args: [VIRTUAL_CLUSTER, EXCEPTION_MESSAGE]
```

This proposal defines `serve` with its two values. The rejected alternative on automatic retry still applies — we're not committing to implementing retry here, but the config structure leaves room for it without a breaking change.

proposals/016-virtual-cluster-lifecycle.md
SamBarker: Agreed on both points. "As best it can" now hedges the in-flight definition — the proxy observes its own reads and writes, not what the client has actually sent or received.

And yes, virtualising the brokers would let the proxy leverage Kafka's own clean shutdown mechanism (leader migration via metadata updates). That would be a really compelling capability if we can figure out how to make it work. For now, best-effort draining with a timeout backstop is as good as we can get without it.

proposals/016-virtual-cluster-lifecycle.md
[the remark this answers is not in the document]
SamBarker: @tombentley can we resolve this thread?

proposals/016-virtual-cluster-lifecycle.md
SamBarker: @tombentley similar question here, are we happy?

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,207 @@
...
+The policy is configurable and applies whenever clusters initialise, whether on first startup or during reload:
+
+```yaml
+proxy:
+  onVirtualClusterFailure:
+    serve: none        # default — any cluster failure prevents serving traffic
+    # serve: remaining # serve clusters that initialised successfully
+```
+
+When `serve` is set to `remaining`, the proxy serves traffic for clusters that initialised successfully, while reporting failed clusters via health endpoints and logs. Kubernetes readiness probes, liveness probes, or monitoring systems can apply their own thresholds (e.g. "all clusters must be serving" vs "at least one cluster must not be failed"). The Kubernetes operator would typically set this policy.
```
Uzziee: Can you help me understand how will the experience look with rollback on failure for hot-reload ?

This is what I think would happen if we mark it as "remaining"

Eg:- 
1. There are 3 virtual clusters running (cluster-1, cluster-2, cluster-3)
2. We apply a change to cluster-2
3. cluster-2 fails to come up, initiate rollback
4. cluster-1, cluster-3 will still continue to serve the traffic during this rollback phase
5. cluster-2 is rolled back to previous state and is up and running

The only risk I see with "remaining" is during the app startup, we will have to put in some extra logic to ensure that the entire start state is stable

SamBarker: Thanks, and sorry this has taken a while to get to, for walking through the scenario — your understanding of the reload flow is correct:

1. cluster-1, cluster-2, cluster-3 all `serving`
2. cluster-2 reloads: `serving` → `draining` → `initializing` → `failed`
3. cluster-1 and cluster-3 continue `serving` throughout
4. Rollback triggers recovery: `failed` → `initializing` with old config
5. If rollback succeeds: `initializing` → `serving` — cluster-2 is back

Your concern about startup needing "extra logic" pointed to a naming and modelling issue.

`remaining` implies there was a prior known-good state — things that are "left over" after a failure. That makes sense during reload but is misleading at startup where nothing was serving yet. We're renaming the values to `serve: none | successful` to remove that implication.

More importantly, your question led us to re-examine where the `serve` policy sits in the model. The original design hung it off `onVirtualClusterFailure`, but `failed` is not a terminal state — it has a recovery transition (`failed` → `initializing`). Putting `serve` on the failure state creates tension with recovery mechanisms like rollback: "shut everything down" and "try to recover" conflict unless you add temporal sequencing to resolve them.

The fix is to move `serve` to the terminal state. `serve` applies when a virtual cluster reaches `stopped` — the point where the runtime has given up on it. Today, without reload, a startup failure goes `failed` → `stopped` immediately (no recovery exists), so `serve` fires straight away. With reload in the future, the cluster stays in `failed` while recovery runs (rollback, retry). If recovery succeeds, it never reaches `stopped` and `serve` never fires. If recovery is exhausted, the cluster transitions `failed` → `stopped` and `serve` applies.

This keeps `serve` simple (it only cares about the terminal state), leaves `onVirtualClusterFailed` as a clean seam for the reload proposal to define recovery policies, and avoids any tension between the two. We're also renaming the config node from `onVirtualClusterFailure` to `onVirtualClusterStopped` to match the state it responds to.

Updating the proposal text to reflect this.

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,209 @@
...
+3. New connections are rejected for draining clusters. New requests on existing connections are also rejected.
+4. For each existing connection, the proxy waits for in-flight requests to complete, up to a configurable drain timeout. The proxy takes, as best it can, the same view of in-flight as the Kafka client: a request is in-flight from the moment the client sends it until the client receives a response.
+5. Once drained (or timed out), connections are closed and clusters move to `stopped`.
+6. The proxy process exits.
+
+The drain timeout should be configurable. Kafka consumers with long poll timeouts (`max.poll.interval.ms` defaults to 5 minutes) or slow producers with `acks=all` can legitimately need more than the 30 seconds assumed in current code.
+
+```yaml
+proxy:
+  drainTimeout: 60s  # default TBD
```
hrishabhg: Since virtual clusters are modelled as independent entities with their own lifecycles, should `drainTimeout` be configured at the virtual cluster level rather than the root `proxy:` level? Different VCs may proxy to Kafka clusters with very different timeout characteristics (e.g. `max.poll.interval.ms`, producer `acks=all` latency), so a single global drain timeout could be too aggressive for some and too lenient for others.

hrishabhg (approved):
The entity model is well-defined, and the state machine is clean and practical. The separation of runtime health from lifecycle state is the right call. One minor suggestion left on `drainTimeout` scoping, but nothing blocking.

cc: @Uzziee

proposals/016-virtual-cluster-lifecycle.md
```diff
@@ -0,0 +1,209 @@
...
+
+By default, the proxy serves no traffic if any cluster reaches `stopped`. This is the correct behaviour for most deployments — configuration errors should be surfaced immediately, especially in development and bare-metal environments.
+
+The policy is configurable:
+
+```yaml
+proxy:
+  onVirtualClusterStopped:
+    serve: none        # default — any cluster reaching stopped prevents serving traffic
+    # serve: successful # serve clusters that initialised successfully
```
hrishabhg, line 90: alternates:
1.
```
partialFailurePolicy: halt | continue/best-effort
```
2.
```
clusterIsolationPolicy: strict | lenient
```

proposals/016-virtual-cluster-lifecycle.md
hrishabhg: Should draining be opt-out? Kafka clients are designed to handle broker disconnections — they retry, reconnect, and rediscover leaders automatically. A proxy restart without switchover looks like a broker disconnect to clients. Draining primarily adds value during switchover scenarios where in-flight requests must complete on the old path before traffic moves to a new one. For simpler cases like reloading authz rules or config changes that don't affect connectivity, the drain phase adds unnecessary latency. Consider supporting `drainTimeout: 0s` (or an explicit `drain: false`) to skip draining entirely and transition directly to `stopped`. This also aligns with making `drainTimeout` per-VC — some VCs may never need draining.

cc: @Uzziee

proposals/016-virtual-cluster-lifecycle.md
SamBarker: Thanks @hrishabhg for the alternatives — they made us stop and think about the naming, which is always valuable.

`clusterIsolationPolicy` is appealing because it connects to the proposal's core concept of virtual clusters as independent failure domains. However issue with it is that the policy's side effects are much wider than the failed cluster — the decision is really about what happens to everything *else*. A name scoped to the failed cluster (`clusterIsolation`) understates the blast radius of the choice. `proxy.onVirtualClusterStopped.serve` makes the scope explicit: when a cluster stops, what does the proxy serve? The values (`none | successful`) directly answer that question without ambiguity, where `strict | lenient` or `halt | continue` describe the policy's character rather than its outcome.

On structure: we're using a block rather than a flat key deliberately. `onVirtualClusterStopped` is scoped to a specific state transition, and the block leaves room for future sibling keys alongside `serve` (e.g. a future proposal might add escalation behaviour) without restructuring the configuration. We should make this rationale more explicit in the proposal text — will do that.

proposals/016-virtual-cluster-lifecycle.md
SamBarker: Good suggestion on per-VC drain timeouts — the proposal now does exactly that. Different clusters can have different workload profiles, and per-cluster config also provides the right foundation for reload where only the affected cluster drains.

On making draining opt-out: draining is best-effort rather than a guarantee. It operates within the Netty shutdown window, which is the hard backstop — if a drain exceeds it, Netty force-closes everything. So the cost of having draining enabled is bounded: at worst it uses time that Netty was going to wait anyway. Making it opt-out adds configuration surface without a clear benefit, since disabling it just means connections close immediately rather than getting a chance to complete gracefully.

The proposal also now acknowledges that some operations (notably long-polling consumers) will routinely exceed any reasonable drain timeout. Synthesising early responses to unblock these is a possible future enhancement, but would require its own proposal as it introduces a new signal to the filters.

hrishabhg: Thanks for accepting the per-VC `drainTimeout`.

 On the other point, my reading of the updated design is that `drainTimeout` must be ≤ `network.proxy.shutdownTimeout` (default 15s), since Netty's graceful shutdown is the hard outer boundary. If so, the default `drainTimeout` of 60s exceeds that bound which gives a misleading impression that drain can run longer than Netty allows. Implementation should have warning log for this and default should be set as 15s.

The proposal should clarify the relationship between `drainTimeout` and `shutdownTimeout`. As I understand it, `drainTimeout` is the lower bound (how long we attempt to let in-flight requests complete) and `shutdownTimeout` is the upper bound (Netty's hard stop). This also means `drainTimeout: 0s` is a natural opt-out — skip straight to Netty's shutdown, no backpressure or drain phase needed. This framing would help implementors understand that precise in-flight tracking isn't necessary — drain is best-effort within the window, and Netty force-closes at the upper bound regardless.

hrishabhg (changes requested):

tombentley (approved):
Many thanks @SamBarker

PR #92, opened 2026-04-08
Strimzi v1 migration proposal

k-wall:
Kroxylicious has a Strimzi integration feature that references the Kafka CR to automatically configure the upstream.  This feature currently relies on the Strimzi's v1beta2 API.  Strimzi is in the process of phasing out support for v1beta2, in favour of a new v1 API. 

This proposal  describes a way that the Kroxylicious operator may be changed to use the v1 API.

### DIFF
--- a/proposals/xxx-strimzi-v1-api-migration.md
+++ b/proposals/xxx-strimzi-v1-api-migration.md
@@ -1,0 +1,130 @@
+# Proposal xxx -Migrate Kroxylicious Operator to Strimzi v1 API
+
+Migrate the Kroxylicious Operator to use the Strimzi Kafka CR `v1` API, dropping support for `v1beta2`.
+
+## Current situation
+
+The Kroxylicious Operator's `KafkaService` CR supports a `strimziKafkaRef` field that allows the proxy to reference an upstream Kafka cluster provided by Strimzi. The operator currently reads the Strimzi `Kafka` CR at version `v1beta2`.
+
+Specifically, the operator reads a small subset of the `Kafka` object:
+- The `listeners` array from both `spec` and `status`
+
+This allows the operator to discover connection details (bootstrap addresses, TLS configuration) for Strimzi-managed Kafka clusters.
+
+## Motivation
+
+Strimzi is evolving its Kubernetes API to `v1`. The timeline is as follows:
+
+- **Strimzi 0.49.0** (September 2025) – introduced the `v1` API alongside `v1beta2`
+- **Strimzi 0.52.0** – will remove support for `v1beta2` entirely
+
+Currently, the Kroxylicious Operator will break when Strimzi 0.52.0 is released, as it will no longer be able to read `Kafka` resources at the `v1beta2` API version.
+
+The Kroxylicious Operator must be updated to support Strimzi 0.52.0 and beyond. If this is not done, the Kroxylicious
+integration will cease to work.  TODO: will the operator actually break completely? 
+
+## Proposal
+
+Migrate the Kroxylicious Operator to read the Strimzi `Kafka` CR at version `v1` only, dropping support for `v1beta2`.
+This will be done for the next Kroxylicious release (currently planned as v0.21.0).
+
+### Approach
+
+1. Update the operator's Kubernetes client configuration to request `Kafka` resources at API version `v1` instead of `v1beta2`
+2. Update any type references or model classes to use the `v1` schema
+3. Verify that the fields used by the operator (`spec.listeners`, `status.listeners`) remain compatible
+
+The [Strimzi v1 API proposal](https://github.com/strimzi/proposals/blob/main/113-Strimzi-v1-CRD-API-and-1.0.0-release.md#kafka) confirms that the listener-related fields used by the Kroxylicious Operator are not impacted by the migration from `v1beta2` to `v1`.
+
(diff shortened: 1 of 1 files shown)


proposals/xxx-strimzi-v1-api-migration.md
```diff
@@ -0,0 +1,130 @@
...
+- The `listeners` array from both `spec` and `status`
+
+This allows the operator to discover connection details (bootstrap addresses, TLS configuration) for Strimzi-managed Kafka clusters.
+
+## Motivation
+
+Strimzi is evolving its Kubernetes API to `v1`. The timeline is as follows:
+
+- **Strimzi 0.49.0** (September 2025) – introduced the `v1` API alongside `v1beta2`
+- **Strimzi 0.52.0** – will remove support for `v1beta2` entirely
```
ppatierno: There won't be any 0.52.0. The next release will be 1.0.0.

proposals/xxx-strimzi-v1-api-migration.md
```diff
@@ -0,0 +1,130 @@
...
+1. **Strimzi 0.48.0 is already 6+ months old** (as of April 2026). The hope is most users will have upgraded. 
+
+2. **Availability of easy workaround**.  Any user unable to upgrade to a newer version of Strimzi, can simply configure the `KafkaService` to point at the Kafka cluster directly (avoiding reliance on the Strimzi integration feature).
+
+3. **Simpler implementation and maintenance.** Supporting a single API version keeps the codebase clean and avoids the complexity of runtime version detection, fallback logic, and testing multiple code paths. Using strongly-typed `Kafka` CR objects (rather than `GenericKubernetesResource`) reduces the risk of runtime errors and improves developer experience.
+
+### Migration path for users
+
+Users running Strimzi 0.48.0 or earlier must:
+1. Upgrade Strimzi to 0.49.0 or later
```
ppatierno: I didn't try but ...
When just upgrading to 0.49.0, the Strimzi operator supports both v1beta2 and v1 so it means that a user can leave a Kafka custom resource to be v1beta2. Meanwhile the Kroxy operator is using the Strimzi API v1. While I would expect it to work (the custom resource is stored as v1beta2 but also v1 is served, as defined in the CRDs), I think it would be better for the users to be fully compatible and also using the migration tool to migrate CRs from v1beta2 to v1.

proposals/xxx-strimzi-v1-api-migration.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Compatibility
+
+### Breaking change
+
+**This proposal introduces a breaking change:** the updated operator will not support Strimzi versions prior to 0.49.0.
+
+Specifically:
+- **Strimzi 0.48.0 and earlier** (released September 2025 and before) will no longer be compatible
+- Users running these versions must upgrade Strimzi to 0.49.0 or later before upgrading the Kroxylicious Operator
```
tombentley: I wonder if this would be better communicated as a compatibility matrix?

tombentley (approved):
Thanks Keith. 

I generally am not in favour of changes which break compatibility. However in this case of a relatively new feature, which is unlikely to be in use in production (lacking TLS support), and with an easy workaround which comes without loss of functionality, I think it's justifiable.

ppatierno (approved):

k-wall:
This is what we see if the Kroxylicious is expecting v1beta2 while ther API server isn't serving it.

```
Caused by: io.javaoperatorsdk.operator.OperatorException: java.util.concurrent.ExecutionException: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://192.168.66.3:8443/apis/kafka.strimzi.io/v1beta2/kafkas?resourceVersion=0. Message: Not Found.
```

proposals/xxx-strimzi-v1-api-migration.md
```diff
@@ -0,0 +1,137 @@
...
+## Motivation
+
+Strimzi is evolving its Kubernetes API to `v1`. The timeline is as follows:
+
+- **Strimzi 0.49.0** (September 2025) – introduced the `v1` API alongside `v1beta2`
+- **Strimzi 1.0.0** (the release that is scheduled to follow 0.51.0) – will remove support for `v1beta2` entirely
+
+Currently, the Kroxylicious Operator will break when Strimzi 1.0.0 is released, as it will no longer be able to read `Kafka` resources at the `v1beta2` API version.
+
+The Kroxylicious Operator must be updated to support Strimzi 1.0.0 and beyond. If this is not done, the Kroxylicious integration with Strimzi will cease to work.
```
k-wall, line 23:
```suggestion
The Kroxylicious Operator must be updated to support Strimzi 1.0.0 and beyond. If this is not done, the Operator will fail completely when Strimzi CRDs are present without v1beta2 support.  It will fail at startup.  It uses the existence of the API group `kafka.strimzi.io` to determine if Strimzi is present (which will pass), but then the Informer will fail because the `/apis/kafka.strimzi.io/v1beta2/kafkas` endpoint will return an error).  It is sufficient to cause the Operator to crash.
```

k-wall: I've updated the proposal based on experimentation.

robobario (approved):
LGTM, thanks @k-wall

PR #93, opened 2026-04-08
93 - Correlated request/response data API

robobario:
Adds a proposal for a framework API to enable Filters to stash some Contextual data for a Request, that it can then access later when processing the Response.

### DIFF
--- a/proposals/093-correlated-request-response-data-api.md
+++ b/proposals/093-correlated-request-response-data-api.md
@@ -1,0 +1,65 @@
+# 93 - Request/Response Correlated Data API
+
+Provide a mechanism within the Kroxylicious Filter API to allow filters to stash state during the Request phase and retrieve it during the corresponding Response phase without manual state management.
+
+## Current situation
+
+Currently, filters that need to associate data from a request with its future response (e.g., `AuthorizationFilter` or `EntityIsolationFilter`) must manually manage an internal `Map`. Typically, the filter stashes data using the Kafka `correlationId` as a key during the request path and removes ("pops") it during the response path.
+
+Each Filter has so far had to hand-roll this code.
+
+## Motivation
+
+Manual state management for correlated data is error-prone and introduces several risks:
+* **Memory Leaks:** If a response never returns or a filter fails to "pop" the data, the internal map grows indefinitely.
+* **Edge Case Complexity:** Authors must be aware of and manually account for "zero-ack" Produce requests (which have no response) and scenarios where requests are dropped or short-circuited via the Filter API.
+* **Boilerplate:** Every filter author must reimplement the same "store-and-retrieve" logic, leading to inconsistent implementations.
+
+## Proposal
+
+The proposal introduces a framework-managed "Correlated Data" store scoped to a single filter instance and a single Request/Response lifecycle.
+
+### API Changes
+To ensure type safety and prevent invalid states, the API will be updated to allow attaching data only when a request is being forwarded. This is best achieved by introducing a more fluent builder pattern.
+
+**Request Path:**
+Currently `FilterContext#forwardRequest` returns a terminal stage.
+
+Instead of a simple terminal stage, we introduce a `forwardRequestBuilder` to handle the transition:
+```java
+return context.requestFilterResultBuilder()
+              .forwardRequestBuilder(header, request)
+              .withCorrelatedData(myStateObject)
+              .completed();
+```
+
+By including it in the command object from the Filter to the Framework we make it sympathetic
+to asynchronously populated correlated data.
+
+**Response Path:**
+The context made available to the response filter will provide access to the stashed object:
+```java
+Optional<MyStateObject> data = context.correlatedData();
+```
+
+### Implementation Considerations
+* **Scoping:** The data is strictly scoped to the filter instance that created it. It is not shared with other filters in the chain to avoid implicit coupling.
+* **Lifecycle Management:** The framework becomes responsible for "popping" the data. If a request is a zero-ack Produce request, the framework will proactively discard the stashed data to prevent leaks. If the Response Filter does not access the data, the framework will handle discarding the stashed data.
+* **Data Structure:** Initially, the API will support a single state object. Since filter authors can define a custom `Record` or `POJO` to hold multiple values, a complex Key-Value map is deemed unnecessary at this stage.
+
+## Affected/not affected projects
+
+* **Affected:** `kroxylicious-api` (interface changes), `kroxylicious-runtime` (implementation of state management).
+* **Not Affected:** `kroxylicious-filters` (until they are refactored to use the new API), external integration tests.
+
+## Compatibility
+
+* **Backward Compatibility:** The `forward` methods are retained as a naive forward is a common action when a Filter does not want to take any action. The convenience is worth carrying the two mechanisms for forwarding.
+* **Forward Compatibility:** By using a builder pattern (`forwardRequestBuilder`), we create a path for future extensions.
+
+## Rejected alternatives
+
+* **Downstream Shared State:** Attaching data that is visible to other filters. This was rejected as it introduces "spooky action at a distance" and makes filter ordering too brittle. We prefer explicit message manipulation or tags for inter-filter communication.
+* **API decoupled from the Filter Command:** You could alternatively offer methods on the FilterContext like `context.pushCorrelatedData(x)`, but then you have to consider more edge-cases like Filters invoking it
+at unpredictable times from uncontrolled threads.
+* **Simple Method Overloading:** Adding `forward(header, request, data)` was rejected as it does not scale well if we need to add more parameters in the future and makes the `FilterContext` API cluttered.
--- a/proposals/xxx-contextual-request-response-data-api.md
+++ b/proposals/xxx-contextual-request-response-data-api.md
@@ -1,0 +1,63 @@
+# Request/Response Contextual Data API
+
+Provide a mechanism within the Kroxylicious Filter API to allow filters to stash state during the Request phase and retrieve it during the corresponding Response phase without manual state management.
+
+## Current situation
+
+Currently, filters that need to associate data from a request with its future response (e.g., `AuthorizationFilter` or `EntityIsolationFilter`) must manually manage an internal `Map`. Typically, the filter stashes data using the Kafka `correlationId` as a key during the request path and removes ("pops") it during the response path.
+
+## Motivation
+
+Manual state management for correlated data is error-prone and introduces several risks:
+* **Memory Leaks:** If a response never returns or a filter fails to "pop" the data, the internal map grows indefinitely.
+* **Edge Case Complexity:** Authors must manually account for "zero-ack" Produce requests (which have no response) and scenarios where requests are dropped or short-circuited via the Filter API.
+* **Boilerplate:** Every filter author must reimplement the same "store-and-retrieve" logic, leading to inconsistent implementations.
+
+## Proposal
+
+The proposal introduces a framework-managed "Contextual Data" store scoped to a single filter instance and a single Request/Response lifecycle.
+
+### API Changes
+To ensure type safety and prevent invalid states, the API will be updated to allow attaching data only when a request is being forwarded. This is best achieved by introducing a more fluent builder pattern.
+
+**Request Path:**
+Currently `FilterContext#forward` returns a terminal stage.
+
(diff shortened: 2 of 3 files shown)


proposals/093-correlated-request-response-data-api.md
```diff
@@ -0,0 +1,63 @@
+# Request/Response Contextual Data API
+
+Provide a mechanism within the Kroxylicious Filter API to allow filters to stash state during the Request phase and retrieve it during the corresponding Response phase without manual state management.
+
+## Current situation
+
+Currently, filters that need to associate data from a request with its future response (e.g., `AuthorizationFilter` or `EntityIsolationFilter`) must manually manage an internal `Map`. Typically, the filter stashes data using the Kafka `correlationId` as a key during the request path and removes ("pops") it during the response path.
```
tombentley, line 7: Worth stating explicitly that filters typically each roll their own code for this.

proposals/xxx-contextual-request-response-data-api.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Current situation
+
+Currently, filters that need to associate data from a request with its future response (e.g., `AuthorizationFilter` or `EntityIsolationFilter`) must manually manage an internal `Map`. Typically, the filter stashes data using the Kafka `correlationId` as a key during the request path and removes ("pops") it during the response path.
+
+## Motivation
+
+Manual state management for correlated data is error-prone and introduces several risks:
+* **Memory Leaks:** If a response never returns or a filter fails to "pop" the data, the internal map grows indefinitely.
+* **Edge Case Complexity:** Authors must manually account for "zero-ack" Produce requests (which have no response) and scenarios where requests are dropped or short-circuited via the Filter API.
```
tombentley:
```suggestion
* **Edge Case Complexity:** Authors must be aware of and manually account for "zero-ack" Produce requests (which have no response) and scenarios where requests are dropped or short-circuited via the Filter API.
```

proposals/xxx-contextual-request-response-data-api.md
```diff
@@ -0,0 +1,63 @@
...
+
+**Response Path:**
+The context made available to the response filter will provide access to the stashed object:
+```java
+Optional<MyStateObject> data = context.contextualData();
+```
+
+### Implementation Considerations
+* **Scoping:** The data is strictly scoped to the filter instance that created it. It is not shared with other filters in the chain to avoid implicit coupling.
+* **Lifecycle Management:** The framework becomes responsible for "popping" the data. If a request is a zero-ack Produce request, the framework will proactively prevent or discard the stashed data to prevent leaks. If the Response Filter does not access the data, the framework will handle discarding the stashed data.
```
tombentley: >proactively prevent or discard

which is it?

proposals/xxx-contextual-request-response-data-api.md
```diff
@@ -0,0 +1,63 @@
...
+To ensure type safety and prevent invalid states, the API will be updated to allow attaching data only when a request is being forwarded. This is best achieved by introducing a more fluent builder pattern.
+
+**Request Path:**
+Currently `FilterContext#forward` returns a terminal stage.
+
+Instead of a simple terminal stage, we introduce a `forwardRequest` to handle the transition:
+```java
+return context.requestFilterResultBuilder()
+              .forwardRequest(header, request)
+              .withContextualData(myStateObject)
```
tombentley: I like the fluent API, but I'm not keen on the method name. I don't think a typical developer, even one with some knowledge of the Kafka protocol, would know what "context" meant without looking at the javadoc. 

Using the word "correlation" or "correlated" would at least be a hint that this was something to do with `correlationId`. So _perhaps_ something like `withCorrelatedData` is better?

But even that doesn't really say what it does clearly. However, that following that path suggests something like `savingStateForResponse`, which seems like a bit of a mouthful.

tombentley (commented):
Thanks Rob.

I left a few comments. 

I also have a slightly larger point:

Currently we assume all filters are stateful (in the sense of keeping conversational state in their fields). We've sometimes talked about the value of being able to declare a filter as being stateless. And when we knew a filter was stateless we would be able to use a `@Stateless` (in the Netty sense) version of `FilterHandler` to invoke it. Having those mechanics would mean stateless filters would require constant memory, rather than memory proportional to the number of connections. 

One thing compromise we might be making by doing what is proposed is that we have no way of knowing whether a filter will actually use this new mechanism or not. The proposal compromises the type safety of the builder API a bit for supporting such _explicitly stateless_ filters in the future: Such filters would have to _not_ use this API (in addition to not having any fields and so on). From a purely API-focussed PoV it would be a little bit nicer if we could have used static typing to avoid stateless filters having access to request builders with these stateful methods _at all_. 

I don't really think that's feasible with the current filter API, however. Therefore what you're proposing here is OK. But it might be a good idea to call out this tradeoff explicitly in the proposal text.

proposals/xxx-contextual-request-response-data-api.md
robobario: I've updated it to `discard`. If we made it eagerly throw an exception on trying to stash data for a zero-ack Produce request, then the author (presumably unaware of zero-ack's special-ness) might not encounter this problem until they've put their Filter out there in the world. It feels like discarding will cause less heartache.

proposals/xxx-contextual-request-response-data-api.md
robobario: Have updated to correlated language

proposals/xxx-correlated-request-response-data-api.md
```diff
@@ -0,0 +1,65 @@
...
+**Response Path:**
+The context made available to the response filter will provide access to the stashed object:
+```java
+Optional<MyStateObject> data = context.correlatedData();
+```
+
+### Implementation Considerations
+* **Scoping:** The data is strictly scoped to the filter instance that created it. It is not shared with other filters in the chain to avoid implicit coupling.
+* **Lifecycle Management:** The framework becomes responsible for "popping" the data. If a request is a zero-ack Produce request, the framework will proactively discard the stashed data to prevent leaks. If the Response Filter does not access the data, the framework will handle discarding the stashed data.
+* **Data Structure:** Initially, the API will support a single state object. Since filter authors can define a custom `Record` or `POJO` to hold multiple values, a complex Key-Value map is deemed unnecessary at this stage.
```
tombentley, line 48: I think you can be a little more assertive here. We have no intention of extending this, so let's not give the impression that we're maybe expecting to.

```suggestion
* **Data Structure:** The API will support a single state object. Filter authors can define a custom `Record` or `POJO` to hold multiple values, a complex Key-Value map is deemed unnecessary.
```

tombentley (approved):
Thanks @robobario, I think this looks good.

proposals/093-correlated-request-response-data-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+Each Filter has so far had to hand-roll this code.
+
+## Motivation
+
+Manual state management for correlated data is error-prone and introduces several risks:
+* **Memory Leaks:** If a response never returns or a filter fails to "pop" the data, the internal map grows indefinitely.
```
k-wall, line 14: Or even the io.kroxylicious.proxy.filter.ResponseFilter#shouldHandleResponse could even say no

proposals/093-correlated-request-response-data-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+**Response Path:**
+The context made available to the response filter will provide access to the stashed object:
+```java
+Optional<MyStateObject> data = context.correlatedData();
+```
+
+### Implementation Considerations
+* **Scoping:** The data is strictly scoped to the filter instance that created it. It is not shared with other filters in the chain to avoid implicit coupling.
+* **Lifecycle Management:** The framework becomes responsible for "popping" the data. If a request is a zero-ack Produce request, the framework will proactively discard the stashed data to prevent leaks. If the Response Filter does not access the data, the framework will handle discarding the stashed data.
```
k-wall, line 47: I'm happy with this approach, but did you consider making it an error to try to store state in a zero-ack produce-request.   should it be a rejected alternative?

proposals/xxx-correlated-request-response-data-api.md
```diff
@@ -0,0 +1,65 @@
...
+**Response Path:**
+The context made available to the response filter will provide access to the stashed object:
+```java
+Optional<MyStateObject> data = context.correlatedData();
+```
+
+### Implementation Considerations
+* **Scoping:** The data is strictly scoped to the filter instance that created it. It is not shared with other filters in the chain to avoid implicit coupling.
+* **Lifecycle Management:** The framework becomes responsible for "popping" the data. If a request is a zero-ack Produce request, the framework will proactively discard the stashed data to prevent leaks. If the Response Filter does not access the data, the framework will handle discarding the stashed data.
+* **Data Structure:** The API will support a single state object. Filter authors can define a custom `Record` or `POJO` to hold multiple values, a complex Key-Value map is deemed unnecessary.
```
k-wall: Do we want to say anything about stashing the request itself?   A filter may stash the request, but there is no guarantee that request will carry any modifications made by filters upsteam of this one.

proposals/093-correlated-request-response-data-api.md
```diff
@@ -0,0 +1,65 @@
...
+              .completed();
+```
+
+By including it in the command object from the Filter to the Framework we make it sympathetic
+to asynchronously populated correlated data.
+
+**Response Path:**
+The context made available to the response filter will provide access to the stashed object:
+```java
+Optional<MyStateObject> data = context.correlatedData();
```
k-wall, line 42: what happens if the request filter calls this method?  is it an error?

k-wall (commented):
couple of comments, but otherwise LGTM

proposals/093-correlated-request-response-data-api.md
tombentley: I think that's a bad idea because when a filter developer forgets to gets the acks=0 case (and we've been guilty of that ourselves!) they release something which will break with an exception. 

But there is a deeper question there: We're assuming that this state that we're keeping around for the response is only really using memory. But what if there are other resources in there, that are invisible to the runtime. Like a socket/connection/file descriptor? So I wonder if we should require that the data they associate should actually be `Closable`. That way we can guarantee that the runtime will fully clean it up even if the filter developer has forgotten that acks=0 is a thing.

Wdyt @robobario ?

proposals/xxx-correlated-request-response-data-api.md
robobario: true, stashing the request would likely be dangerous, like you might be earlier in the chain than an isolation/virtualization filter and have entity names change under you. 

There's not a lot we could do to prevent it, the user can pass through an arbitrary object that could have a request reference (or reference to some sub-member of Request) somewhere in it. I guess we'd have to rely on javadoc describing this limit and suggesting `.duplicate()` to take a copy of the request.

proposals/093-correlated-request-response-data-api.md
robobario: then in the simple case they'd have to implement a no-op close in their record, and be forced to wrap simple data types in a Closable. Maybe it would be enough to optionally support it, so we close at pop-time if it's `Closable`.

proposals/093-correlated-request-response-data-api.md
robobario: mm annoying, throwing is probably a good idea to signal it's being misused. I wonder if we should look into splitting a request and response context. We already have the awkwardness that `FilterContext` carries methods for both request and response (`forward(..)` etc).

k-wall:

```bash
git mv proposals/xxx-correlated-request-response-data-api.md proposals/093-correlated-request-response-data-api.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 93 - /; t; s/^# /# 93 - /}' proposals/093-correlated-request-response-data-api.md && rm proposals/093-correlated-request-response-data-api.md.bak
git add proposals/093-correlated-request-response-data-api.md
git push
```

proposals/xxx-correlated-request-response-data-api.md
k-wall: Duplicate would need to come with a health warning as it for Produce/Fetch, the record content the ByteBuffer is duplicated.  The bytes themselves are not.  Whether that's good or not, depends on the use-case.

k-wall (approved):
Couple of nits to iron out but LGTM

PR #96, opened 2026-04-13
Reworking proxy configuration

tombentley:
This proposal covers a significant reworking of the Kroxylicious proxy's configuration system.

### DIFF
--- a/proposals/015-config2-multi-file-plugin-configuration.md
+++ b/proposals/015-config2-multi-file-plugin-configuration.md
@@ -138,12 +138,8 @@ name: encrypt
 type: io.kroxylicious.filter.encryption.RecordEncryption
 version: v1
 config:
-  kms:
-    type: io.kroxylicious.kms.service.KmsService
-    name: vault-kms
-  selector:
-    type: io.kroxylicious.filter.encryption.config.KekSelectorService
-    name: my-selector
+  kms: vault-kms
+  selector: my-selector
 ```
 
 The `name` field must match the filename (without the `.yaml` extension). This is validated during
@@ -172,39 +168,51 @@ or an in-memory representation in tests.
 ### Plugin references
 
 When a plugin's configuration needs to refer to another plugin instance (e.g. a filter that
-uses a KMS service), the versioned config type uses `PluginReference<T>`:
+uses a KMS service), the versioned config type declares its dependencies by implementing
+`HasPluginReferences`. The YAML representation of the reference is entirely up to the plugin
+author — typically a bare instance name string, since the plugin interface type is statically
+known:
 
 ```java
 public record RecordEncryptionConfigV1(
-        PluginReference<KmsService> kms,
-        PluginReference<KekSelectorService> selector,
+        String kms,
+        String selector,
         Map<String, Object> experimental,
         UnresolvedKeyPolicy unresolvedKeyPolicy)
     implements HasPluginReferences {
 
     @Override
     public Stream<PluginReference<?>> pluginReferences() {
-        return Stream.of(kms, selector);
+        return Stream.of(
+                new PluginReference<>(KmsService.class.getName(), kms),
+                new PluginReference<>(KekSelectorService.class.getName(), selector));
     }
 }
 ```
 
-`PluginReference<T>` is a record of `(String type, String name)` where `type` is the fully
-qualified plugin interface name and `name` is the instance name. In the YAML, this serialises
-as a `{type, name}` map — a uniform, explicit pattern for all cross-file references.
+`PluginReference<T>` is a runtime type — a record of `(String type, String name)` where `type`
+is the fully qualified plugin interface name and `name` is the instance name. It is not a
+serialisation type: it does not appear in the YAML. Plugin authors construct `PluginReference`
+values in their `pluginReferences()` implementation, combining the statically-known interface
+type with the instance name read from YAML.
 
-By introducing `PluginReference` we impose a consistent reference syntax on all plugin
-developers. This is a deliberate trade-off: it constrains the shape of versioned config types,
-but in return every dependency is visible to the framework without requiring type-specific
-introspection. The alternative — scanning config objects for fields that look like references,
-or using reflection to discover dependencies at runtime — would be fragile, plugin-specific,
-and invisible to tooling.
+This separation is deliberate. A plugin like `RecordEncryption` knows statically that its `kms`
+field always refers to a `KmsService`. Forcing the user to write
+`kms: {type: io.kroxylicious.kms.service.KmsService, name: vault-kms}` in the YAML would be
+redundant — the `type` can only ever be one thing. By keeping `PluginReference` out of the
+YAML, each plugin is free to choose the most natural representation for its references.
 
-This is not a breaking change. Plugin developers adopt `PluginReference` fields only when they
+The `HasPluginReferences` interface is the contract that makes this work. The runtime calls
+`pluginReferences()` to discover cross-file dependencies, build the dependency graph, and
+validate that all referenced plugin instances exist. The alternative — scanning config objects
+for fields that look like references, or using reflection to discover dependencies — would be
+fragile, plugin-specific, and invisible to tooling.
+
+This is not a breaking change. Plugin developers implement `HasPluginReferences` only when they
 introduce a new versioned config type (e.g. `RecordEncryptionConfigV1`). The legacy config type
 (`RecordEncryptionConfig`) continues to use `@PluginImplName` / `@PluginImplConfig` as before.
 Because the config versioning mechanism (`@Plugin(configVersion = "v1", configType = ...)`)
-allows both old and new config types to coexist, the `PluginReference` pattern is adopted
+allows both old and new config types to coexist, the `HasPluginReferences` pattern is adopted
 incrementally, version by version.
 
 ### Dependency graph and referential integrity
@@ -245,8 +253,8 @@ Versioned filter configs obtain their dependencies from the registry:
 ResolvedPluginRegistry registry = context.resolvedPluginRegistry()
         .orElseThrow(() -> new PluginConfigurationException(
                 "v1 config requires a ResolvedPluginRegistry"));
-KmsService kmsPlugin = registry.pluginInstance(KmsService.class, kmsRef.name());
-Object kmsConfig = registry.pluginConfig(kmsRef.type(), kmsRef.name());
+KmsService kmsPlugin = registry.pluginInstance(KmsService.class, v1.kms());
+Object kmsConfig = registry.pluginConfig(KmsService.class.getName(), v1.kms());
 ```
 
 ### Dual `@Plugin` annotations and version dispatch
--- a/proposals/015-config2-multi-file-plugin-configuration.md
+++ b/proposals/015-config2-multi-file-plugin-configuration.md
@@ -29,16 +29,16 @@ The proxy resolves these at runtime by discovering `@PluginImplName`-annotated f
 the named plugin implementation via `ServiceLoader`, and deserialising the corresponding
 `@PluginImplConfig` field into the plugin's config type. This has several consequences:
 
-- **No reuse**: two filters that need the same KMS must each embed a copy of the KMS configuration.
-- **No independent versioning**: a filter's config schema is monolithic. Changing the KMS config
-  format requires the filter to understand and migrate its nested portion.
-- **No schema validation**: there is no declarative way to validate plugin configurations before
-  the proxy starts.
+- **No reuse**: two filters that need the same KMS must each embed a copy of the same KMS configuration.
+- **No independent versioning**: a proxy config file is monolithic and lacks explicit version numbers,
+  so every plugin needs to honour backwards compatibility of its configuration essentially for ever.
 - **No dependency ordering**: the proxy discovers plugin dependencies implicitly at initialisation
   time, with no upfront visibility into the dependency graph.
+- **No schema validation**: there is no declarative way to validate plugin configurations 
+  except by trying to start the proxy.
 - **Opaque to tooling**: since plugin configurations are embedded inside filter configs, external
   tools (Kubernetes operators, CI validators, config linters) cannot inspect or validate them
-  independently.
+  independently (see [#2436](https://github.com/kroxylicious/kroxylicious/issues/2436).
 
 ## Motivation
 
@@ -135,7 +135,7 @@ Each plugin instance file specifies `name`, `type`, `version`, and optionally `c
 
 ```yaml
 name: encrypt
-type: RecordEncryption
+type: io.kroxylicious.filter.encryption.RecordEncryption
 version: v1
 config:
   kms:
@@ -149,14 +149,21 @@ config:
 The `name` field must match the filename (without the `.yaml` extension). This is validated during
 loading. Including `name` in the file content ensures that plugin instance files are
 self-describing: a file can be understood without knowing its path, which matters for debugging,
-code review, and non-filesystem configuration sources like ConfigMaps.
+code review, and non-filesystem configuration sources.
+
+In old-style proxy configuration files we allowed unqualified type names, 
+which were resolved based on the loadable plugins.
+If we kept with that choice in the new format then plugin configurations would not be fully self-describing. 
+So to facilitate the unambiguous sharing of configurations the `type` is fully qualified in the new format.
 
 ### Snapshot abstraction
 
 A `Snapshot` interface abstracts the configuration source. It provides access to `proxy.yaml`
 content and the set of plugin instances grouped by interface. This allows the same resolution
-logic to work whether configuration comes from a filesystem directory, a Kubernetes ConfigMap,
-or an in-memory representation in tests.
+logic to work whether configuration comes from a filesystem directory, 
+an in-memory representation in tests or some other source.
+
+The `Snapshot` interface will not be public API, but internal to the runtime.
 
 ### Plugin references
 
--- a/proposals/015-config2-multi-file-plugin-configuration.md
+++ b/proposals/015-config2-multi-file-plugin-configuration.md
@@ -42,20 +42,54 @@ the named plugin implementation via `ServiceLoader`, and deserialising the corre
 
 ## Motivation
 
-The Kubernetes operator needs to map custom resources to proxy configuration. A multi-file layout
-where each plugin instance is a separate document is a natural fit: each custom resource maps to
-one file. This also enables:
+There are five specific drawbacks of the current configuration system that motivate this change.
 
-- **Shared plugin instances**: a single KMS plugin instance can be referenced by multiple filters,
-  reducing configuration duplication and ensuring consistent credentials.
-- **Declarative dependency validation**: the system can detect missing or circular dependencies
-  before any plugin is initialised.
-- **Schema validation**: each plugin version can ship a JSON Schema, enabling early validation by
-  the proxy, the operator, CI pipelines, and IDE tooling.
-- **Independent evolution**: plugin authors can evolve their config schemas independently, using
-  Kubernetes-style version progression (v1alpha1 -> v1beta1 -> v1).
-- **Migration tooling**: a deterministic mapping from legacy to multi-file format enables automated
-  migration.
+### 1. Per-plugin JSON schemas are impossible today
+
+It is impossible to write a single JSON schema for the configuration file, because what is allowed
+depends on which plugins are present at runtime. But having schemas is desirable for documentation,
+editor assistance, and automated validation. By giving each plugin instance its own file, each
+plugin version can ship its own JSON schema. These per-plugin schemas can be published to schema
+catalogues (e.g. schemastore.org) and consumed by IDEs and CI tools independently of the proxy.
+
+### 2. Lack of uniformity in how plugins are configured
+
+Each plugin point embeds its plugins differently. Filter factories are explicitly named (via
+`NamedFilterDefinition`), but other plugins like `KmsService` are anonymous because they are
+referenced once by their parent. This inconsistency extends across the codebase: filters,
+authorisers, KMS services, subject builders, and key selectors all follow slightly different
+patterns. A uniform model where every plugin instance is a named, versioned resource eliminates
+this inconsistency.
+
+### 3. The runtime cannot understand inter-plugin dependencies
+
+The lack of uniformity means the runtime has no meaningful understanding of the dependency
+relationships between plugins. This matters for dynamic reloading: when a plugin configuration
+changes, the runtime needs to know which virtual clusters are affected. Today, dependencies are
+discovered implicitly at initialisation time, with no upfront visibility into the graph. Explicit
+dependency declarations enable the runtime to reason about the impact of configuration changes.
+
+Note that the runtime does not (and should not) know about all plugin types. Plugins are allowed
+to have plugins of their own (e.g. `RecordEncryption` depends on `KmsService` and
+`KekSelectorService`, which are not known to the runtime). This rules out approaches that add
+top-level definitions for each plugin type (e.g. `kmsDefinitions`), because the runtime cannot
+enumerate plugin types it does not know about.
+
+### 4. Plugin instances are not individually identifiable
+
+Eventually we may need to build a control plane for clusters of proxies. Plugin instances would be
+important entities in the API of that control plane. For this to work, each plugin instance must
+have a unique identity. The current system gives names to filter definitions but not to the plugins
+they embed. A model where every plugin instance has a name, a type, and a version makes each one
+individually addressable.
+
+### 5. No principled way to evolve plugin configuration
+
+None of the configurations (the proxy's or any of the plugins') are explicitly versioned. A plugin
+developer can deprecate properties, but cannot remove them in a controlled manner where the API
+evolution is obvious to the user. Explicit version strings (following Kubernetes conventions:
+v1alpha1, v1beta1, v1) give plugin developers a principled mechanism for config schema evolution,
+including support for multiple concurrent versions during migration periods.
 
 ## Proposal
 
--- a/proposals/015-config2-multi-file-plugin-configuration.md
+++ b/proposals/015-config2-multi-file-plugin-configuration.md
@@ -131,9 +131,10 @@ micrometer:
   - common-tags
 ```
 
-Each plugin instance file specifies `type`, `version`, and optionally `config`:
+Each plugin instance file specifies `name`, `type`, `version`, and optionally `config`:
 
 ```yaml
+name: encrypt
 type: RecordEncryption
 version: v1
 config:
@@ -144,6 +145,11 @@ config:
     type: io.kroxylicious.filter.encryption.config.KekSelectorService
     name: my-selector
 ```
+
+The `name` field must match the filename (without the `.yaml` extension). This is validated during
+loading. Including `name` in the file content ensures that plugin instance files are
+self-describing: a file can be understood without knowing its path, which matters for debugging,
+code review, and non-filesystem configuration sources like ConfigMaps.
 
 ### Snapshot abstraction
 
--- a/proposals/015-config2-multi-file-plugin-configuration.md
+++ b/proposals/015-config2-multi-file-plugin-configuration.md
@@ -135,7 +135,7 @@ Each plugin instance file specifies `name`, `type`, `version`, and optionally `c
 
 ```yaml
 name: encrypt
-type: RecordEncryption
+type: io.kroxylicious.filter.encryption.RecordEncryption
 version: v1
 config:
   kms:
@@ -150,6 +150,17 @@ The `name` field must match the filename (without the `.yaml` extension). This i
 loading. Including `name` in the file content ensures that plugin instance files are
 self-describing: a file can be understood without knowing its path, which matters for debugging,
 code review, and non-filesystem configuration sources like ConfigMaps.
+
+The `type` field must be a fully qualified class name. The legacy single-file format allows
+short names (e.g. `RecordEncryption`), but the config2 format requires the FQCN
+(e.g. `io.kroxylicious.filter.encryption.RecordEncryption`). This makes each plugin file
+completely self-describing: its interpretation does not depend on which plugins happen to be
+loadable by a particular proxy binary. Short names are ambiguous when two plugins share a
+simple class name; FQCNs eliminate that problem entirely.
+
+The loss of brevity is mitigated by JSON Schema: each plugin version's schema can constrain
+`type` using `const` or `enum`, giving editors code completion and validation without requiring
+the user to type the full name.
 
 ### Snapshot abstraction
 
--- a/proposals/015-config2-multi-file-plugin-configuration.md
+++ b/proposals/015-config2-multi-file-plugin-configuration.md
@@ -6,9 +6,14 @@ cross-file dependency references, and optional JSON Schema validation.
 
 ## Current situation
 
-Today, all configuration lives in a single YAML file. Filter definitions embed their nested plugin
-configurations inline using `@PluginImplName` / `@PluginImplConfig` annotation pairs on the config
-record's constructor parameters:
+Today, all configuration lives in a single YAML file. 
+
+At places where plugins can be configured, `@PluginImplName` / `@PluginImplConfig` annotation pairs are used
+so that a plugin's configuration can be included in the YAML tree.
+Plugins are allowed to have plugins of their own.
+
+For example, the proxy runtime uses those annotations for the `type` and `config` properties, and 
+the `RecordEncryption` filter uses them for `kms` + `kmsConfig` and for `selector` and `selectorConfig`:
 
 ```yaml
 filterDefinitions:
@@ -25,31 +30,44 @@ filterDefinitions:
         template: "KEK_${topicName}"
 ```
 
-The proxy resolves these at runtime by discovering `@PluginImplName`-annotated fields, looking up
+The proxy runtime resolves these during startup by discovering `@PluginImplName`-annotated fields, looking up
 the named plugin implementation via `ServiceLoader`, and deserialising the corresponding
-`@PluginImplConfig` field into the plugin's config type. This has several consequences:
+`@PluginImplConfig` field into the plugin's config type. 
 
-- **No reuse**: two filters that need the same KMS must each embed a copy of the KMS configuration.
-- **No independent versioning**: a filter's config schema is monolithic. Changing the KMS config
-  format requires the filter to understand and migrate its nested portion.
-- **No schema validation**: there is no declarative way to validate plugin configurations before
-  the proxy starts.
-- **No dependency ordering**: the proxy discovers plugin dependencies implicitly at initialisation
-  time, with no upfront visibility into the dependency graph.
-- **Opaque to tooling**: since plugin configurations are embedded inside filter configs, external
-  tools (Kubernetes operators, CI validators, config linters) cannot inspect or validate them
-  independently.
+The existing APIs work, and plugins-to-plugins has proven to be useful.
 
 ## Motivation
+
+Although it works, the current configuration APIs have several consequences:
+
+1. **No reuse**: The model is tree-like. For example, if two filters need the same KMS 
+  then each must embed a copy of the KMS configuration.
+2. **Poor documentation**: a plugin developer ought to be able to document the YAML accepted by their plugin, 
+  but this is difficult because the source of truth is Java code: End users cannot be assumed to know Java.
+3. **Opaque to tooling**: similarly, external tools (editors, config linters, CI validators, Kubernetes operators) 
+  cannot inspect or  validate plugin configurations. (See [2436](https://github.com/kroxylicious/kroxylicious/issues/2436))
+4. **No schema validation**: there is no declarative way to validate plugin configurations before
+  the proxy starts. (Somewhat related: [3267](https://github.com/kroxylicious/kroxylicious/issues/3267#issuecomment-4042031788)
+5. **No independent versioning**: As a plugin evolves there are limited facilities for end users 
+  and plugin developers to agree common expectations about a plugin configuration. The plugin 
+  implementation either accepts the YAML or it does not. If newer version of a plugin removes
+  support for a configuration property the user will likely be confused, because the configuration 
+  used to work. But the plug in has no effective way to communicate that the user is speaking 
+  the old language.
+6. **No dependency ordering**: the proxy discovers plugin dependencies implicitly at initialisation
+  time, with no upfront visibility into the dependency graph.
 
 There are five specific drawbacks of the current configuration system that motivate this change.
 
 ### 1. Per-plugin JSON schemas are impossible today
 
-It is impossible to write a single JSON schema for the configuration file, because what is allowed
-depends on which plugins are present at runtime. But having schemas is desirable for documentation,
-editor assistance, and automated validation. By giving each plugin instance its own file, each
-plugin version can ship its own JSON schema. These per-plugin schemas can be published to schema
+Items 2, 3 and 4 are quite tightly related, and all relate to poorer user experience.
+
+However it is currently impossible to write a single JSON schema for the configuration file, because what is allowed
+depends on which plugins are present at runtime. 
+By giving each plugin instance its own file, each
+plugin version can ship its own JSON schema to be used to validate that file. 
+These per-plugin schemas can be published to schema
 catalogues (e.g. schemastore.org) and consumed by IDEs and CI tools independently of the proxy.
 
 ### 2. Lack of uniformity in how plugins are configured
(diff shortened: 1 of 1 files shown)


proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+The runtime will validate that `configVersion` is unique across all the `@Plugin` annotations on a plugin implementation class.
+
+```java
+@Plugin(configType = RecordEncryptionConfig.class) // The legacy config schema
+@Plugin(configVersion = "v1", configType = RecordEncryptionConfigV1.class) // The new config schema, called v1
+public class RecordEncryption<K, E>
+    implements FilterFactory<Object, SharedEncryptionContext<K, E>> {
+```
+
+When implementing a plugin like `FilterFactory` which accepts a type parameter representing the configuration type, `Object` will be used as the type argument.
```
tombentley, line 126: I should call out another design decision: I'm not attempting to change the Java `FilterFactory` API (or other plugin interface APIs consumed by the runtime) to pass the `version` string as an argument to `initialize()`. We could do that, of course, and then the implementation of `initialize()` could switch on the `version` instead of the type of the configuration.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+### JSON Schema validation
+
+Plugin developers can ship a JSON Schema for each config version at:
+
+```
+META-INF/kroxylicious/schemas/{PluginSimpleName}/{version}.schema.yaml
+```
+
+When a schema is present, the runtime will validate the raw configuration against it before deserialisation. 
+This is opt-in: if no schema exists, the config is accepted without validation.
```
tombentley: TBH, I'm in two minds about this. I think there's a good argument that we should force plugins which use versions to have a schema (they can always provide a trivial schema which accepts any JSON).

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+contract, where the interface FQCN is the natural key.
+
+The loss of brevity is mitigated by JSON Schema: each plugin version's schema can constrain
+`type` using `const` or `enum`, giving editors code completion and validation without requiring
+the user to type the full name.
+This mitigation relies on plugin developers adding such constraints.
+
+### JSON Schema validation
+
+Plugin developers can ship a JSON Schema for each config version at:
```
tombentley, line 228: We need to define what version of JSON Schema to use. For the sake of simplicity I think we should mandate one version. I think there are realistically two options:

* The most recent version
* The version supported for use in Kubernetes CRDs

The former is attractive because it's the most expressive. The latter is attractive because it would be simpler for contributors to only have to know one version of JSON Schema, and it would also make it easier to represent plugins as custom resources. I've intentionally ducked the question of how this proposal would affect our Kubernetes API (because this proposal is already big), but we need to decide this JSON Schema version question in this proposal.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+The `HasPluginReferences` interface is the contract that allows the runtime to resolve the plugin instance graph. 
+The runtime calls
+`pluginReferences()` to discover cross-file dependencies, build the dependency graph, and
+validate that all referenced plugin instances exist. 
+
+This is not a breaking change. Plugin developers implement `HasPluginReferences` only when they
+introduce a new versioned config type (e.g. `RecordEncryptionConfigV1`). The legacy config type
+(`RecordEncryptionConfig`) continues to use `@PluginImplName` / `@PluginImplConfig` as before.
+Because the config versioning mechanism (`@Plugin(configVersion = "v1", configType = ...)`)
```
tombentley, line 345: We should consider whether the runtime should validate that `HasPluginReferences.class.isAssignableFrom(configType)` when `!configVersion.isEmpty()`.

That:

* Would force plugin developers adopting the API to make use of this functionality. (Make it harder for them to forget/not know about `HasPluginReferences`).
* Allow us to switch to declaring `Class<? extends HasPluginReferences> configType()` in `@Plugin` in the future, when/if we eventually drop supporting the old-style configuration system where `configVersion = ""`.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+### Dependency graph and referential integrity
+
+The runtime builds a dependency graph from `HasPluginReferences` declarations. Versioned config
+types implement this interface to enumerate their `PluginReference` values. The runtime then
+validates the graph using DFS-based topological sort, detecting:
+
+- **Dangling references**: a plugin references an instance that does not exist.
+- **Cycles**: A depends on B depends on A.
+
```
tombentley, line 359: We should note that it's not an error for a plugin to be declared but not reachable from the `proxy.yaml`.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+### `@Stateless` annotation
+
+Plugin developers can apply a new `@Stateless` annotation to their plugin implementation.
+This signals to the runtime that an instance may be shared across multiple consumers. 
+
+The decision about whether to actually do that belongs to the end user who controls the configuration used to refer to a plugin instance.
+The plugin instance YAML can set `shared: true` to enable this. 
+
+The framework will reject `shared: true` on plugins not annotated `@Stateless`.
```
tombentley, line 397: This would benefit from an example and a bit more detail around the lifecycle.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+### Resolved plugin registry
+
+A `ResolvedPluginRegistry` is built during plugin resolution.
```
tombentley: Make clear that this is new public API

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+It instantiates non-filter plugin instances (KMS services, authorisers, etc.) eagerly, in
+dependency order, at startup. These are shared services whose lifecycle is tied to the proxy
+process.
+
+Filter instances are different: the proxy creates a new filter chain for each client connection,
+so filter instances cannot be pre-created at startup. Instead, filter factories obtain their
+dependencies from the registry when the runtime invokes them:
+
+```java
```
tombentley, line 411: We should add the surrounding code so that the reader understands that this is code in `FilterFactory.initialize()`

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+                             String name,
+                             String type,
+                             String version,
+                             boolean shared,
+                             long generation) {}
+```
+
+#### Change detection: generation numbers
+
+Each resource in the `Snapshot` carries a **generation number** — a monotonically increasing
```
tombentley: To avoid confusion we should either define "resource" or stick to "plugin instance".

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+value that changes when the resource content changes. Comparing generation numbers rather than
+data bytes avoids:
+
+- Non-determinism from keystore salting (two logically equivalent stores can have different bytes).
+- Security concerns around comparing key material.
+- Expensive byte-level comparison of large blobs.
+
+For the operator, generation maps to Kubernetes `metadata.generation` / `resourceVersion`.
+For filesystem deployments, it is derived from file modification time.
+
```
tombentley, line 527: Per-plugin instance generations are all well and good, but this proposal should address how this would fit with dynamic reloading. In particular the scope of dynamic reloading is the virtual cluster, so we should explain how two Snapshots (a currently active one, and a new incoming one) would be compared to determine which virtual clusters need reloading.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+### Non-YAML resources
+
+For the `Snapshot` to fully encapsulate a proxy's configuration state, it must include these resources alongside YAML-based plugin configs.
```
tombentley: The reordering of sections has gone awry: What are "these resources"?

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+Not all plugin instance data is YAML. 
+* The existing AclAuthorizer uses a non-YAML text format for its ACL rules.
+* Plugins commonly accept `Tls` objects for specifying secret data via Java KeyStores. 
+
+It is possible to provide plugin data in any format.
+However, this mainly exists to allow KeyStores to have first class support within the configuration system.
+For the filesystem Snapshot non-YAML data is provided in a separate file in the same directory as the YAML metadata file.
```
tombentley: Show an example directory tree like we did at the start. We do provide an example below, but it should be here where the idea is first introduced. Alternatively avoid this much detail at this point in the doc and move it to where the example is now.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+ 
+#### Text-based resources: inline YAML
+
+For text-based resources like ACL rules, YAML multi-line syntax (`|`) is simplest.
```
tombentley, line 570: Simplest except when there's an error in the rules: Antlr will report line numbers that are indexed on the first line of the multi-line string, not on the first line of the YAML file.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+version: v1
+config:
+  rules: |
+    allow User with name = "alice" to READ Topic with name = "foo";
+    otherwise deny;
+```
+ 
+#### Filesystem layout
+
+Binary resources use a sidecar `.data` file alongside the YAML metadata file:
```
tombentley: Requiring the `.data` file extension sucks. It makes it much harder to work on the filesystem using tools that depend on file extensions. All we actually need is collision avoidance between the metadata files (which are `.yaml`) and the data files. For the YAML data case we're already using just a single combined YAML file, so there is no collision. So we can probably allow any extension provided that if we have a `foo.yaml` there's only a single other file whose name without extension is `foo`.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+#### `@ResourceType` annotation
+
+For binary resource types, the plugin implementation declares its binary format via a `@ResourceType` annotation:
+
+```java
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ResourceType {
+    Class<? extends ResourceSerializer<?>> serializer();
+    Class<? extends ResourceDeserializer<?>> deserializer();
```
tombentley, line 597: I guess we should show the declaration of these types

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+    io.kroxylicious.proxy.tls.KeyMaterialProvider/
+      my-keystore.yaml      # metadata: name, type, version
+      my-keystore.data      # binary keystore bytes
+    io.kroxylicious.proxy.filter.FilterFactory/
+      encrypt.yaml           # metadata + YAML config
+```
+
+#### `@ResourceType` annotation
+
+For binary resource types, the plugin implementation declares its binary format via a `@ResourceType` annotation:
```
tombentley: Say which package this is in.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+```java
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ResourceType {
+    Class<? extends ResourceSerializer<?>> serializer();
+    Class<? extends ResourceDeserializer<?>> deserializer();
+}
+```
+
+The runtime discovers the serde from the annotation via reflection — the same pattern used for `@Plugin(configType = ...)`.
```
tombentley, line 613: Say explicitly whether that `@ResourceType` would applied to the plugin implementation, i.e. a sibling to `@Plugin`.

But also: Is the model correct? By making the `@ResourceType` a top level annotation we're saying that the plugin data is always of that single binary type. i.e. we're preventing the case where a binary file format is changed/switched as part of the evolution of the plugin. Perhaps those serializer and deserializer methods should actually be on `@Plugin`?

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+}
+```
+
+For the filesystem-based snapshot, passwords are stored in a `passwords.yaml` file in the
+configuration directory. This file maps resource names to their passwords:
+
+```yaml
+# passwords.yaml
+my-keystore: changeit
+```
```
tombentley, line 649: This is a bit weak. It precludes the possibility of different secret content coming from different trust zones, because someone has to assemble a single file where all the passwords are in the clear. 
It would be better if this allows people to add their key store and their password each as its own file. That way FS permissions could be used so that Bob can add his keystore, but not decrypt Alice's keystore when he does so.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+Other `Snapshot` implementations may source passwords differently — for example, the operator
+could read them from Kubernetes Secrets, or from Vault.
```
tombentley: It would be nice to flesh out the Kubernetes and Vault ideas in more detail. It's not really explained very clearly, I feel.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+Other `Snapshot` implementations may source passwords differently — for example, the operator
+could read them from Kubernetes Secrets, or from Vault. 
+
+
+#### Store type auto-detection
+
+The keystore format need not be user-specified. 
+Common formats have distinctive magic bytes (JKS: `0xFEEDFEED`, PKCS12: ASN.1 `0x30`, PEM: `-----BEGIN`). 
+The deserialiser probes the bytes and selects the appropriate format. 
+PKCS12 is the default keystore type since Java 9; JKS is legacy.
```
tombentley, line 664: If we support file extensions other than just `.data` then maybe the file extension could also be used. It's not clear whether the consumer of the bytes gets passed just the file content (bytes) or can also see the extension.

Are the magic bytes and trial deserialization hard coded to just the common stores? Should probably explain in a bit more detail how this might work for other keystore types, like EJKS.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+
+/** Provides TLS trust material (trusted CA certificates). */
+public interface TrustMaterialProvider {
+    java.security.KeyStore trustStore();
+}
+```
+
+Consumers reference these by name in their versioned config, constructing `PluginReference`
+values in `pluginReferences()` — the same pattern used for all other cross-file references.
+
```
tombentley, line 699: We need an example here. Probably record encryption & kms, since that's been used extensively for examples so far.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,710 @@
...
+### Migration tool
+
+A `migrate-config` CLI subcommand converts legacy single-file configurations into the multi-file
+format:
+
+```
+kroxylicious migrate-config -i legacy-config.yaml -o output-dir/
```
tombentley, line 740: Do we really want to bolt keywords onto the existing `kroxylicious` app that's the entry point to running the proxy? It could be the top of a slippery slope, where we add utility functionality which ends up bloating the container image for the proxy, and including functionality which we might not want to be available to the proxy. 

As an alternative, we could have a separate `kroxyctl` which had this functionality (and other similar code that's not needed for the proxy process).

k-wall:
Discussed at Community Call 23rd April. Tom is looking for feedback.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+
+## Motivation
+
+Although it works, the current configuration model has several drawbacks that motivate a redesign.
+
+### 1. No reuse of plugin configurations
+
+The configuration model is tree-like. If two filters need the same KMS
+then each must embed a copy of the KMS configuration. There is no way to define a plugin
+instance once and reference it from multiple consumers.
```
k-wall, line 48: YAML does have anchors and aliases

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+
+### 2. Per-plugin JSON schemas are impossible today
+
+A plugin developer ought to be able to document the YAML accepted by their plugin,
+but this is difficult because the source of truth is Java code: end users cannot be assumed to know Java.
+Similarly, external tools (editors, config linters, CI validators, Kubernetes operators)
+cannot inspect or validate plugin configurations (see [2436](https://github.com/kroxylicious/kroxylicious/issues/2436)),
+and there is no declarative way to validate plugin configurations before
+the proxy starts (somewhat related: [3267](https://github.com/kroxylicious/kroxylicious/issues/3267#issuecomment-4042031788)).
```
k-wall, line 57: > there is no declarative way to validate plugin configurations before the proxy starts

that one aspect is solvable without redesigning the config model.  Tools such as httpd and squid have `config-check` command line parameter.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+`type` using `const` or `enum`, giving editors code completion and validation without requiring
+the user to type the full name.
+This mitigation relies on plugin developers adding such constraints.
+
+### JSON Schema validation
+
+Plugin developers can ship a JSON Schema for each config version at:
+
+```
+META-INF/kroxylicious/schemas/{PluginSimpleName}/{version}.schema.yaml
```
k-wall, line 231: puzzled why you chose simple name, when you made the case for FQCN above.  Why the discrepancy?

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+
+#### Schema requirement for versioned plugins
+
+For versioned plugin configurations (`configVersion` is non-empty), a JSON schema is **required**. The plugin may provide a trivial schema that accepts any JSON object if it does not wish to constrain the configuration further. For legacy (unversioned) configurations, schema validation remains opt-in.
+
+A test utility `SchemaValidationAssert` is provided in `kroxylicious-filter-test-support` so plugin authors can verify their schemas accept valid configs.
+
+Discovery of schemas by external tooling depends on plugin developers publishing the schema so that it can be discovered by tools.
+One way to do that is to publish the schema on a public webserver, allowing end users to make use of JSON Schema's `$schema` keyword.
+Editors often support downloading such schemas.
```
k-wall, line 253: I must confess, I've actually rarely done this myself.  Maybe I'm an outlier?  Is this something others commonly do?

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+A model where every plugin instance has a name, a type, and a version would make each one individually addressable.
+
+### 6. No principled way to evolve plugin configuration
+
+None of the configurations (the proxy's or any of the plugins') are explicitly versioned. 
+A plugin developer can deprecate properties, but cannot remove them in a controlled manner where the API evolution is obvious to the user. 
+As a plugin evolves, the plugin implementation either accepts the YAML or it does not. 
+If a newer version of a plugin removes support for a configuration property the user will likely be confused, because the configuration used to work — but the plugin has no effective way to communicate that the user is speaking the old language. 
+Explicit version strings (following Kubernetes conventions: v1alpha1, v1beta1, v1) would give plugin developers a principled mechanism for config schema evolution, including support for multiple concurrent versions during migration periods.
+
```
k-wall, line 105: One place I see the configuration schemas being very useful is validating the documentation.   As our documentation gets larger, keeping the configuration snippets accurate will become increasing hard.  Tool support will help greatly.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+We make a different choice because our needs are different: the proxy cannot start with a dangling KMS reference or a cyclic dependency. 
+Fail-fast validation at configuration load time is preferable to discovering broken references at runtime when the first message arrives. 
+The cost is that the runtime must understand — and validate — the dependency structure, which is why `HasPluginReferences` exists as an explicit contract rather than relying on introspection or convention.
+
+The runtime does not (and should not) statically know about all plugin types. 
+Plugins can depend on other plugins that the runtime has never heard of (e.g. `RecordEncryption` depends on `KmsService` and `KekSelectorService`). The `HasPluginReferences` interface lets each config type declare its own dependencies without requiring the runtime to enumerate every possible plugin interface.
+
+### `@Stateless` annotation
+
+Plugin developers can apply a new `@Stateless` annotation to their plugin implementation.
```
k-wall, line 377: Shared plugins sounds useful.  KMS caches could become chunky and sharing them across many VKC seems useful.  However, it seems to me to be a feature we could defer from the initial implementation.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+                        "v1 config requires a ResolvedPluginRegistry"));
+        KmsService kmsPlugin = registry.pluginInstance(KmsService.class, cfg.kms());
+        Object kmsConfig = registry.pluginConfig(KmsService.class.getName(), cfg.kms());
+        return initializeV1(context, cfg, kmsPlugin, kmsConfig);
+    }
+    // ... legacy path
+}
+```
+
+### Snapshot abstraction
```
k-wall, line 427: to self: haven't followed this properly eyt

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+1. Parses the legacy configuration via `ConfigParser`.
+2. Reflects on each filter's legacy config type to discover `@PluginImplName` / `@PluginImplConfig`
+   pairs on the canonical constructor parameters.
+3. Extracts each nested plugin into its own file under `plugins.d/{interface}/{name}.yaml`.
+4. Rewrites the filter config to use `PluginReference`-style maps, selecting the best available
+   config version.
+5. Writes `proxy.yaml` by passing through the raw YAML (stripping `filterDefinitions`, replacing
+   `micrometer` with instance name references).
+
+**Open question:** whether the migration tool should be a subcommand of the proxy entrypoint or
```
k-wall, line 754: I'd vote for the separate CLI

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+  passwords.yaml
+  plugins.d/
+    io.kroxylicious.proxy.tls.KeyMaterialProvider/
+      my-keystore.yaml      # metadata: name, type, version
+      my-keystore.p12        # binary keystore bytes (any extension except .yaml)
+    io.kroxylicious.proxy.filter.FilterFactory/
+      encrypt.yaml           # metadata + YAML config (no sidecar needed)
+```
+
+For a plugin instance named `foo`, the metadata is always in `foo.yaml`. If a sidecar data file exists, it must be the only other file in the directory whose name without extension is `foo` (e.g. `foo.p12`, `foo.jks`, `foo.pem`). Any extension is permitted (except `.yaml` which is taken), which allows filesystem tools that depend on extensions to work naturally. The runtime validates that at most one sidecar file exists per plugin instance.
```
k-wall, line 566: What about plugins that need more than one data file?  Limiting to a single file seems quite restrictive.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+
+#### Text-based resources: inline YAML
+
+For text-based resources like ACL rules, YAML multi-line syntax (`|`) is simplest. 
+The versioned config type can use a `String` field instead of a file path:
+
+```yaml
```
k-wall, line 573: the example here is not yaml

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+    
+    /**
+     * Returns the password for a named resource, if one is configured. Passwords are
+     * provided out-of-band from the resource data for security.
+     *
+     * @param resourceName the name of the resource
+     * @return the password as a char array, or {@code null} if no password is configured
+     */
+    @Nullable
+    char[] resourcePassword(String resourceName);
```
k-wall, line 639: resources like keystores have a store password and key password, which maybe distinct.

k-wall:
I see the attractions of the proposals - the modularity, re-use, schema validation are all desirable features.  But this already looks like a significant piece of work and it'll take significant effort. There's bound to be issues that crop up during the implementation. That's all opportunity cost where we don't get to work on features (that are more directly applicable to use-cases).  Do we really need it?  Components like Httpd and Tomcat seem to have done fine without something analogous.

I'm floating around +0.5

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: It does, but that's more of a syntax level thing. The API we have for the config file (Jackson) would essentially duplicate those `JsonNode` or mapped POJOs, so we would see two objects which were `.equals`, but not `==`. You can use `@JsonIdentityInfo` to make the references first class. But if you're seriously proposing that as an alternative then I think you should elaborate how you'd expect it to work.

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: I think you're commenting on part of one sentence, and not really accounting for the full context in the paragraph as a whole. Sure, we can provide a tool. But that won't be built into anyone's editor, and won't provide anything like code completion, and doesn't address the documentation part.

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: That's what I said on line 52.

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: Claude chose that and I was too tired to argue. I agree with you.

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: You might find, if you're using intellij for instance, that you're getting the benefits of schemas published to schemastore without even knowing it. It is, admittedly, more painful to configure an unpublished schema manually. 

Personally I find the code completion and documentation part of using schemas at least as useful as the validation part.

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: It is restrictive to some extent. Maybe we should make it more flexible, but can you provide some more concrete use cases where it would be a benefit?

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: I believe it is. What makes you say that?

proposals/015-config2-multi-file-plugin-configuration.md
tombentley: Correct, but look at the direction of travel: JKS is not recommented for new applications (since Java 9).
What became the new default, PKCS#12, in its Java implementation, more or less expects the passwords to be the same. Quoting gemini, the "best practice":

> If you are using PKCS12 in Java, you should keep the store password and the key password identical. If you provide a different password for the key, many non-Java tools (and even some older Java application servers like Tomcat or JBoss) may fail to read the private key because they expect the passwords to match. If you need distinct security layers, it is better to use file-system-level permissions rather than relying on different PKCS12 passwords.

tombentley (commented):
Thanks for taking a look @k-wall. 

> I see the attractions of the proposals - the modularity, re-use, schema validation are all desirable features. But this already looks like a significant piece of work and it'll take significant effort. There's bound to be issues that crop up during the implementation. That's all opportunity cost where we don't get to work on features (that are more directly applicable to use-cases). Do we really need it? Components like Httpd and Tomcat seem to have done fine without something analogous.

One thing you've not pointed to there is the configuration versioning and schema evolution part. Our conversations about Kroxylicious 1.0 have landed on that as an important thing to resolve. This proposal addresses that, but you're saying we don't need it, or at least are unwilling to commit to the effort needed to implement this proposal. So I'm wondering whether you see a conflict in holding both those positions, or whether you think we can solve them in some other way, or just that 1.0 is not something we need.

k-wall:

```bash
git mv proposals/015-config2-multi-file-plugin-configuration.md proposals/096-config2-multi-file-plugin-configuration.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 96 - /; t; s/^# /# 96 - /}' proposals/096-config2-multi-file-plugin-configuration.md && rm proposals/096-config2-multi-file-plugin-configuration.md.bak
git add proposals/096-config2-multi-file-plugin-configuration.md
git push
```

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+Each plugin instance file specifies `name`, `type`, `version`, and optionally `config`:
+
+```yaml
+# plugins.d/io.kroxylicious.proxy.filter.FilterFactory/encrypt.yaml
+name: encrypt
+type: io.kroxylicious.filter.encryption.RecordEncryption
+version: v1
+config:
+  kms: vault-kms
+  selector: my-selector
```
robobario, line 199: Since we are so simpatico with kubernetes in this proposal, would it be worth taking `metadata` and `spec` objects? 

```
version: v1
type: io.kroxylicious.filter.encryption.RecordEncryption
metadata:
  name: encrypt
spec:
  kms: vault-kms
  selector: my-selector
```

I guess there's risks to _looking very like a CRD_ but not really. But I quite like metadata as a logical home for any future fields that are meta if we wanted to label plugins in future

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+    allow User with name = "alice" to READ Topic with name = "foo";
+    otherwise deny;
+```
+
+A limitation of inline text is that parsers (e.g. ANTLR) will report error line numbers relative to the start of the multi-line string, not the enclosing YAML file. This may complicate debugging for large rule sets.
+
+#### `@ResourceType` annotation
+
+For binary resource types, the plugin implementation declares its binary format via a `@ResourceType` annotation.
+`@ResourceType` is applied to the plugin implementation class, alongside `@Plugin`:
```
robobario, line 588: feel like I'm missing something with ResourceType annotations.

Are we saying that there are two classes of Plugin Configuration, one where the "configuration" is uniquely defined by the YAML config, like all our Filters.

Then we would have a second, new class of Plugin Configuration, where the "configuration" is some binary and we would use the supplied serdes to deserialize their binary to a type that can be used with the plugin api, so the deserialized object would be passed to the plugin as configuration, the same way they take the deserialized YAML currently.

The aim being that authors would migrate to defining a plugin for every file (or other environmental resource) they need to load, so the runtime knows the full tree? And you're proposing in the other sections that text formats should be handled as inline YAML config so that we are basically just focusing on TLS files (and I'd suggest API keys too).

Maybe that KMS Api key example is a good one. Would the user define:

```
name: my-vault
type: io.kroxylicious.kms.vault.VaultKms
version: v1
config:
  apiKey: vault-key
  tls:
    trust: vault-trust
---
name: vault-key
type: io.kroxylicious.token.FileToken
version: v1
---
name: vault-trust
type: io.kroxylicious.security.TrustStoreFile
version: v1
```
where the framework would handle taking `vault-key.token` and `vault-trust.jks` from their prescribed locations and building from there?

proposals/015-config2-multi-file-plugin-configuration.md
robobario, line 566: the naming requirements might be a fiddle too, thinking about kubernetes integration, if we wanted to use cert-man generated certs they might have some established naming conventions in the keys of their secret like `ca.crt`, `tls.crt`, `tls.key`. It might be easier for us to be flexible with the naming that have users try to bash other technology into the right naming.

proposals/015-config2-multi-file-plugin-configuration.md
```diff
@@ -0,0 +1,824 @@
...
+
+3. **Map affected filters to virtual clusters.** The affected filter instances are mapped to the virtual clusters whose filter chains reference them. Only those virtual clusters need to be restarted — unaffected clusters continue serving traffic.
+
+4. **Detect proxy-level changes.** Changes to `proxy.yaml` itself (virtual cluster definitions, gateways, default filter lists) feed into the `VirtualClusterChangeDetector` from Proposal 012.
+
+This approach also addresses part of the "plugin resource tracking" gap identified in Proposal 012. That proposal notes that the runtime cannot detect changes to external resources (keystores, ACL rule files) that plugins read during initialisation. The `Snapshot` abstraction makes the runtime aware of all plugin data — including binary sidecar files — so generation changes on those files are visible to the change detection pipeline without requiring plugins to opt in to resource tracking.
+
+
+### Non-YAML resources
```
robobario, line 543: It still requires Plugins to be well coded and not load other resources at runtime. So we trust Plugins Authors. Could we alternatively make all other files in a named directory part of the configuration for a plugin and give the Plugins a way to access these files at initialization time?

So if you had files:

```
config-dir/
  proxy.yaml
  passwords.yaml
  plugins.d/
    io.kroxylicious.authorizer.provider.acl.AclAuthorizerService/
      my-acl-authorizer.yaml
      my-acl-authorizer/
        my-rules.txt
```

In filter config you might say:

```yaml
  name: my-acl-authorizer
  type: io.kroxylicious.authorizer.provider.acl.AclAuthorizerService
  version: v1
  config:
    rules: my-rules.txt
```
The plugin could declare it needs additional resources via an interface:

```java
  public interface NeedsAdditionalResources {
      void initialize(ResourceContext ctx);
  }

  public interface ResourceContext {
      byte[] getResourceData(String name);
      String getResourcePassword(String name); // for encrypted resources
  }
```

And the implementation:

```java
  class AclAuthorizerService implements AuthorizerService, NeedsAdditionalResources {
      private AclAuthorizerConfigV1 config;
      private AclRules rules;

      public AclAuthorizerService(AclAuthorizerConfigV1 config) {
          this.config = config;
      }

      @Override
      public void initialize(ResourceContext ctx) {
          byte[] in = ctx.getResourceData(config.rules());
          this.rules = AclRules.parse(in);
      }
  }
```
To make dependencies explicit (similar to HasPluginReferences), we could add:

```java
  public interface HasAdditionalResources {
      Stream<String> additionalResourceNames();
  }

  public record AclAuthorizerConfigV1(String rules) implements HasAdditionalResources {
      @Override
      public Stream<String> additionalResourceNames() {
          return Stream.of(rules);
      }
  }
```

This would bring additional files into the runtime's awareness for:
  - Change detection and generation numbers
  - Validation (file exists before initialization)

Without introducing a second type of Plugin Configuration.

Maybe something like this could complement `@ResourceType` for plugins that aren't 1:1 with a binary file, like the ACL rules I think makes more sense to be reffed in the AclAuthorizer configuration, potentially alongside other configuration in the YAML or requiring inline YAML for things like the rules.

robobario:
From the other PR:

I'll admit that 96 doesn't do a great job of calling out all of its goals, but let me try to compare the two:

> | Goal | #96 | #110 |
> | -------|-------|------|
> |Explicit plugin config version | Yes | Yes |
> |Support config schema evolution | Yes | Yes |
> |Plugin identification | Yes | No |
> |Deeper runtime understanding of nested plugins | Yes | No |
> |Fine grained dependency analysys | Yes | No |
> |Better documentation of schema configs | Yes | No |
> |External validation of schema configs | Yes | No |
> |Editor support for writing configs | Yes | No |
> |Globally shared interpretation of proxy config | Yes | No |
> |Synergy with K8S | Stronger | Weaker |
> 
> It's clear that there's a very different level of ambition between the two proposals. My fear is that, out of a desire to see 1.0 happen sooner, we make decisions which undermine our ability to address some of the concerns in the left hand column. 
> 
> ...
> 
> So I think it would be helpful to hear if you think any of those goals are not actually worth striving for. If they are worth solving then I'd appreciate your review on 96: is it a good way of solving them (maybe there are better ways)? In the light of that we can figure out if 110 is the best way of doing something that's "good enough for 1.0", or if we can tweak 110 to be more compatible with 96.
> 
> I also think it's worth considering that https://github.com/kroxylicious/kroxylicious/pull/3691, which is a POC of the changes proposed in 96, is "only" a +7k change. Sure that's a lot, and it needs work on the tests and so on, so would likely grow a bit. But in the light of some of the recent big things we've merged, I don't think we should immediately be thinking that 96 is a huge piece of work that would take many months to implement.

I agree that a lot of those goals are worth striving for. Editor support for writing configs seems more like a bonus than something required. I don't know what "Globally shared interpretation of proxy config" means precisely, could you expand on that point?

Complete knowledge of the configuration tree and plugin relationships sounds like we'd be heading to a more satisfactory Hot Reload story, and could pull in knowledge of related files like rules files, truststores, api tokens so they are considered for reload.

I was wondering if multi-file might put people off in the quickstart and learning stages, so whether we should support some kind of convenience single-file mode containing a `plugins` array. But I think users of other proxy tech should be accustomed to the idea of decomposed configuration.

I guess your argument is that if we do all this now we don't have to carry the legacy format at 1.0.0, which is compelling.

Maybe one question mark against it is that we've had this word Control Plane bubbling up a bit, if the need for that increased maybe we'd be better off implementing an HTTP(or GRPC etc) control plane first and have the proxy consume that as it's sole configuration input, rather than reworking the internals of the proxy to the new format and then having to reimplement to control plane later. Some of the work could potentially be done by the control plane to ensure the configuration has referential integrity between plugins etc. I imagine that this proposal will correspond neatly to a Control Plane API too, we'd need a similar set of capabilities.

tombentley:
@robobario 

> I don't know what "Globally shared interpretation of proxy config" means precisely, could you expand on that point?

I mean if I share my config with you there is no ambiguity about what it means. e.g. Simple class names arg ambiguous and the ambiguity can only be resolved by knowing that plugins are available. FQCN is unambiguous.

proposals/015-config2-multi-file-plugin-configuration.md
k-wall: Fine. I wasn't aware it was considered best practice.

SamBarker:
Thanks for putting this together Tom — there's a lot of good thinking in here, particularly around `Snapshot`, plugin versioning, dependency tracking, and JSON Schema validation. Those are real problems worth solving.

I want to raise two structural concerns.

**1. The organising principle**

Working through Proposal 117 (the VirtualClusterConfigController SPI), we've arrived at a logical split in proxy configuration:

- **Process configuration** — management endpoints, metrics, admin, controller selection. Static, read once at startup, never changes without a restart.
- **Virtual cluster configuration** — virtual clusters, filters, plugin instances. Dynamic, delivered as a `Snapshot` via `reconfigure()`. This is the configuration that changes at runtime — everything in service of what users actually care about.

This split falls out naturally from Proposal 083's `reconfigure()` contract: it only applies virtual-cluster and filter configuration, and rejects changes to management/metrics/admin. The virtual cluster is our unit of blast radius.

The concern with the `plugins.d/` structure is that it's organised by a technical dimension — the plugin interface FQCN — rather than a dimension that means something to operators. An operator managing a production deployment doesn't think "I need to update the `io.kroxylicious.kms.service.KmsService` directory." They think "I need to update the encryption config for virtual cluster X." The filesystem layout should mirror the process/virtual-cluster split — what's static vs what changes — not the Java type system.

**2. Separating the configuration model from its materialisation**

I think the proposal may be conflating two concerns:

- **The abstract configuration model** — a DAG of versioned, typed plugin instances with explicit dependencies (`HasPluginReferences`), generation-based change detection, JSON Schema validation, dependency ordering, referential integrity. This is the critical stuff.
- **How that model is materialised** — files on disk in a particular directory structure, with naming conventions and sidecar files. This is an implementation of the model.

Right now the model is expressed *through* the filesystem layout — the `plugins.d/` directory structure, the filename-must-match-name rule, sidecar file conventions. That makes the filesystem the primary abstraction rather than one materialisation of a deeper one.

If we separate these, the abstract model becomes an API that different backends can implement. A filesystem layout is one backend (and probably the first one we ship). But a REST control plane that builds the plugin DAG via API calls is another. A Kubernetes operator that assembles the DAG from CRD resources is another. The `Snapshot` abstraction already hints at this separation — it abstracts the source — but the model itself (how plugins are identified, versioned, and wired together) should be defined independently of any particular backend.

This would also let Proposal 117's `VirtualClusterConfigController` work naturally with non-filesystem sources. A controller that talks to a REST control plane doesn't need to fake a `plugins.d/` directory — it builds the same abstract DAG from API responses and delivers it as a `Snapshot`.

The useful building blocks in this proposal — `@Plugin` versioning, `PluginReference`, `HasPluginReferences`, `ResolvedPluginRegistry`, generation numbers, JSON Schema validation — all belong in the abstract model. They don't depend on files. Pulling them out into a backend-agnostic layer would make the proposal's real contribution clearer and the filesystem layout a natural first implementation rather than the design itself.

PR #97, opened 2026-04-13
017, 018: AWS KMS credentials config restructure & EKS workload authentication

oleksiyp:
## Summary

Two design proposals for the AWS KMS Record Encryption filter:

- **017 — Credentials config restructure**: Groups all credential providers under a single `credentials` YAML node (`credentials.longTerm`, `credentials.ec2Metadata`, etc.) instead of flat top-level fields. Backward compatible — existing `longTermCredentials` / `ec2MetadataCredentials` YAML keeps working via migration in the record's compact constructor.

- **018 — EKS workload authentication (IRSA & Pod Identity)**: Adds two new credential providers for Amazon EKS pods. IRSA exchanges a projected OIDC token via STS `AssumeRoleWithWebIdentity`; Pod Identity calls the local Pod Identity Agent. Both return temporary credentials compatible with the existing SigV4 signing. Includes extraction of a shared async-refresh base class from `Ec2MetadataCredentialsProvider`. Depends on 017 for config structure.

Prompted by review feedback on the implementation PR kroxylicious/kroxylicious#3684.

References: kroxylicious/kroxylicious#1295, kroxylicious/kroxylicious#3684

### DIFF
--- a/proposals/017-aws-kms-credentials-config-restructure.md
+++ b/proposals/017-aws-kms-credentials-config-restructure.md
@@ -157,6 +157,6 @@ The `CredentialsProviderFactory` is simplified to dispatch solely from `config.c
 
 1. **Jackson polymorphic deserialization (`@JsonTypeInfo`)** — using a type discriminator inside `credentials` (e.g. `type: webIdentity`) would enforce single-provider semantics at the deserialization layer.  Rejected because it requires a discriminator property that differs from the key-per-provider style used elsewhere in Kroxylicious (Azure KMS `entraIdentity`, Fortanix `apiKeySession`) and makes the YAML less intuitive.
 
-2. **Break existing config without backward compat** — since the project is pre-1.0, breaking changes are nominally acceptable.  Rejected because existing users have working configurations and migration should be painless.  The cost of the backward-compat shim (a few lines in the compact constructor) is negligible compared to the user friction of a forced migration.
+2. **Break existing config without backward compat** — rejected because the project has a [deprecation policy](https://github.com/kroxylicious/kroxylicious/blob/main/DEV_GUIDE.md#deprecation-policy) that requires maintaining backward compatibility even pre-1.0.  The cost of the backward-compat shim (a few lines in the compact constructor) is negligible compared to the user friction of a forced migration.
 
 3. **Keep flat fields, just add more** — rejected because it scales poorly.  With four providers, the `Config` record would have four nullable credential fields interspersed with `endpointUrl`, `region`, and `tls`, making the YAML hard to read and the factory logic fragile.
--- a/proposals/017-aws-kms-credentials-config-restructure.md
+++ b/proposals/017-aws-kms-credentials-config-restructure.md
@@ -1,0 +1,162 @@
+# 017 - AWS KMS: credentials configuration restructure
+
+<!-- TOC -->
+* [017 - AWS KMS: credentials configuration restructure](#017---aws-kms-credentials-configuration-restructure)
+  * [Current situation](#current-situation)
+  * [Motivation](#motivation)
+  * [Proposal](#proposal)
+    * [New `credentials` configuration node](#new-credentials-configuration-node)
+    * [Backward compatibility for existing providers](#backward-compatibility-for-existing-providers)
+    * [Implementation](#implementation)
+  * [Affected/not affected projects](#affectednot-affected-projects)
+  * [Compatibility](#compatibility)
+  * [Rejected alternatives](#rejected-alternatives)
+<!-- TOC -->
+
+## Current situation
+
+The AWS KMS provider for the Record Encryption filter authenticates to AWS using one of two mechanisms, each configured as a separate top-level field on the `Config` record:
+
+```yaml
+kmsConfig:
+  endpointUrl: https://kms.us-east-1.amazonaws.com
+  longTermCredentials:          # option A
+    accessKeyId:
+      password: AKIA...
+    secretAccessKey:
+      password: wJalr...
+  region: us-east-1
+```
+
+```yaml
+kmsConfig:
+  endpointUrl: https://kms.us-east-1.amazonaws.com
+  ec2MetadataCredentials:       # option B
+    iamRole: KroxyliciousInstance
+  region: us-east-1
+```
+
+Both fields are nullable on the `Config` record.  `CredentialsProviderFactory` enforces at runtime that exactly one is non-null, rejecting configurations that specify zero or both.
+
+This structure works for two providers, but additional provider types are planned (see proposal 018).  Each new type would add another nullable top-level field and another branch in the mutual-exclusivity check.
+
+## Motivation
+
+1. **Scalability** — with four or more providers the flat-field approach results in a growing number of nullable fields on `Config` and an increasingly complex exclusivity check in the factory.
+2. **Clarity** — a single `credentials` key in the YAML makes it self-evident that exactly one authentication mechanism is configured, rather than requiring the user to know that `longTermCredentials` and `ec2MetadataCredentials` are mutually exclusive siblings.
+3. **Extensibility** — future credential providers (IRSA, Pod Identity, SSO, STS AssumeRole) can be added by inserting one field into a dedicated record, with no changes to the top-level `Config` shape.
+
+## Proposal
+
+### New `credentials` configuration node
+
+All credential providers are grouped under a single `credentials` key:
+
+```yaml
+kmsConfig:
+  endpointUrl: https://kms.us-east-1.amazonaws.com
+  credentials:
+    longTerm:
+      accessKeyId:
+        password: AKIA...
+      secretAccessKey:
+        password: wJalr...
+  region: us-east-1
+```
+
+```yaml
+kmsConfig:
+  endpointUrl: https://kms.us-east-1.amazonaws.com
+  credentials:
+    ec2Metadata:
+      iamRole: KroxyliciousInstance
+  region: us-east-1
+```
+
+The `credentials` object is a new record with one nullable field per provider.  The `CredentialsProviderFactory` enforces that exactly one field is non-null:
+
+```java
+public record CredentialsConfig(
+    @JsonProperty("longTerm")    @Nullable LongTermCredentialsProviderConfig longTerm,
+    @JsonProperty("ec2Metadata") @Nullable Ec2MetadataCredentialsProviderConfig ec2Metadata) {}
+```
+
+Future providers (proposal 018) will add fields to this record.
+
+### Backward compatibility for existing providers
+
+The existing flat-field configuration keys (`longTermCredentials`, `ec2MetadataCredentials`) are preserved as deprecated fields on `Config`.  The record's compact constructor transparently migrates them into the `credentials` node:
+
+```yaml
+# Old style — still works, deprecated
+kmsConfig:
+  endpointUrl: https://kms.us-east-1.amazonaws.com
+  longTermCredentials:
+    accessKeyId:
+      password: AKIA...
+    secretAccessKey:
+      password: wJalr...
+  region: us-east-1
+```
+
+Specifying both a deprecated flat field **and** the `credentials` node simultaneously is a configuration error.
+
+### Implementation
+
+The `Config` record gains a `credentials` field and retains the two existing flat fields for backward compatibility:
+
+```java
+public record Config(
+    @JsonProperty(value = "endpointUrl", required = true) URI endpointUrl,
+    @JsonProperty(value = "longTermCredentials") @Nullable LongTermCredentialsProviderConfig longTermCredentialsProviderConfig,
+    @JsonProperty(value = "ec2MetadataCredentials") @Nullable Ec2MetadataCredentialsProviderConfig ec2MetadataCredentialsProviderConfig,
+    @JsonProperty(value = "credentials") @Nullable CredentialsConfig credentials,
+    @JsonProperty(value = "region", required = true) String region,
+    @JsonProperty(value = "tls") @Nullable Tls tls) {
+
+    public Config {
+        Objects.requireNonNull(endpointUrl);
+        Objects.requireNonNull(region);
+        // Migrate deprecated flat fields into credentials node
+        var migrated = credentials;
+        if (longTermCredentialsProviderConfig != null) {
+            if (migrated != null) throw ...;
+            migrated = new CredentialsConfig(longTermCredentialsProviderConfig, null);
+        }
+        if (ec2MetadataCredentialsProviderConfig != null) {
+            if (migrated != null) throw ...;
+            migrated = new CredentialsConfig(null, ec2MetadataCredentialsProviderConfig);
+        }
+        credentials = migrated;
+    }
+}
+```
+
+The `CredentialsProviderFactory` is simplified to dispatch solely from `config.credentials()`.
+
+## Affected/not affected projects
+
+**Affected:**
+- `kroxylicious-kms-provider-aws-kms` — new `CredentialsConfig` record, modified `Config`, modified `CredentialsProviderFactory`
+- `kroxylicious-kms-provider-aws-kms-test-support` — `Config` constructor update
+- `kroxylicious-systemtests` — `Config` constructor update
+- `kroxylicious-docs` — updated configuration snippets, deprecation notes
+
+**Not affected:**
+- Other KMS providers (HashiCorp Vault, Azure Key Vault, Fortanix DSM, in-memory)
+- The Kroxylicious operator
+- Proxy runtime core
+
+## Compatibility
+
+- **Backward compatible:** existing YAML with `longTermCredentials` or `ec2MetadataCredentials` at the top level continues to work unchanged.
+- **Deprecation:** the flat-field style will emit a log warning encouraging migration to the `credentials` node.  Removal of the deprecated fields is deferred to a future major version.
+- **Forward compatible:** adding a new credential provider requires only a new field on `CredentialsConfig` — no changes to the top-level `Config` shape.
+
+## Rejected alternatives
+
+1. **Jackson polymorphic deserialization (`@JsonTypeInfo`)** — using a type discriminator inside `credentials` (e.g. `type: webIdentity`) would enforce single-provider semantics at the deserialization layer.  Rejected because it requires a discriminator property that differs from the key-per-provider style used elsewhere in Kroxylicious (Azure KMS `entraIdentity`, Fortanix `apiKeySession`) and makes the YAML less intuitive.
+
+2. **Break existing config without backward compat** — since the project is pre-1.0, breaking changes are nominally acceptable.  Rejected because existing users have working configurations and migration should be painless.  The cost of the backward-compat shim (a few lines in the compact constructor) is negligible compared to the user friction of a forced migration.
+
+3. **Keep flat fields, just add more** — rejected because it scales poorly.  With four providers, the `Config` record would have four nullable credential fields interspersed with `endpointUrl`, `region`, and `tls`, making the YAML hard to read and the factory logic fragile.
--- a/proposals/018-aws-kms-eks-authentication.md
+++ b/proposals/018-aws-kms-eks-authentication.md
@@ -1,0 +1,168 @@
+# 018 - AWS KMS: EKS workload authentication (IRSA & Pod Identity)
+
+<!-- TOC -->
+* [018 - AWS KMS: EKS workload authentication (IRSA & Pod Identity)](#018---aws-kms-eks-workload-authentication-irsa--pod-identity)
+  * [Current situation](#current-situation)
+  * [Motivation](#motivation)
+  * [Proposal](#proposal)
+    * [IRSA (AssumeRoleWithWebIdentity)](#irsa-assumerolewithwebidentity)
+      * [Request flow](#request-flow)
+      * [Configuration](#configuration)
+    * [EKS Pod Identity](#eks-pod-identity)
+      * [Request flow](#request-flow-1)
+      * [Configuration](#configuration-1)
+    * [Shared refresh infrastructure](#shared-refresh-infrastructure)
+    * [Environment variable defaults](#environment-variable-defaults)
+  * [Affected/not affected projects](#affectednot-affected-projects)
+  * [Compatibility](#compatibility)
+  * [Rejected alternatives](#rejected-alternatives)
+<!-- TOC -->
+
+## Current situation
+
+The AWS KMS provider for the Record Encryption filter supports two authentication mechanisms: long-term IAM credentials and EC2 instance metadata.  Per proposal 017, these are being restructured under a `credentials` configuration node.
+
+The existing `Credentials` interface already supports an optional `securityToken()` for temporary credentials, and `AwsV4SigningHttpRequestBuilder` adds the `X-Amz-Security-Token` header when present — so the signing layer is already ready for temporary credentials from STS.
+
+The implementation is AWS-SDK-free: all HTTP calls use `java.net.http.HttpClient` with a custom SigV4 implementation.
+
+Neither existing mechanism works for pods on **Amazon EKS**: long-term keys are a security anti-pattern, and EC2 IMDS is only reachable from EC2 instances.
+
+## Motivation
+
+Amazon EKS is the primary deployment target for Kroxylicious on AWS.  AWS supports two pod-level authentication mechanisms:
+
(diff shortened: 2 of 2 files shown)


proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+#### Configuration
+
+Registered under `credentials.podIdentity` (per proposal 017):
+
+```yaml
+credentials:
+  podIdentity:
+    credentialsFullUri: http://169.254.170.23/v1/credentials
+    authorizationTokenFile: /var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token
+    credentialLifetimeFactor: 0.8          # optional
```
robobario, line 112: might be worth a brief description of what this does, even though it's already been used for the existing credentials source

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+- `close()` to shut down the executor
+
+Subclasses implement:
+- `fetchCredentials()` — provider-specific HTTP call
+- `expirationOf(C)` — extract expiration instant
+- `onRefreshFailure(Throwable)` / `onRefreshSuccess(C)` — optional log hooks
+
+`Ec2MetadataCredentialsProvider` becomes a subclass with identical external behaviour.
+
+### Environment variable defaults
```
robobario, line 150: Maybe better to have this be a comprehensive table of all the configuration options, brief description of what they do, default values, and their env var if present.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+
+```yaml
+credentials:
+  podIdentity:
+    credentialsFullUri: http://169.254.170.23/v1/credentials
+    authorizationTokenFile: /var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token
+    credentialLifetimeFactor: 0.8          # optional
+```
+
+On a properly-associated EKS pod the agent injects `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`, so the minimal configuration is:
```
robobario, line 125: Maybe we should define somewhere that it will fail fast if these are not configured or available from the environment.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+
+```yaml
+credentials:
+  webIdentity:
+    roleArn: arn:aws:iam::123456789012:role/KroxyliciousIRSA
+    webIdentityTokenFile: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
+    roleSessionName: my-session            # optional
+    stsEndpointUrl: https://sts.us-east-1.amazonaws.com  # optional, derived from region
+    stsRegion: us-east-1                   # optional, falls back to Config.region
+    durationSeconds: 3600                  # optional
```
robobario, line 72: a description of what this does and it's limits would be useful ([900-43200 as a valid range for AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) assuming that's what we are controlling)

is there a default, or does absence imply we don't set it on the request?

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+### Environment variable defaults
+
+| Provider | Config field | Env var fallback |
+|----------|-------------|-----------------|
+| IRSA | `roleArn` | `AWS_ROLE_ARN` |
+| IRSA | `webIdentityTokenFile` | `AWS_WEB_IDENTITY_TOKEN_FILE` |
+| IRSA | `roleSessionName` | `AWS_ROLE_SESSION_NAME` |
+| IRSA | `stsRegion` | `AWS_REGION` |
+| Pod Identity | `credentialsFullUri` | `AWS_CONTAINER_CREDENTIALS_FULL_URI` |
+| Pod Identity | `authorizationTokenFile` | `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` |
```
robobario: Using environment variables is a bit pioneering. Until now the proxy has operated on explicit YAML configuration, preferring to have everything visible in the single file for validation.

However in this case it is justified to use these standard env vars for the file/uri endpoints, the SDK docs is in terms of the env vars and the URI/file locations aren't defined. As opposed to the EC2 metadata endpoint which is documented and has a stable contract.

https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html

We should add a paragraph justifying the use of environment variables.

It would also be good to limit them as much as possible since we generally favour explicit config in the YAML. I think we could drop the env var support for `stsRegion`, `roleArn`, `roleSessionName`.

robobario: @tombentley does that stir any thoughts about your Proxy config ideas #96. Having part of the Proxy configuration be determined by uncontrolled environment variables isn't great in terms of modelling when we need to reload changes parts of the config graph. But in this case it really does seem like something that should be pulled from the environment.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+
+Registered under `credentials.webIdentity` (per proposal 017):
+
+```yaml
+credentials:
+  webIdentity:
+    roleArn: arn:aws:iam::123456789012:role/KroxyliciousIRSA
+    webIdentityTokenFile: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
+    roleSessionName: my-session            # optional
+    stsEndpointUrl: https://sts.us-east-1.amazonaws.com  # optional, derived from region
```
robobario: we should define the derivation, I assume it uses `stsRegion` else the higher level `Config.region`.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,168 @@
...
+#### Configuration
+
+Registered under `credentials.webIdentity` (per proposal 017):
+
+```yaml
+credentials:
+  webIdentity:
+    roleArn: arn:aws:iam::123456789012:role/KroxyliciousIRSA
+    webIdentityTokenFile: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
+    roleSessionName: my-session            # optional
```
robobario, line 70: what is the default value? The AWS SDK would generate one including a timestamp according to claude (haven't followed that up)

robobario (commented):
Thanks @oleksiyp great to have your proposal, support for EKS auth would be brilliant to have.

Have put up some points, mostly asking for elaboration on the new config options. Also a paragraph of justification for using environment variables would be great as we have so far preferred explicit YAML configuration.

@k-wall is also expert in the AWS KMS so keen for him to take a look.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Indeed all these environment variables are added automatically by EKS. When required to be specified explicitly it starts leaking configuration of AWS into Kubernetes manifests, which results in non-clean separation between infrastructure and "platform level" configuration.

I will add a paragraph justifying that.

oleksiyp:
@robobario addressed comments you provided in https://github.com/kroxylicious/design/pull/97/changes/d711ee554cb601b72678366f19e3c265d5062c49

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,185 @@
...
+    credentialLifetimeFactor: 0.8          # optional
+```
+
+| Field | Description | Required? | Default | Env var fallback |
+|-------|-------------|-----------|---------|-----------------|
+| `roleArn` | ARN of the IAM role to assume via STS. | No | — | `AWS_ROLE_ARN` |
+| `webIdentityTokenFile` | Path to the projected service-account OIDC token file. Read fresh on every credential refresh (kubelet rotates it roughly hourly). | No | — | `AWS_WEB_IDENTITY_TOKEN_FILE` |
+| `roleSessionName` | Identifier for the assumed-role session, visible in AWS CloudTrail. Must match `[\w+=,.@-]{2,64}`. | No | `kroxylicious-<uuid>` (generated at construction time) | `AWS_ROLE_SESSION_NAME` |
+| `stsEndpointUrl` | STS endpoint URL for the `AssumeRoleWithWebIdentity` call. Override for non-standard partitions (GovCloud, China). | No | `https://sts.<stsRegion>.amazonaws.com` | — |
+| `stsRegion` | AWS region used to derive the STS endpoint URL. | No | Value of `Config.region` | `AWS_REGION` |
```
robobario: `Config.region` is required. So I guess we don't need to support the `AWS_REGION` env var?

oleksiyp: hm. as far as I know "authentication region" might be slightly different from "current EKS region".  for sure passing it through is correct implementation, while we risk to introduce rarely observable bug. let me check that

oleksiyp: ah now I see what you mean

oleksiyp: You're right - since Config.region is required in YAML and no case STS needs to be in other region than KMS let's drop then stsRegion alltogether. Always possible to use stsEndpointUrl to redefine full URL. I dropped it in https://github.com/kroxylicious/design/pull/97/commits/0558bd8d1c7d5bdfdb69c0f743183fc1426a2a39

proposals/017-aws-kms-credentials-config-restructure.md
```diff
@@ -0,0 +1,162 @@
...
+- **Deprecation:** the flat-field style will emit a log warning encouraging migration to the `credentials` node.  Removal of the deprecated fields is deferred to a future major version.
+- **Forward compatible:** adding a new credential provider requires only a new field on `CredentialsConfig` — no changes to the top-level `Config` shape.
+
+## Rejected alternatives
+
+1. **Jackson polymorphic deserialization (`@JsonTypeInfo`)** — using a type discriminator inside `credentials` (e.g. `type: webIdentity`) would enforce single-provider semantics at the deserialization layer.  Rejected because it requires a discriminator property that differs from the key-per-provider style used elsewhere in Kroxylicious (Azure KMS `entraIdentity`, Fortanix `apiKeySession`) and makes the YAML less intuitive.
+
+2. **Break existing config without backward compat** — since the project is pre-1.0, breaking changes are nominally acceptable.  Rejected because existing users have working configurations and migration should be painless.  The cost of the backward-compat shim (a few lines in the compact constructor) is negligible compared to the user friction of a forced migration.
```
k-wall: FYI we acually have a pre-1.0 deprecation policy in place https://github.com/kroxylicious/kroxylicious/blob/main/DEV_GUIDE.md#deprecation-policy so 2. wouldn't actually have been an accepted alterantive.

k-wall (approved):
LGTM - thank you @oleksiyp, both proposals look good to me.

k-wall:
@ddaviesbrackett-dwp I know you run Kroxylicious on EC2 so are closer to the issues than me. If you have bandwidth would you mind reviewing?

oleksiyp:
@robobario do you have any other requests to fix in this proposals?

robobario:
nope, looks good to me thank you! I'll put a call out on the dev list to see if anyone else wants to review it, else will merge it in 24 hours.

robobario (approved):

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+
+1. Read the projected JWT from `webIdentityTokenFile` — **fresh on every refresh**, since kubelet rotates this file roughly hourly.
+2. Build an **unsigned** `POST` to the regional STS endpoint:
+   ```
+   Action=AssumeRoleWithWebIdentity&Version=2011-06-15
+   &RoleArn=<url-encoded>&RoleSessionName=<url-encoded>
+   &WebIdentityToken=<url-encoded JWT>
+   ```
+   `AssumeRoleWithWebIdentity` is unsigned by design — the JWT is the credential.
+3. Request `Accept: application/json` so STS replies with JSON (no XML parser needed).
```
tombentley: The `Accept: application/json` behaviour is not documented in the [STS API reference](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html). The AWS SDK v2 uses it internally so it works in practice, but it's not a contractual guarantee. Worth acknowledging this risk, or having a fallback to XML parsing (it's a small, fixed schema).

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+    webIdentityTokenFile: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
+    roleSessionName: my-session            # optional
+    stsEndpointUrl: https://sts.us-east-1.amazonaws.com  # optional, derived from Config.region
+    durationSeconds: 3600                  # optional
+    credentialLifetimeFactor: 0.8          # optional
+```
+
+| Field | Description | Required? | Default | Env var fallback |
+|-------|-------------|-----------|---------|-----------------|
+| `roleArn` | ARN of the IAM role to assume via STS. | No | — | `AWS_ROLE_ARN` |
```
tombentley, line 78: "Required? No" is misleading — `roleArn` is mandatory, it just has an env-var fallback source. Same for `webIdentityTokenFile` below. Consider "Yes (env var fallback)" or similar wording so the table alone doesn't suggest these are truly optional.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+| `webIdentityTokenFile` | Path to the projected service-account OIDC token file. Read fresh on every credential refresh (kubelet rotates it roughly hourly). | No | — | `AWS_WEB_IDENTITY_TOKEN_FILE` |
+| `roleSessionName` | Identifier for the assumed-role session, visible in AWS CloudTrail. Must match `[\w+=,.@-]{2,64}`. | No | `kroxylicious-<uuid>` (generated at construction time) | `AWS_ROLE_SESSION_NAME` |
+| `stsEndpointUrl` | STS endpoint URL for the `AssumeRoleWithWebIdentity` call. Override for non-standard partitions where the endpoint pattern differs (e.g. China: `sts.<region>.amazonaws.com.cn`, ISO: `sts.<region>.c2s.ic.gov`). | No | `https://sts.<Config.region>.amazonaws.com` | — |
+| `durationSeconds` | Requested duration of the assumed-role session, in seconds. Valid range: [900, 43200](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html). When absent the field is omitted from the STS request and STS applies the role's configured maximum session duration. | No | Omitted (STS default) | — |
```
tombentley: The [900, 43200] range is the API parameter constraint, but the *effective* maximum is capped by the IAM role's `MaxSessionDuration` setting (default 1 hour). If a user specifies 43200 but the role allows only 3600, STS rejects the request. A parenthetical note here would save users some debugging.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+
+The implementation is AWS-SDK-free: all HTTP calls use `java.net.http.HttpClient` with a custom SigV4 implementation.
+
+Neither existing mechanism works for pods on **Amazon EKS**: long-term keys are a security anti-pattern, and EC2 IMDS is only reachable from EC2 instances.
+
+## Motivation
+
+Amazon EKS is the primary deployment target for Kroxylicious on AWS.  AWS supports two pod-level authentication mechanisms:
+
+1. **IAM Roles for Service Accounts (IRSA)** — a projected OIDC token is exchanged for temporary credentials via STS `AssumeRoleWithWebIdentity`.  This requires an OIDC trust-policy on the IAM role and an annotated Kubernetes service account.
```
tombentley: Consider adding a brief note (here or in the docs section) on what IAM permissions the assumed role needs. EKS users may create a dedicated role for this, and a hint that it should be scoped to the minimum KMS actions (e.g. `kms:Decrypt`, `kms:Encrypt`, `kms:GenerateDataKey*`, `kms:DescribeKey`) would help them avoid overly broad policies.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+4. Validate the URI scheme (`http`/`https` only) at construction time.
+
+#### Configuration
+
+Registered under `credentials.podIdentity` (per proposal 017):
+
+```yaml
+credentials:
+  podIdentity:
+    credentialsFullUri: http://169.254.170.23/v1/credentials
```
tombentley, line 110: The Pod Identity Agent uses plain HTTP here (link-local `169.254.170.23`, same-node only). This is by design, but worth a sentence in the proposal acknowledging the trust boundary — any process that can sniff or spoof traffic in the node's network namespace can intercept these credentials.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+| `stsEndpointUrl` | STS endpoint URL for the `AssumeRoleWithWebIdentity` call. Override for non-standard partitions where the endpoint pattern differs (e.g. China: `sts.<region>.amazonaws.com.cn`, ISO: `sts.<region>.c2s.ic.gov`). | No | `https://sts.<Config.region>.amazonaws.com` | — |
+| `durationSeconds` | Requested duration of the assumed-role session, in seconds. Valid range: [900, 43200](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html). When absent the field is omitted from the STS request and STS applies the role's configured maximum session duration. | No | Omitted (STS default) | — |
+| `credentialLifetimeFactor` | Controls preemptive refresh: the credential is refreshed in the background once it reaches this fraction of its total lifetime. For example, 0.8 means the credential is refreshed at 80% of its lifetime. This behaviour is shared with the existing EC2 metadata provider via the `AbstractRefreshingCredentialsProvider` base class. | No | `0.8` | — |
```
tombentley: The valid range for this field isn't specified. Values ≤ 0 or ≥ 1 would cause pathological refresh behaviour (never refresh, or refresh only after expiry). Worth stating the constraint explicitly, e.g. "must be in (0, 1)".

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+
+The provider **fails fast** at construction time with a `KmsException` if `roleArn` and `webIdentityTokenFile` cannot be resolved from either YAML configuration or their respective environment variables.
```
tombentley: The fail-fast behaviour for missing config is good. Consider extending it to also validate `roleSessionName` against the documented regex and `durationSeconds` against [900, 43200] at construction time, rather than deferring to STS error responses. Client-side validation gives clearer error messages.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+  * [Affected/not affected projects](#affectednot-affected-projects)
+  * [Compatibility](#compatibility)
+  * [Rejected alternatives](#rejected-alternatives)
+<!-- TOC -->
+
+## Current situation
+
+The AWS KMS provider for the Record Encryption filter supports two authentication mechanisms: long-term IAM credentials and EC2 instance metadata.  Per proposal 017, these are being restructured under a `credentials` configuration node.
+
+The existing `Credentials` interface already supports an optional `securityToken()` for temporary credentials, and `AwsV4SigningHttpRequestBuilder` adds the `X-Amz-Security-Token` header when present — so the signing layer is already ready for temporary credentials from STS.
```
tombentley, line 25: Pre-existing observation, not introduced by this proposal: the current `AwsV4SigningHttpRequestBuilder` adds `X-Amz-Security-Token` *after* computing the signature (line 241), so it's not a signed header. The [AWS docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html) say this varies by service — "Some services require that you add `X-Amz-Security-Token` to the canonical request. Other services require only that you add `X-Amz-Security-Token` at the end, after you calculate the signature." Since EC2 metadata credentials already use this path, it presumably works for KMS, but there are no tests covering the security token header. Worth verifying this is intentional and adding test coverage before the new providers depend on it.

proposals/018-aws-kms-eks-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+On a properly-associated EKS pod the agent injects `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`, so the minimal configuration is:
+
+```yaml
+credentials:
+  podIdentity: {}
+```
+
+### Shared refresh infrastructure
+
+The async refresh / expiry / exponential-backoff state machine embedded in `Ec2MetadataCredentialsProvider` is extracted into a package-private `AbstractRefreshingCredentialsProvider<C extends Credentials>`.
```
tombentley, line 134: Two questions about the shared refresh infrastructure:

1. **Startup behaviour**: The proposal says fail-fast for missing *config*, but what about network unreachability on first credential fetch? Is the first fetch lazy (async, retried via backoff) or eager (blocks startup)? This affects pod readiness semantics.

2. **Credential sharing**: If multiple filter instances use the same IRSA/Pod Identity config, do they each get their own credential provider instance? STS has per-account rate limits; independent fetches could matter at scale.

tombentley (commented):
Thanks for this! It basically LGTM.

Claude found a few minor points which we can improve. They're mostly just clarifying the proposal itself, not changing what's being proposed.

Thanks again.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Acknowledged in the proposal. Keeping JSON-only — used by AWS SDK v2 and avoids an XML parser dependency. If AWS breaks it we'll add XML parsing then.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: "No" is from the YAML perspective — the user doesn't have to specify it if the env var is set. The fail-fast paragraph below the table clarifies that the value must exist from one source or the other.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Good point — added a note about the role's MaxSessionDuration capping the effective maximum, and that we validate [900, 43200] at construction time.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Added a note. The full IAM setup (alias-based policy with the specific KMS actions) is covered in the docs procedures.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Added a note — same trust model as EC2 IMDS. The projected SA token authenticates each pod to the agent.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Added explicit (0, 1) range constraint.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Agreed — added that roleSessionName and durationSeconds are validated at construction time.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Pre-existing behaviour proven by the EC2 metadata provider. Adding test coverage is worthwhile but out of scope for this proposal.

proposals/018-aws-kms-eks-authentication.md
oleksiyp: Both are existing framework-level behaviours, not introduced by this proposal. First fetch is lazy/async — `getCredentials()` returns a `CompletionStage`, nothing blocks startup. Credential sharing: one `CredentialsProvider` per `AwsKmsService.initialize()`, shared across all `buildKms()` instances. Same pattern as Fortanix DSM's `SessionProvider`.

oleksiyp:
@tombentley checked comments and fixed what makes sense to me in  https://github.com/kroxylicious/design/pull/97/commits/83ee98d5335fda63f7d282fa38db4246a2f8f598

tombentley (approved):
Many thanks @oleksiyp

oleksiyp:
@robobario can you please merge?

robobario:
Done, thanks for that!

PR #98, opened 2026-04-13
feat: KafkaProxy threading and lifecycle contract

SamBarker:
## Summary

Define the threading and lifecycle contract for `KafkaProxy`: single-use, idempotent `startup()`/`shutdown()`, future-returning `startup()` API, deprecation of `block()`.

This is a prerequisite for the virtual cluster lifecycle work in [016](./proposals/016-virtual-cluster-lifecycle.md) — defining the outer proxy contract first so that inner components can make informed concurrency decisions.

## Key decisions

- `KafkaProxy` is single-use: `NEW` → `STARTING` → `STARTED` → `STOPPING` → `STOPPED`
- `startup()` returns a `CompletableFuture<Void>` that completes when the proxy stops — replaces `block()`
- Both `startup()` and `shutdown()` are idempotent — callers do not need to coordinate
- `shutdown()` may be called from any thread (JVM shutdown hook)
- The only error is `startup()` after `STOPPING`/`STOPPED` (not restartable)
- No `FAILED` state at proxy level — failure propagates via exception, proxy is not reconfigurable

## Test plan

- [ ] Review proposal for completeness and consistency
- [ ] Confirm alignment with virtual cluster lifecycle proposal (016)

### DIFF
--- a/proposals/019-kafkaproxy-threading-and-lifecycle.md
+++ b/proposals/019-kafkaproxy-threading-and-lifecycle.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Define the threading and lifecycle contract for `KafkaProxy`: it is a single-use object with a linear lifecycle (`NEW` -> `STARTING` -> `STARTED` -> `STOPPING` -> `STOPPED`). Both `startup()` and `shutdown()` are idempotent — they may be called multiple times, but only one call performs the work.
+Define the threading and lifecycle contract for `KafkaProxy`: it is a single-use object with a linear lifecycle (`NEW` -> `STARTING` -> `STARTED` -> `STOPPING` -> `STOPPED`). `startup()` returns a single `CompletableFuture<Void>` representing the full proxy lifetime — it completes only when the proxy reaches `STOPPED`. Both `startup()` and `shutdown()` are idempotent — they may be called multiple times, but only one call performs the work.
 
 ## Current Situation
 
@@ -10,7 +10,7 @@ Define the threading and lifecycle contract for `KafkaProxy`: it is a single-use
 
 - There is no documentation of which threads may call which methods.
 - The two-state boolean cannot distinguish "not yet started" from "stopped," so `startup()` after `shutdown()` silently succeeds in setting `running` back to `true` rather than being rejected.
-- The `block()` method checks `running.get()` but does not define what happens if called before `startup()` or after `shutdown()`. More fundamentally, `block()` exists as a separate method only because `startup()` does not return a future — the caller has no other way to wait for shutdown.
+- The `block()` method checks `running.get()` but does not define what happens if called before `startup()` or after `shutdown()`. More fundamentally, `block()` exists as a separate method only because `startup()` returns `this` rather than a future — the caller has no other way to wait for shutdown. A `shutdown` `CompletableFuture` already exists internally and is what `block()` joins, but it is not surfaced to callers.
 - The `close()` method (from `AutoCloseable`) conditionally calls `shutdown()` based on `running.get()`, but the interaction between `close()` and explicit `shutdown()` calls is undocumented.
 
 The lack of a formal contract means that components within the proxy — such as the virtual cluster lifecycle state machine ([016](./016-virtual-cluster-lifecycle.md)) — cannot make informed decisions about their own concurrency models.
@@ -43,27 +43,37 @@ NEW  ──>  STARTING  ──>  STARTED  ──>  STOPPING  ──>  STOPPED
 - **STOPPING**: `shutdown()` is in progress. The caller may join the future returned by `startup()` to wait for completion.
 - **STOPPED**: terminal. All resources released.
 
-If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before returning a failed future. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the exception from `startup()`, so there is no need to hold the error in state for later inspection.
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before returning a failed future. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the future's exceptional completion, so there is no need to hold the error in state for later inspection.
 
 The lifecycle is **not restartable**. Once a proxy reaches `STOPPED`, it cannot return to any earlier state. To restart, create a new `KafkaProxy` instance. This holds across all deployment models: standalone and Kubernetes today, and sidecar or embedded library if those models are supported in the future.
 
-Dynamic configuration reload handles changes within a running proxy; there is no use case for tearing down and re-creating Netty event loops in-process.
+Dynamic configuration reload will handle changes within a running proxy; there is no use case for tearing down and re-creating a proxy instance.
 
-### `startup()` Returns a Future
+### The Lifecycle Future
 
-`startup()` returns a `CompletableFuture<Void>` that completes when the proxy reaches `STOPPED`. This lets the caller decide how to wait:
+`startup()` returns a `CompletableFuture<Void>` that represents the proxy's **full lifetime** — it is the same future on every call, and completes only when the proxy reaches `STOPPED`. The caller always discovers the final outcome by joining the future received from `startup()`:
 
 ```java
 // Standalone — block the main thread until shutdown
 proxy.startup().join();
 
 // Embedded / tests — start and hold the future
-var stopped = proxy.startup();
+CompletableFuture<Void> stopped = proxy.startup();
 // ... do other work ...
 stopped.join();
 ```
 
-If `startup()` fails, the returned future completes exceptionally — the caller receives the failure whether they call `join()` immediately or later. The proxy also holds the future internally. The existing `block()` method is deprecated and delegates to `join()` on the same future.
+The future has three terminal outcomes:
+
+| Path | Completion |
+|---|---|
+| Startup failure (`STARTING → STOPPED`) | Exceptionally — the startup error |
+| Clean shutdown (`STOPPING → STOPPED`) | Normally |
+| Shutdown failure (`STOPPING → STOPPED`) | Exceptionally — the shutdown error |
+
+Calling `cancel()` on the future is equivalent to calling `shutdown()` — it initiates a graceful shutdown. The `mayInterruptIfRunning` parameter is ignored; shutdown is always graceful regardless of its value.
+
+The existing `block()` method is deprecated and delegates to `join()` on the lifecycle future.
 
 ### Threading Contract
 
@@ -79,7 +89,7 @@ Both `startup()` and `shutdown()` are idempotent — callers do not need to coor
 
 1. **At most one startup and one shutdown sequence will execute.** Multiple concurrent calls to `startup()` all receive the same future; only one thread performs initialisation. Multiple concurrent calls to `shutdown()` are similarly safe; only one thread performs the shutdown sequence.
 2. **`shutdown()` may be called from a different thread than `startup()`.** This is the common case: `startup()` on the application thread, `shutdown()` from a JVM shutdown hook.
-3. **The future returned by `startup()` is safe to join concurrently with `shutdown()`.** The future is completed at the end of the shutdown sequence, regardless of whether shutdown succeeds or fails.
+3. **The caller joins the same future for both startup and shutdown outcomes.** It is safe to join the future concurrently with `shutdown()` — `shutdown()` is what completes it, at the end of the shutdown sequence.
 
 ### Error Behaviour
 
@@ -93,16 +103,17 @@ All other repeated or concurrent calls are idempotent no-ops (or return the exis
 
 ### Process Exit Codes
 
-There are two edges into `STOPPED`, and they carry different meaning at the process level:
+There are three paths into `STOPPED`, each reflected in the lifecycle future:
 
-1. **`STARTING → STOPPED`** (startup failure): `startup()` completes the future exceptionally. The caller (typically `main()`) receives the exception and should exit with a non-zero status code.
-2. **`STOPPING → STOPPED`** (clean shutdown): the future completes normally. The caller exits with zero.
+1. **Startup failure** (`STARTING → STOPPED`): the future completes exceptionally. The caller (typically `main()`) receives the exception and should exit with a non-zero status code.
+2. **Clean shutdown** (`STOPPING → STOPPED`): the future completes normally. The caller exits with zero.
+3. **Shutdown failure** (`STOPPING → STOPPED`): the future completes exceptionally with the shutdown error. The caller should exit with a non-zero status code.
 
 `KafkaProxy` itself does not call `System.exit()` — translating the outcome into a process exit code is the responsibility of the application entry point. This keeps `KafkaProxy` usable in embedded and test contexts where process termination is not appropriate.
 
 ### Metrics
 
-While the lifecycle state enum remains internal (see [Rejected Alternatives](#public-lifecycle-state-enum)), the proxy should expose metrics that reflect its lifecycle state. Standard Kubernetes metrics (`kube_pod_status_phase`, restart counts, container termination reasons) treat the process as a black box — they can tell an operator *that* a pod restarted but not *why*, and they cannot distinguish a startup failure from a crash during normal operation without heuristics.
+While the lifecycle state enum remains internal (see [Rejected Alternatives](#public-lifecycle-state-enum)), the proxy should expose metrics that reflect its lifecycle state. Platform-level observability (restart counts, container termination reasons) treats the process as a black box — it can tell an operator *that* a process restarted but not *why*, and cannot distinguish a startup failure from a crash during normal operation without heuristics.
 
 The proxy should expose:
 
@@ -126,6 +137,8 @@ These metrics are public API surface — once exposed, their names and semantics
 
 This proposal formalises existing behaviour rather than changing it. The proxy is already single-use in practice — no code path calls `startup()` twice or restarts after `shutdown()`. The change makes the contract explicit and enforces it.
 
+The API change is that `startup()` changes its return type from `KafkaProxy` to `CompletableFuture<Void>`. Since `KafkaProxy` is an internal class, there are no external callers to migrate.
+
 The one behavioural change is stricter error detection: calling `startup()` on a stopped proxy currently appears to succeed but would leave the proxy in a corrupt state. After this change, it throws `IllegalStateException`. This is a bug fix, not a compatibility break.
 
 ## Rejected Alternatives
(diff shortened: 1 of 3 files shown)


proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,119 @@
+# Proposal 019: KafkaProxy Threading and Lifecycle Contract
+
+## Summary
+
+Define the threading and lifecycle contract for `KafkaProxy`: it is a single-use object with a linear lifecycle (`NEW` -> `STARTING` -> `STARTED` -> `STOPPING` -> `STOPPED`). Both `startup()` and `shutdown()` are idempotent — they may be called multiple times, but only one call performs the work.
+
+## Current Situation
+
```
tombentley, line 8: I think it's worth calling out that `KafkaProxy` is **not** public API.

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,119 @@
...
+- **STARTED**: the proxy is serving traffic.
+- **STOPPING**: `shutdown()` is in progress. The caller may join the future returned by `startup()` to wait for completion.
+- **STOPPED**: terminal. All resources released.
+
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before the exception propagates. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the exception from `startup()`, so there is no need to hold the error in state for later inspection.
+
+The lifecycle is **not restartable**. Once a proxy reaches `STOPPED`, it cannot return to any earlier state. To restart, create a new `KafkaProxy` instance. This holds across all deployment models: standalone, Kubernetes, sidecar, and embedded library.
```
tombentley: We should be clear that not all of these (sidecar and embedded library) are supported deployment models today. But the intent is clear: If/when they are they won't change the contract we're defining here.

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,119 @@
...
+                └──────────  (startup failure)  ─────────┘
+```
+
+- **NEW**: constructed, not yet started.
+- **STARTING**: `startup()` is in progress.
+- **STARTED**: the proxy is serving traffic.
+- **STOPPING**: `shutdown()` is in progress. The caller may join the future returned by `startup()` to wait for completion.
+- **STOPPED**: terminal. All resources released.
+
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before the exception propagates. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the exception from `startup()`, so there is no need to hold the error in state for later inspection.
```
tombentley: If there is no failed state how are the following cases distinguished:

* Proxy tries to start up, finds a problem with its configuration, comes to a stop.
* Proxy starts up, succeeds, but is later shutdown intentionally via a signal?

I think the difference matters because the first represents an error, and we'd want the process status code to return non-zero. The second is not an error, and we'd want the process status code to return zero.

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,119 @@
...
+```java
+// Standalone — block the main thread until shutdown
+proxy.startup().join();
+
+// Embedded / tests — start and hold the future
+var stopped = proxy.startup();
+// ... do other work ...
+stopped.join();
+```
+
```
tombentley, line 65: You've not said anything about the exceptional completion of the future. In particular, if the proxy cannot start up will the future complete with an exception, or will it complete normally (because according to your assertion that there is no `FAILED` state we have reached the `STOPPED` state)?

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,119 @@
...
+## Proposal
+
+### KafkaProxy is Single-Use
+
+A `KafkaProxy` instance follows a linear, non-repeatable lifecycle:
+
+```
+NEW  ──>  STARTING  ──>  STARTED  ──>  STOPPING  ──>  STOPPED
+                │                                        ▲
+                └──────────  (startup failure)  ─────────┘
```
tombentley, line 37: This explains the conceptual model, but do you anticipate whether these states will (eventually) be exposed via a public API. Metrics or the mgmt server, for example? Should this proposal define how they relate to the health endpoints?

tombentley (commented):
Thanks @SamBarker

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+The lack of a formal contract means that components within the proxy — such as the virtual cluster lifecycle state machine ([016](./016-virtual-cluster-lifecycle.md)) — cannot make informed decisions about their own concurrency models.
+
+## Motivation
+
+The proxy's lifecycle is simple by design: start once, run, stop once. The threading context is similarly simple but not obvious — `shutdown()` is called from a JVM shutdown hook, which runs on a dedicated thread. The virtual cluster lifecycle ([016](./016-virtual-cluster-lifecycle.md)) introduces per-cluster state machines that execute within this proxy-level lifecycle. Without a documented proxy-level contract, each component must independently guess at the concurrency model.
+
+Formalising the contract provides:
+
+- **Correctness**: the only genuinely invalid sequence (start-after-stop) is detected and rejected. All other repeated or concurrent calls are idempotent.
+- **Guidance for implementors**: components within the proxy can choose appropriate synchronisation strategies based on documented guarantees rather than assumptions.
```
tombentley: Not sure what "implementors" means here. Implementors of _what_? You're talking about other proxy-internal things like the VC Manager stuff, so I think what you're getting at is that with this proxy-level lifecycle we can better define/have guarantees around the lifecycles of those other things.

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+                └──────────  (startup failure)  ─────────┘
+```
+
+- **NEW**: constructed, not yet started.
+- **STARTING**: `startup()` is in progress.
+- **STARTED**: the proxy is serving traffic.
+- **STOPPING**: `shutdown()` is in progress. The caller may join the future returned by `startup()` to wait for completion.
+- **STOPPED**: terminal. All resources released.
+
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before the exception propagates. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the exception from `startup()`, so there is no need to hold the error in state for later inspection.
```
tombentley: > partially-acquired resources are released before the exception propagates

The way I've always used "propagates" when talking about exceptions is synonymous with Java's  `throw`/`catch`. When dealing with futures I think it's better to use different terminology, and talk about "exceptional completion". Using separate terminology allows us to talk precisely about a method which returns a future, where the method might throw (and thus an exception propagate), or return a future which completes exceptionally (no exception propagation, but an obligation to handle the exception at some point). 

So I wonder if we can reword this to talk about "releasing resources before returning a failed future". I think that would have avoid confusing readers.

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before the exception propagates. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the exception from `startup()`, so there is no need to hold the error in state for later inspection.
+
+The lifecycle is **not restartable**. Once a proxy reaches `STOPPED`, it cannot return to any earlier state. To restart, create a new `KafkaProxy` instance. This holds across all deployment models: standalone and Kubernetes today, and sidecar or embedded library if those models are supported in the future.
+
+Dynamic configuration reload handles changes within a running proxy; there is no use case for tearing down and re-creating Netty event loops in-process.
+
+### `startup()` Returns a Future
+
```
tombentley, line 53: I think this section would be clearer if you at least started by laying out the signatures of the relevant lifecycle methods of `KafkaProxy`: `startup()`, `shutdown()` and `block()`. I think it might even just be simpler to show the javadoc for the contract you're proposing.

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+// Standalone — block the main thread until shutdown
+proxy.startup().join();
+
+// Embedded / tests — start and hold the future
+var stopped = proxy.startup();
+// ... do other work ...
+stopped.join();
+```
+
+If `startup()` fails, the returned future completes exceptionally — the caller receives the failure whether they call `join()` immediately or later. The proxy also holds the future internally. The existing `block()` method is deprecated and delegates to `join()` on the same future.
```
tombentley: The importance of the "The proxy also holds the future internally." is not clear to me.

```suggestion
If `startup()` succeeds then:
* the returned future completes normally
* the caller can assume the shutdown hook has been installed (there's no need for the caller to register their own shutdown hook to perform shutdown)
* beyond JVM shutdown, the caller is obliged to arrange calling `shutdown()` as appropriate for them.

If `startup()` fails:
* the returned future completes exceptionally — the caller receives the failure whether they call `join()` immediately or later. 
* There is no obligation on call `shutdown()` (resources are already cleaned up), but if `shutdown()` _is_ called it is a no-op.
* The proxy also holds the future internally. 

When `shutdown()` is called:
* it will block until shutdown has finished. 
* it will clean up resources all resources (i.e. it won't give up on the first exception during the shutdown process).
* if there are any problems while shutting down `shutdown()` will throw.

The existing `block()` method is deprecated and delegates to `join()` on the same future.
```

My only other question is whether we want `shutdown()` to support a timeout parameter... Eventually we will probably improve the clean shutdown problem (wrt idempotent producers), and the solution will likely involve waiting for the client to discover that all the (virtual) brokers that it's connected to are no longer leaders or coordinators. i.e. our shutdown times will come to depend on the client libraries own state (where today they "just" depend on things like closing sockets). If this comes to pass shutting down all the clients will take a while, but we will want a bound on just how long it can take. We already have some configs for the timeout for netty shutdown. I guess my question is: Is the `Configuration` the right place for that? Or is it better as a parameter to `shutdown()`?

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+| `startup()` | `STOPPING` or `STOPPED` | `IllegalStateException` — the proxy is not restartable |
+
+All other repeated or concurrent calls are idempotent no-ops (or return the existing future).
+
+### Process Exit Codes
+
+There are two edges into `STOPPED`, and they carry different meaning at the process level:
+
+1. **`STARTING → STOPPED`** (startup failure): `startup()` completes the future exceptionally. The caller (typically `main()`) receives the exception and should exit with a non-zero status code.
+2. **`STOPPING → STOPPED`** (clean shutdown): the future completes normally. The caller exits with zero.
```
tombentley: This needs to be explained more clearly. Yes, the startup future completes normally in this case, but (I'm assuming) shutdown itself can fail.

It might be as simple to work this into the previous sections where (if you accept my suggestions) we will have mostly laid out the different combinations of exceptions. You'd only need to add wording about the states, and even then maybe only if these state machine is more concrete than a conceptual model.

tombentley (commented):
Thanks Sam! I think the proposal is becoming clearer to me now, and it makes sense. I've made a few suggestions for how it's being explained.

proposals/019-kafkaproxy-threading-and-lifecycle.md
SamBarker: Yeah I can see the tension, I'll re-word.

proposals/019-kafkaproxy-threading-and-lifecycle.md
SamBarker: > The importance of the "The proxy also holds the future internally." is not clear to me.

Its a bit clumsy, but what I'm trying to get at is that future returned from start is state that KafkaProxy is required to track to make everything else work. 
- 2nd call to start just returns the same future that already exists, it doesn't attempt to start a new startup sequence.  
- We can preserve `block()` for backwards compatibility reason, by having it call `future.join()` as its implementation. 

> My only other question is whether we want shutdown() to support a timeout parameter...

For now, I think configuration is still the right home. I'm not convinced the config we have now which is Netty shutdown timeouts is really the same thing your describing. My current thinking is what we have in config is, and will remain, the hard backstop. I think there is a tension between the graceful protocol aware teardown that you are outlining and the pod timeouts orchestrators impose on pods. 

I can see that we might want to pass some protocol level timeouts through the shutdown call tree at some point in the future. However I don't think this would be the right point in call stack to inject that. `KakfaProxy` itself is the owner of the config tree so callers would have to 1. parse the config themselves or 2. ask `KakfaProxy` what timeout should I give you? At which point KafkaProxy might as well look it up itself...

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before the exception propagates. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the exception from `startup()`, so there is no need to hold the error in state for later inspection.
+
+The lifecycle is **not restartable**. Once a proxy reaches `STOPPED`, it cannot return to any earlier state. To restart, create a new `KafkaProxy` instance. This holds across all deployment models: standalone and Kubernetes today, and sidecar or embedded library if those models are supported in the future.
+
+Dynamic configuration reload handles changes within a running proxy; there is no use case for tearing down and re-creating Netty event loops in-process.
```
k-wall: This sentence is quite low level. It starts talking about Netty and event loops.   You need to know that the netty event loops belong to the proxy (and not the VCs).  You also need to know by dynamic configuration refers to the virtual-cluster.

I actually wonder if the sentence could just be removed.

proposals/019-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+There are two edges into `STOPPED`, and they carry different meaning at the process level:
+
+1. **`STARTING → STOPPED`** (startup failure): `startup()` completes the future exceptionally. The caller (typically `main()`) receives the exception and should exit with a non-zero status code.
+2. **`STOPPING → STOPPED`** (clean shutdown): the future completes normally. The caller exits with zero.
+
+`KafkaProxy` itself does not call `System.exit()` — translating the outcome into a process exit code is the responsibility of the application entry point. This keeps `KafkaProxy` usable in embedded and test contexts where process termination is not appropriate.
+
+### Metrics
+
+While the lifecycle state enum remains internal (see [Rejected Alternatives](#public-lifecycle-state-enum)), the proxy should expose metrics that reflect its lifecycle state. Standard Kubernetes metrics (`kube_pod_status_phase`, restart counts, container termination reasons) treat the process as a black box — they can tell an operator *that* a pod restarted but not *why*, and they cannot distinguish a startup failure from a crash during normal operation without heuristics.
```
k-wall: Why do we need to mention `kube_pod_status_phase`?   The proxy might be running in a pod, it might not.

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,139 @@
...
+- **State gauge**: a metric indicating the current lifecycle state (`starting`, `started`, `stopping`, `stopped`). This gives dashboards and alerts a definitive signal rather than requiring inference from Kubernetes-level indicators.
+- **Startup duration**: time elapsed from `STARTING` to `STARTED` (or failure). This helps operators identify configuration or environment issues that slow startup across restarts.
+
+These metrics are public API surface — once exposed, their names and semantics become a compatibility commitment. The specific metric names and labelling conventions should be consistent with any metrics framework adopted for virtual cluster lifecycle ([016](./016-virtual-cluster-lifecycle.md)).
+
+## Affected/Not Affected Projects
+
+**Affected:**
+- **kroxylicious-proxy (runtime)**: `KafkaProxy` gains a formal lifecycle contract. `startup()` returns a `CompletableFuture<Void>`. `block()` is deprecated.
```
k-wall: as block is an internal API, we could just kill it.

k-wall (approved):
LGTM - thanks for formalizing this.

k-wall:

```bash
git mv proposals/019-kafkaproxy-threading-and-lifecycle.md proposals/098-threading-and-lifecycle.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 98 - /; t; s/^# /# 98 - /}' proposals/098-threading-and-lifecycle.md && rm proposals/098-threading-and-lifecycle.md.bak
git add proposals/098-threading-and-lifecycle.md
git push
```

k-wall:

```bash
git mv proposals/019-kafkaproxy-threading-and-lifecycle.md proposals/098-kafkaproxy-threading-and-lifecycle.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 98 - /; t; s/^# /# 98 - /}' proposals/098-kafkaproxy-threading-and-lifecycle.md && rm proposals/098-kafkaproxy-threading-and-lifecycle.md.bak
git add proposals/098-kafkaproxy-threading-and-lifecycle.md
git push
```

Sorry for the confusion!

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,160 @@
+# 98 - Proposal 019: KafkaProxy Threading and Lifecycle Contract
```
k-wall: FIx the proposal number in the title.

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,160 @@
...
+                └──────────  (startup failure)  ─────────┘
+```
+
+- **NEW**: constructed, not yet started.
+- **STARTING**: `startup()` is in progress.
+- **STARTED**: the proxy is serving traffic.
+- **STOPPING**: `shutdown()` is in progress. The caller may join the future returned by `startup()` to wait for completion.
+- **STOPPED**: terminal. All resources released.
+
+If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before returning a failed future. There is no `FAILED` state at the proxy level; a failed startup is simply a proxy that went straight to `STOPPED`. This is intentional for two reasons: the proxy is not reconfigurable, so there is no recovery transition from a hypothetical `FAILED` state (unlike virtual clusters, which support `failed` -> `initializing` for retry with corrected configuration); and the failure is communicated directly to the caller via the future's exceptional completion, so there is no need to hold the error in state for later inspection.
```
k-wall: > If `startup()` fails partway through, the proxy transitions directly to `STOPPED` — partially-acquired resources are released before returning a failed future. 

The startup future will have already been returned to the caller.  I think you might mean something like: "partially-acquired resources are released before the startup future is failed exceptionally. "

k-wall (approved):
Couple of nits but other LGTM.
Thanks for hardening this contract up.

proposals/098-kafkaproxy-threading-and-lifecycle.md
SamBarker: The API surface is captured in the Threading Contract table and the Lifecycle Future section — method names, return types, and behaviour for each call sequence. Javadoc will be the definitive specification in the implementation; keeping it out of the proposal avoids duplicating it and lets the proposal stay at the contract level.

proposals/098-kafkaproxy-threading-and-lifecycle.md
```diff
@@ -0,0 +1,160 @@
+# 98 - KafkaProxy Threading and Lifecycle Contract
+
+## Summary
+
+Define the threading and lifecycle contract for `KafkaProxy`: it is a single-use object with a linear lifecycle (`NEW` -> `STARTING` -> `STARTED` -> `STOPPING` -> `STOPPED`). `startup()` returns a single `CompletableFuture<Void>` representing the full proxy lifetime — it completes only when the proxy reaches `STOPPED`. Both `startup()` and `shutdown()` are idempotent — they may be called multiple times, but only one call performs the work.
```
tombentley:
```suggestion
Define the threading and lifecycle contract for `KafkaProxy`: it is a single-use object with a linear lifecycle (`NEW` -> `STARTING` -> `STARTED` -> `STOPPING` -> `STOPPED`). `startup()` will always return the same `CompletableFuture<Void>` representing the full proxy lifetime — it completes only when the proxy reaches `STOPPED`. Both `startup()` and `shutdown()` are idempotent — they may be called multiple times, but only one call performs the work.
```

tombentley (approved):
Thanks @SamBarker

PR #99, opened 2026-04-13
feat: Micrometer configuration in the KafkaProxy CRD

SamBarker:
## Summary

- Proposes adding a structured `micrometer` field to the `KafkaProxy` spec to close the configuration gap between operator-managed and standalone proxy deployments
- Covers `standardBinders` (JVM/process metrics) and `commonTags` (static deployment-wide metric tags) with CRD-level schema validation
- Documents the static-only limitation of `commonTags` (per-pod values like AZ require future runtime enhancement)

## Test plan

- [ ] Review proposal for completeness against runtime's `MicrometerDefinition` model
- [ ] Verify the structured schema covers all `StandardBindersHook` binder names
- [ ] Confirm rejected alternatives adequately justify design decisions

### DIFF
--- a/proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
+++ b/proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
@@ -1,0 +1,224 @@
+# Micrometer Configuration in the KafkaProxy CRD
+
+Expose the proxy's existing Micrometer configuration through the `KafkaProxy` CRD so that
+operator-managed proxies have parity with standalone proxy deployments for metrics
+instrumentation.
+
+## Current situation
+
+The Kroxylicious runtime supports two Micrometer configuration hooks via the `micrometer`
+section in its YAML configuration:
+
+**`StandardBindersHook`** — registers Micrometer binders that expose JVM and process-level
+metrics:
+
+```yaml
+micrometer:
+  - type: StandardBindersHook
+    config:
+      binderNames:
+        - ProcessorMetrics
+        - JvmMemoryMetrics
+        - JvmGcMetrics
+```
+
+**`CommonTagsHook`** — adds common tags to all metrics emitted by the proxy, useful for
+distinguishing metrics from multiple proxy instances or environments:
+
+```yaml
+micrometer:
+  - type: CommonTagsHook
+    config:
+      commonTags:
+        zone: "euc-1a"
+        environment: "production"
+```
+
+However, the operator's `KafkaProxyReconciler` hardcodes the micrometer list to `List.of()`,
+and the `KafkaProxy` CRD has no field to configure either hook. This means operator-managed
+proxies cannot emit JVM-level metrics or apply common metric tags.
+
+## Motivation
+
+Without micrometer binders, the Prometheus `/metrics` endpoint only exposes application-level
+metrics (filter timings, connection counts). Resource utilisation metrics are absent:
+
+- `process_cpu_seconds_total` / `system_cpu_usage` (from `ProcessorMetrics`)
+- `jvm_memory_used_bytes` / `jvm_memory_max_bytes` (from `JvmMemoryMetrics`)
+- `jvm_gc_pause_seconds` (from `JvmGcMetrics`)
+
+These are essential for:
+
+1. **Capacity planning and sizing** — understanding proxy resource consumption under load.
+2. **Operational monitoring** — alerting on JVM health (memory pressure, excessive GC).
+3. **Benchmarking** — the OMB benchmarking infrastructure polls the Prometheus endpoint during
+   runs; without these metrics the sizing guide cannot be produced.
+
+The `micrometer` field is the last remaining proxy configuration concept without CRD
+representation.
+
+## Proposal
+
+### CRD schema
+
+Add an optional `micrometer` field to the `KafkaProxy` spec with structured sub-fields for
+each known hook type, rather than the freeform `type` + `config` pattern used in the runtime
+YAML:
+
+```yaml
+apiVersion: kroxylicious.io/v1alpha1
+kind: KafkaProxy
+metadata:
+  name: benchmark-proxy
+spec:
+  micrometer:
+    standardBinders:
+      - ProcessorMetrics
+      - JvmMemoryMetrics
+      - JvmGcMetrics
+    commonTags:
+      zone: "euc-1a"
+      environment: "production"
+```
+
+The `micrometer` field is an optional object. When omitted, no hooks are configured (current
+behaviour is preserved).
+
+| Field              | Type              | Required | Description                                                        |
+|--------------------|-------------------|----------|--------------------------------------------------------------------|
+| `standardBinders`  | array of strings  | No       | List of Micrometer binder names to register (see table below)      |
+| `commonTags`       | map<string,string> | No      | Key-value pairs added as tags to all metrics emitted by the proxy  |
+
+#### `commonTags` limitation: static values only
+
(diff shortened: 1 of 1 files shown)


proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### `commonTags` limitation: static values only
+
+`CommonTagsHook` currently accepts only static string values. This means `commonTags` is
+useful for deployment-wide tags (e.g. `environment: "production"`, `team: "platform"`) but
+cannot express per-pod values such as availability zone — all replicas would receive the
+same tag value.
+
+A future enhancement to the runtime could support environment variable interpolation in tag
+values (e.g. `zone: "${AZ}"` or `zone: "env.AZ"`), allowing the operator to inject per-pod
```
hrishabhg: rather than environment, annotations can be good alternate.

hrishabhg (approved):

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
```diff
@@ -0,0 +1,224 @@
...
+- Infrastructure admins get IDE autocompletion and clearer documentation.
+- Micrometer hooks are not a primary extensibility point the way filters are — they don't
+  warrant the same open-ended plugin model in the CRD.
+
+### Persona ownership
+
+Micrometer configuration is a proxy-wide, infrastructure concern — it affects what metrics the
+proxy process emits and has implications for Prometheus scrape targets and dashboard
+provisioning. It belongs on `KafkaProxy` (infrastructure admin) rather than
+`VirtualKafkaCluster` (developer).
```
k-wall, line 128: I could imagine a future enhancement that allows commonTags to be also defined at the virtualCluster level.  This is out of scope for this proposal and would obviously need a operand change.

k-wall (approved):
LGTM

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
SamBarker: Good point — labels would indeed be the more natural source for this kind of per-pod metadata. The proposal has been updated to include `commonTags.fromLabels`, which maps tag names to label keys on the `KafkaProxy` CR. The operator promotes these into per-pod metric tags via the Kubernetes downward API, with the runtime resolving them at startup.

We chose labels over annotations as the source: labels are the appropriate construct for identifying and categorising resources (zone, environment, team), whereas annotations are intended for arbitrary tooling metadata with opaque string values that would require a parsing convention. This structure also gives us a natural extension point for something like`fromAnnotations`.

Note that Kubernetes label names can contain characters illegal in Prometheus tag names (`.`, `/`, `-`), so the tag name is specified explicitly — e.g. `zone: "topology.kubernetes.io/zone"` — rather than derived from the label key.

If the intent was to use annotations as the transport mechanism (operator writes the resolved value into a pod annotation, runtime reads it), the drawback is that it requires the runtime to become Kubernetes-aware — it would need API server access and a ServiceAccount with pod-read permissions. Environment variables via the downward API avoid that coupling.

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
SamBarker: Yeah I can see that we would want to control tags at that level as well, but its not a thread I want to go near today.

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
```diff
@@ -77,37 +77,48 @@ spec:
...
+
+#### `commonTags.static`: deployment-wide values
+
+Suited for values that are uniform across all replicas and known at deploy time
+(e.g. `environment: "production"`, `team: "platform"`).
+
+#### `commonTags.fromLabels`: per-pod values via label promotion
+
+Some tag values — such as availability zone — differ per pod and cannot be resolved at reconcile time because the proxy config file is shared across all replicas.
+`commonTags.fromLabels` maps tag names (valid Prometheus label names) to Kubernetes label keys on the `KafkaProxy` CR, which the operator promotes into per-pod metric tags.
```
k-wall, line 105: Some labels on the pods may come from other sources (e.g.SSA).

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
```diff
@@ -77,37 +77,48 @@ spec:
...
-| Field              | Type              | Required | Description                                                        |
-|--------------------|-------------------|----------|--------------------------------------------------------------------|
-| `standardBinders`  | array of strings  | No       | List of Micrometer binder names to register (see table below)      |
-| `commonTags`       | map<string,string> | No      | Key-value pairs added as tags to all metrics emitted by the proxy  |
+`commonTags` is an optional object grouping all common tag sources:
+
+| Field                    | Type               | Required | Description                                                              |
+|--------------------------|--------------------|----------|--------------------------------------------------------------------------|
+| `commonTags.static`      | map<string,string> | No       | Static key-value pairs, uniform across all replicas                      |
+| `commonTags.fromLabels`  | map<string,string> | No       | Tag name → label key; per-pod values resolved via the downward API       |
```
k-wall, line 95: Isn't this duplicating what https://github.com/kubernetes/kube-state-metrics gives a user already?  kube_<resource>_labels/kube_<resource>_annotations metrics

SamBarker: No, this is about using labels to tag the metrics the proxy produces rather than metrics about labels etc

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
SamBarker: Without giving the proxy the ability to query K8s apis we can't really capture any other source other than KafkaProxy as that's the only one the reconciler can configure the downwards api to populate env vars for.

k-wall:

```bash
git mv proposals/020-micrometer-configuration-in-kafkaproxy-crd.md proposals/099-configuration-in-kafkaproxy-crd.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 99 - /; t; s/^# /# 99 - /}' proposals/099-configuration-in-kafkaproxy-crd.md && rm proposals/099-configuration-in-kafkaproxy-crd.md.bak
git add proposals/099-configuration-in-kafkaproxy-crd.md
git push
```

k-wall:

```bash
git mv proposals/020-micrometer-configuration-in-kafkaproxy-crd.md proposals/099-micrometer-configuration-in-kafkaproxy-crd.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 99 - /; t; s/^# /# 99 - /}' proposals/099-micrometer-configuration-in-kafkaproxy-crd.md && rm proposals/099-micrometer-configuration-in-kafkaproxy-crd.md.bak
git add proposals/099-micrometer-configuration-in-kafkaproxy-crd.md
git push
```

Sorry for the confusion!

proposals/020-micrometer-configuration-in-kafkaproxy-crd.md
k-wall: ah yes, I got my wires crossed

robobario (approved):
LGTM, thanks @SamBarker

PR #100, opened 2026-04-17
Add sidecar injection webhook proposal

tombentley:
Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/proposals/100-sidecar-injection-webhook.md
+++ b/proposals/100-sidecar-injection-webhook.md
@@ -116,6 +116,39 @@ spec:
 
 **Why not reuse the operator's CRDs?** The operator CRDs model a shared proxy deployment with ingress networking, multi-cluster support, and cross-resource references. The sidecar use case is fundamentally simpler — one virtual cluster, localhost binding, no ingress. Coupling them would constrain both models.
 
+#### Status
+
+The CRD has a `status` subresource with the following fields:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `observedGeneration` | `int64` | The `metadata.generation` most recently observed by the webhook |
+| `conditions` | `[]Condition` | Standard Kubernetes conditions (see below) |
+
+The webhook maintains a single condition type:
+
+| Condition | Meaning |
+|-----------|---------|
+| `Ready` | The webhook has observed and accepted this configuration. `observedGeneration` on the condition tracks which generation was acknowledged. |
+
+The webhook sets `Ready=True` (reason `Accepted`) when it first observes the config via its informer. The condition is only updated when the generation changes, avoiding unnecessary status writes. If the status update fails (e.g. conflict), the webhook logs a warning but continues to inject pods normally — status is informational, not load-bearing.
+
+Example status:
+
+```yaml
+status:
+  observedGeneration: 3
+  conditions:
+    - type: Ready
+      status: "True"
+      reason: Accepted
+      message: ""
+      lastTransitionTime: "2025-01-15T10:30:00Z"
+      observedGeneration: 3
+```
+
+This gives operators visibility into whether the webhook has picked up the latest configuration, complementing the per-pod `sidecar.kroxylicious.io/config-generation` annotation for drift detection.
+
 ### Config injection
 
 The webhook generates proxy configuration YAML from the `KroxyliciousSidecarConfig` spec, using the same `Configuration` model from `kroxylicious-runtime`. The generated config is stored in a pod annotation (`sidecar.kroxylicious.io/proxy-config`) and projected into the sidecar container via a `downwardAPI` volume.
@@ -337,17 +370,16 @@ The webhook is packaged as a container image and deployed as a single-replica `D
 
 The webhook operates independently of the operator. It does not depend on the operator being deployed, does not use JOSDK, and does not reference operator CRDs.
 
-The only shared dependencies are:
+The only shared dependency is:
 
-- `kroxylicious-kubernetes-api` — for the CRD Java types
 - `kroxylicious-runtime` — for the proxy `Configuration` model, used to generate valid proxy config YAML
 
 ## Affected/not affected projects
 
 | Project | Affected | Nature of change |
 |---------|----------|-----------------|
-| `kroxylicious-kubernetes-web-hook` | Yes | New module |
-| `kroxylicious-kubernetes-api` | Yes | New CRD: `KroxyliciousSidecarConfig` |
+| `kroxylicious-kubernetes/kroxylicious-admission` | Yes | New module: webhook implementation |
+| `kroxylicious-kubernetes/kroxylicious-admission-api` | Yes | New module: `KroxyliciousSidecarConfig` CRD |
 | `kroxylicious-app` | Already merged | `classpath-plugins/` directory scanning |
 | `kroxylicious-operator` | No | |
 | `kroxylicious-runtime` | No | Used as a dependency, not modified |
--- a/proposals/100-sidecar-injection-webhook.md
+++ b/proposals/100-sidecar-injection-webhook.md
@@ -132,6 +132,7 @@ When the webhook skips injection for a reason other than pod opt-out, it labels 
 |-------|---------|
 | `no-config` | No `KroxyliciousSidecarConfig` was found for the pod's namespace |
 | `already-injected` | A container named `kroxylicious-proxy` already exists in the pod |
+| `multiple-configs` | Multiple `KroxyliciousSidecarConfig` resources exist in the namespace and no explicit config was selected via annotation |
 
 Pods that opted out via `sidecar.kroxylicious.io/injection: disabled` are not labelled — they already carry a label that identifies them.
 
--- a/proposals/100-sidecar-injection-webhook.md
+++ b/proposals/100-sidecar-injection-webhook.md
@@ -70,7 +70,7 @@ A namespaced CRD (group `kroxylicious.io`, version `v1alpha1`) defines the sidec
 
 1. **No config in namespace**: the pod is admitted without injection (debug log only). This is the common case for namespaces where the admin has enabled the namespace label but not yet created a config.
 2. **Multiple configs in namespace**: the pod is admitted without injection (warning logged). The pod can select a specific config via the `sidecar.kroxylicious.io/config` annotation; without this annotation the webhook cannot choose and skips injection.
-3. **Config is invalid in a way the webhook can detect** (e.g. malformed delegated annotation values, plugin image without a digest): the webhook logs a warning and admits the pod without injection. Consistent with fail-open semantics.
+3. **Config is invalid in a way the webhook can detect** (e.g. malformed delegated annotation values): the webhook logs a warning and admits the pod without injection. Consistent with fail-open semantics.
 4. **Config is invalid in a way only the proxy can detect** (e.g. unreachable target Kafka cluster, wrong TLS trust anchor, non-existent filter type): the webhook injects the sidecar normally. The proxy will fail its startup probe and the pod will not become ready, surfacing the problem via standard Kubernetes health-check mechanisms.
 
 ```yaml
@@ -344,11 +344,10 @@ This is the lowest-risk form of delegation — the app owner chooses a network d
 
 ### Future delegation
 
-The delegated annotations mechanism (bootstrap port, node ID range, resource requests, plugin images) described above provides a general-purpose extension point for further delegation. These are ordered roughly by blast radius:
+The delegated annotations mechanism (bootstrap port, node ID range, resource requests) described above provides a general-purpose extension point for further delegation. These are ordered roughly by blast radius:
 
 1. **Port and resource overrides** — app owner adjusts operational parameters. Low risk.
 2. **Filter configuration** — app owner adjusts parameters on admin-selected filters. Medium risk: bounded by the filter's config surface.
-3. **Plugin image selection** — app owner chooses what code runs in the proxy JVM. High risk: arbitrary code execution (see security analysis above).
 
 All delegation beyond target cluster selection requires the admin to explicitly list the delegated annotations. Nothing is delegated by default.
 
--- a/proposals/100-sidecar-injection-webhook.md
+++ b/proposals/100-sidecar-injection-webhook.md
@@ -138,9 +138,9 @@ The management endpoint binds to `0.0.0.0` because kubelet HTTP probes target th
 
 ### Native sidecar containers
 
-On Kubernetes 1.28+, the webhook injects the proxy as a native sidecar — an init container with `restartPolicy: Always`. This gives proper startup ordering (proxy starts before the application) and shutdown ordering (proxy stops after the application). On older clusters, the webhook falls back to injecting into `spec.containers`.
+On Kubernetes 1.29+ (where the `SidecarContainers` feature gate is enabled by default), the webhook injects the proxy as a native sidecar — an init container with `restartPolicy: Always`. This gives proper startup ordering (proxy starts before the application) and shutdown ordering (proxy stops after the application). On older clusters, the webhook falls back to injecting into `spec.containers`.
 
-The webhook detects the cluster's Kubernetes version at startup and chooses the injection strategy accordingly.
+The webhook detects the cluster's Kubernetes version at startup and chooses the injection strategy accordingly. For clusters running alpha versions of features (e.g. native sidecars on 1.28, OCI image volumes on 1.31-1.32), the deployer can set the `FEATURE_GATES` environment variable on the webhook (e.g. `FEATURE_GATES=SidecarContainers=true,ImageVolume=true`) to override the version-based defaults.
 
 ### Sidecar container spec
 
--- a/proposals/100-sidecar-injection-webhook.md
+++ b/proposals/100-sidecar-injection-webhook.md
@@ -51,7 +51,7 @@ Injection is opt-in at the namespace level and opt-out at the pod level, followi
 | Namespace label | `sidecar.kroxylicious.io/injection: enabled` | Webhook intercepts pod creates in this namespace |
 | Pod label | `sidecar.kroxylicious.io/injection: disabled` | Pod is excluded via `objectSelector` — never reaches the webhook |
 
-The `MutatingWebhookConfiguration` uses `namespaceSelector` to scope interception and `objectSelector` to exclude opted-out pods. The webhook itself is idempotent: if a container named `kroxylicious-proxy` already exists, injection is skipped.
+The `MutatingWebhookConfiguration` uses `namespaceSelector` to scope interception and `objectSelector` to exclude opted-out pods. The webhook itself is idempotent: if a container named `kroxylicious-proxy` already exists, injection is skipped. This is a name-based check, so a pod owner could circumvent injection by pre-adding a container with that name. This is accepted for the alpha — a determined pod owner can also opt out via labels. See [Configuration drift detection](#configuration-drift-detection) for a stronger approach planned for a future iteration.
 
 The failure policy of the webhook will be configurable.
 It will default to fail closed (`failurePolicy: Fail`), which is safe, but sacrifices availability of the Kubernetes control plane to admit workloads in cases where the webhook experiences internal errors.
@@ -246,6 +246,14 @@ kubectl get pods -n my-ns -o json | jq '[.items[] |
 ```
 
 In a future iteration, a reconciler could watch for pods with outdated generations and surface an `UpToDate` condition on the `KroxyliciousSidecarConfig` status, giving operators visibility into configuration drift without requiring manual queries.
+
+#### Annotation-based idempotency (future)
+
+The current idempotency check (container name) is weak: it can be circumvented by a pod owner pre-adding a container named `kroxylicious-proxy`, and it does not detect configuration drift during reinvocation. A stronger approach is to compare the `sidecar.kroxylicious.io/proxy-config` annotation on the pod against the configuration the webhook would generate right now. If the annotation is absent or differs, the webhook injects (or re-injects); if it matches, injection is skipped.
+
+This works because `ProxyConfigGenerator.generateConfig()` is deterministic: given the same `KroxyliciousSidecarConfigSpec` inputs, Jackson serialisation produces identical YAML. String equality on the annotation value is therefore a reliable idempotency check within the same webhook version. This also handles the reinvocation case (`reinvocationPolicy: IfNeeded`): if another mutating webhook modifies the pod after initial injection, the webhook can verify that the proxy config annotation is still correct.
+
+The generation stamp remains useful for fleet-wide drift detection (comparing against the current `KroxyliciousSidecarConfig` generation without reconstructing the expected config), but the annotation comparison becomes the primary idempotency mechanism.
 
 ### Third-party plugin support
 
--- a/proposals/100-sidecar-injection-webhook.md
+++ b/proposals/100-sidecar-injection-webhook.md
@@ -1,0 +1,444 @@
+# Proposal 016 — Sidecar Injection Webhook
+
+A Kubernetes mutating admission webhook that automatically injects a Kroxylicious proxy sidecar into application pods. The sidecar intercepts Kafka traffic on localhost, allowing filters to be applied transparently without changes to the application.
+
+## Current situation
+
+Kroxylicious is deployed either standalone or via the operator as a shared proxy tier, fronting one or more Kafka clusters with ingress networking. Applications connect to the proxy over the network.
+
+There is no mechanism for running Kroxylicious as a per-pod sidecar. Users who want per-pod proxying must manually construct the sidecar container spec, generate proxy configuration, and manage the lifecycle themselves.
+
+The proxy already has properties that make it suitable for sidecar use: it can bind to localhost, and runs as non-root with no special capabilities.
+
+## Motivation
+
+A sidecar model is useful when:
+
+- The application should connect to Kafka via `localhost` rather than through a shared proxy tier.
+- Per-pod filter configuration is needed (e.g. different encryption keys per tenant).
+- The organisation prefers a service mesh-style deployment where each pod carries its own proxy.
+
+Manual sidecar construction is error-prone and creates a maintenance burden. A webhook automates injection, enforces a consistent security posture, and gives the webhook administrator control over what runs in the sidecar.
+
+## Proposal
+
+### Trust model
+
+The webhook operates under a strict two-party trust model:
+
+- **Webhook administrator**: controls what gets injected — the proxy image, upstream Kafka address, filter definitions, security context. These are never overridable by the app owner.
+- **Application pod owner**: can opt out of injection, and may override specific settings (bootstrap port, node ID range, resource requests) if the admin explicitly delegates those annotations.
+
+The webhook always overwrites the `kroxylicious.io/proxy-config` annotation on the pod, regardless of any value the app owner may have set. Non-delegated annotations in the `kroxylicious.io/` namespace are silently ignored with a logged warning.
+
+### Injection decision
+
+Injection is opt-in at the namespace level and opt-out at the pod level, following the Istio/Linkerd convention:
+
+| Mechanism | Key | Effect |
+|-----------|-----|--------|
+| Namespace label | `kroxylicious.io/sidecar-injection: enabled` | Webhook intercepts pod creates in this namespace |
+| Pod label | `kroxylicious.io/inject-sidecar: "false"` | Pod is excluded via `objectSelector` — never reaches the webhook |
+
+The `MutatingWebhookConfiguration` uses `namespaceSelector` to scope interception and `objectSelector` to exclude opted-out pods. The webhook itself is idempotent: if a container named `kroxylicious-proxy` already exists, injection is skipped.
+
+The webhook always returns `allowed: true`, even on internal errors. Errors are logged; the pod is admitted unmodified. This fail-open policy (`failurePolicy: Ignore`) ensures a broken webhook never blocks workloads.
+
+### CRD: `KroxyliciousSidecarConfig`
+
+A namespaced CRD (group `kroxylicious.io`, version `v1alpha1`) defines the sidecar configuration. The webhook admin creates one per namespace. The following edge cases are handled:
+
+1. **No config in namespace**: the pod is admitted without injection (debug log only). This is the common case for namespaces where the admin has enabled the namespace label but not yet created a config.
+2. **Multiple configs in namespace**: the pod is admitted without injection (warning logged). The pod can select a specific config via the `kroxylicious.io/sidecar-config` annotation; without this annotation the webhook cannot choose and skips injection.
+3. **Config is invalid in a way the webhook can detect** (e.g. malformed delegated annotation values, plugin image without a digest): the webhook logs a warning and admits the pod without injection. Consistent with fail-open semantics.
+4. **Config is invalid in a way only the proxy can detect** (e.g. unreachable upstream Kafka, wrong TLS trust anchor, non-existent filter type): the webhook injects the sidecar normally. The proxy will fail its startup probe and the pod will not become ready, surfacing the problem via standard Kubernetes health-check mechanisms.
+
+```yaml
+apiVersion: kroxylicious.io/v1alpha1
+kind: KroxyliciousSidecarConfig
+metadata:
+  name: my-config
+spec:
+  upstreamBootstrapServers: kafka-prod.internal:9092
+  bootstrapPort: 19092             # default, configurable
+  nodeIdRange:
+    startInclusive: 0
+    endInclusive: 2
+  managementPort: 9190
+  proxyImage: quay.io/kroxylicious/proxy:0.21.0   # optional override
+  setBootstrapEnvVar: true         # sets KAFKA_BOOTSTRAP_SERVERS on app containers
+  filterDefinitions:
+    - name: my-filter
+      type: io.example.MyFilterFactory
+      config: { ... }
+  upstreamTls:
+    trustAnchorSecretRef:
+      name: kafka-ca
+      key: ca.crt
+  plugins:
+    - name: my-plugin
+      image:
+        reference: registry.example.com/my-filter:v1.0@sha256:abc123
+        pullPolicy: IfNotPresent
+  delegatedAnnotations:
+    - kroxylicious.io/sidecar-bootstrap-port
+    - kroxylicious.io/sidecar-node-id-range
+```
+
+**Why a CRD, not a ConfigMap?** Schema validation by the API server, RBAC separation (admin creates, app owners can't modify), status conditions for observability, consistency with the existing Kroxylicious Kubernetes API.
+
+**Why not reuse the operator's CRDs?** The operator CRDs model a shared proxy deployment with ingress networking, multi-cluster support, and cross-resource references. The sidecar use case is fundamentally simpler — one virtual cluster, localhost binding, no ingress. Coupling them would constrain both models.
+
+### Config injection
+
+The webhook generates proxy configuration YAML from the `KroxyliciousSidecarConfig` spec, using the same `Configuration` model from `kroxylicious-runtime`. The generated config is stored in a pod annotation (`kroxylicious.io/proxy-config`) and projected into the sidecar container via a `downwardAPI` volume.
+
+This avoids creating per-pod ConfigMaps, which would require additional RBAC, lifecycle management for orphaned ConfigMaps, and unique name generation. The annotation approach is self-contained within the pod.
+
+A typical sidecar config is a few hundred bytes, well within the ~256KB practical annotation size limit.
+
+### Port allocation
+
+| Port | Purpose | Bind address |
+|------|---------|-------------|
+| 19092 | Kafka bootstrap | `localhost` |
+| 19093+ | Per-broker ports (one per node ID) | `localhost` |
+| 9190 | Management (`/livez`, `/metrics`) | `0.0.0.0` |
+
+Bootstrap defaults to 19092 rather than 9092 to avoid clashing with Kafka client libraries' default port. The webhook sets `KAFKA_BOOTSTRAP_SERVERS=localhost:19092` on application containers (configurable, can be disabled).
+
+The management endpoint binds to `0.0.0.0` because kubelet HTTP probes target the pod IP, not loopback. This means the application container can also reach `/livez` and `/metrics`, but neither endpoint exposes sensitive data.
+
+### Native sidecar containers
+
+On Kubernetes 1.28+, the webhook injects the proxy as a native sidecar — an init container with `restartPolicy: Always`. This gives proper startup ordering (proxy starts before the application) and shutdown ordering (proxy stops after the application). On older clusters, the webhook falls back to injecting into `spec.containers`.
+
+The webhook detects the cluster's Kubernetes version at startup and chooses the injection strategy accordingly.
+
+### Sidecar container spec
+
+The injected sidecar follows the same patterns as `ProxyDeploymentDependentResource` in the operator:
+
+- `securityContext`: `allowPrivilegeEscalation: false`, `capabilities: drop ALL`, `readOnlyRootFilesystem: true`
+- Probes: `startupProbe` (30 x 2s), `livenessProbe`, `readinessProbe` — all HTTP GET `/livez` on port 9190
+- `terminationMessagePolicy: FallbackToLogsOnError`
+
+The security context is never weakened. If the pod already has a stricter security context, it is preserved.
+
+### Upstream TLS
+
+When `spec.upstreamTls.trustAnchorSecretRef` is set, the webhook adds a volume mounting the referenced Secret into the sidecar and configures the proxy to use it as a PEM trust store. The Secret must exist in the pod's namespace.
+
+### Delegated annotations
+
+The `delegatedAnnotations` field in `KroxyliciousSidecarConfig` lists which annotations the app owner may set to override sidecar parameters:
+
+| Annotation | Effect |
+|-----------|--------|
+| `kroxylicious.io/sidecar-bootstrap-port` | Override bootstrap port |
+| `kroxylicious.io/sidecar-node-id-range` | Override node ID range (e.g. `"0-5"`) |
+| `kroxylicious.io/sidecar-resources-cpu` | Override CPU request/limit |
+| `kroxylicious.io/sidecar-resources-memory` | Override memory request/limit |
+| `kroxylicious.io/sidecar-plugin-images` | Additional plugin images (JSON array) |
+
+Delegation is opt-in per annotation. By default nothing is delegated. Upstream Kafka address, filter definitions, proxy image, and security context are never delegatable.
+
+### Configuration drift detection
+
+The webhook stamps each injected pod with metadata annotations:
+
+| Annotation | Value |
+|-----------|-------|
+| `kroxylicious.io/config-generation` | The `metadata.generation` of the `KroxyliciousSidecarConfig` at injection time |
+| `kroxylicious.io/injection-timestamp` | ISO-8601 timestamp of when the sidecar was injected |
+
+Because the webhook only mutates pods at creation time, configuration changes to `KroxyliciousSidecarConfig` do not propagate to running pods. This matches how Istio and Linkerd handle sidecar injection. Users must restart pods to pick up new configuration.
+
+The generation stamp allows operators to identify stale pods:
+
+```
+kubectl get pods -n my-ns -o json | jq '[.items[] |
+  select(.metadata.annotations["kroxylicious.io/config-generation"] != null) |
+  {name: .metadata.name, generation: .metadata.annotations["kroxylicious.io/config-generation"]}]'
+```
+
+In a future iteration, a reconciler could watch for pods with outdated generations and surface an `UpToDate` condition on the `KroxyliciousSidecarConfig` status, giving operators visibility into configuration drift without requiring manual queries.
+
+### Third-party plugin support
+
+#### The problem
+
+Users will want to run third-party Kroxylicious plugins (custom filters, KMS providers) in the sidecar. The proxy discovers plugins via `ServiceLoader` from the classpath. Built-in plugins live in `libs/`. Third-party plugin JARs must be delivered separately.
+
+OCI image volumes (KEP-4639) allow mounting an OCI image as a read-only volume in a pod. This is the cleanest delivery mechanism: plugin authors package their JARs in a `FROM scratch` image, and the webhook mounts it into the sidecar at a known path.
+
+#### Solution
+
+The proxy startup script already scans `/opt/kroxylicious/classpath-plugins/*/` for subdirectories containing JARs and adds them to the classpath. The webhook mounts each plugin's OCI image at `/opt/kroxylicious/classpath-plugins/<name>/`.
+
+For each plugin in `spec.plugins`, the webhook adds:
+
+1. An OCI image volume referencing the plugin image.
+2. A read-only volume mount on the sidecar container.
+
+```yaml
+volumes:
+  - name: plugin-my-filter
+    image:
+      reference: registry.example.com/my-filter:v1.0@sha256:abc123
+      pullPolicy: IfNotPresent
+```
+
+```yaml
+volumeMounts:
+  - name: plugin-my-filter
+    mountPath: /opt/kroxylicious/classpath-plugins/my-filter
+    readOnly: true
+```
+
+ServiceLoader discovers the plugin implementations from the combined classpath. Multiple plugin images can be mounted simultaneously, each at its own subdirectory.
+
+#### Plugin image convention
+
+Plugin images should be built `FROM scratch` with JARs at the image root:
+
+```dockerfile
+FROM scratch
+COPY target/my-filter.jar /my-filter.jar
+COPY target/dependency/*.jar /
+```
+
+#### Flat classpath limitations
+
+All plugin JARs share the proxy's flat classpath. There is no classloader isolation. If two plugins bundle different versions of the same library, the one the classloader finds first wins — silently, without error.
+
+Jackson is the concrete concern. The proxy ships Jackson and uses it for filter config deserialization. A plugin bundling an incompatible Jackson version could cause silent serialization differences. Other proxy-provided libraries (Netty, Kafka clients, SLF4J, Micrometer) carry the same risk.
+
+**Mitigations:**
+
+- **Document the constraint**: plugin images should not bundle libraries the proxy already provides. Plugin authors should treat the proxy's dependencies as `provided` scope. Publishing the proxy's transitive dependency closure as a Maven BOM would make this mechanical.
+- **Shade transitive dependencies**: plugin authors should shade (relocate) any transitive dependency that might conflict.
+
+Classloader isolation (a classloader per plugin directory, similar to what application servers do) would eliminate this problem but is a significant architectural change. It should be treated as a known future requirement, not a hypothetical.
+
+#### Kubernetes version requirements
+
+| Feature | K8s version | OpenShift version | Status |
+|---------|-------------|-------------------|--------|
+| OCI image volumes (alpha) | 1.31+ | 4.18+ | Feature gate `ImageVolume` must be enabled |
+| OCI image volumes (beta) | 1.33+ | 4.20+ | Feature gate `ImageVolume` must be enabled |
+| OCI image volumes (default on) | 1.35+ | 4.22+ | Enabled by default |
+
+**Container runtime support**: OpenShift uses CRI-O exclusively, which supports OCI image volumes from v1.31+ (matching the Kubernetes version). containerd support is maturing (alpha in v2.1.0) but is not relevant to OpenShift deployments. For non-OpenShift clusters using containerd, the init-container fallback (below) is the practical path until containerd support stabilises.
+
+#### Init-container fallback
+
+For clusters without OCI image volume support, the webhook supports an init-container fallback:
+
+1. An init container per plugin image copies JARs to an `emptyDir` volume.
+2. The `emptyDir` is mounted at `/opt/kroxylicious/classpath-plugins/<name>` on the sidecar with `readOnly: true`.
+
+This works on any Kubernetes version but adds startup latency and uses writable storage (though the mount itself is read-only on the sidecar).
+
+The webhook auto-detects OCI image volume support via the Kubernetes API server version.
+
+#### Security analysis: admin-controlled plugin images
+
+When the admin specifies plugin images in `KroxyliciousSidecarConfig.spec.plugins`:
+
+- The admin trusts the plugin image publisher.
+- The app owner has no control over which images are mounted.
+- OCI image volumes are read-only and `noexec` by design.
+- Plugin JARs run on the proxy's classpath with the proxy's JVM permissions (non-root, no capabilities, read-only root filesystem).
+
+**Remaining risks:**
+
+- **Supply chain**: a compromised plugin image contains malicious code with access to Kafka traffic and mounted credentials. Mitigate with image signing and digest-pinned references.
+- **Dependency conflicts**: as described above under flat classpath limitations.
+
+#### Security analysis: delegated plugin image selection
+
+If the admin delegates plugin image selection to app owners (via the `kroxylicious.io/sidecar-plugin-images` annotation), the risks escalate:
+
+| Risk | Severity | Description |
+|------|----------|-------------|
+| Arbitrary code on proxy classpath | Critical | App owner specifies an image containing malicious JARs. The proxy has access to upstream Kafka credentials, TLS certs, and all Kafka traffic. |
+| Registry credential leakage | High | OCI image volumes reuse the pod's `imagePullSecrets` and node-level credentials. An attacker-controlled registry receives pull requests carrying bearer tokens. |
+| Image tag mutation | High | A tag (not a digest) can be replaced with malicious content between pulls. |
+| Resource exhaustion | Medium | Large OCI images consume node disk. No per-volume size limits exist. |
+
+**Mitigations:**
+
+1. **Registry allow-list**: `allowedPluginRegistries` in `KroxyliciousSidecarConfig`. The webhook rejects delegated image references not matching an allowed prefix.
+2. **Require digest pinning**: delegated image references without `@sha256:` are rejected.
+3. **Audit logging**: all plugin image references are logged at INFO, with warnings for delegated images.
+
+Even with these mitigations, the fundamental trust grant is unchanged: a digest-pinned image from an allowed registry still runs arbitrary code in the proxy JVM. The mitigations reduce supply-chain risk but do not constrain what the code does once loaded. Delegating plugin image selection is equivalent to allowing the app owner to run arbitrary code with the proxy's identity.
+
+**Delegation is disabled by default.** When enabled, both `allowedPluginRegistries` and digest pinning are enforced.
+
+#### Future direction: PluginRegistry CRD
+
+A cleaner model for pre-approved plugins would be a cluster-scoped `PluginRegistry` CRD where the admin defines approved plugin images with namespace-level scoping:
+
+```yaml
+apiVersion: kroxylicious.io/v1alpha1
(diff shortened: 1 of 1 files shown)


proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+
+#### Defence-in-depth: NetworkPolicy + upstream authentication
+
+A stronger enforcement model combines two mechanisms:
+
+1. **NetworkPolicy restricting pod egress** — the admin creates a `NetworkPolicy` in the application namespace that limits egress to only the upstream Kafka cluster IPs/ports defined in `allowedUpstreamClusters`. This prevents the application from connecting to arbitrary external services. Because NetworkPolicy operates at the pod level, both the sidecar and the application container are subject to the same egress rules — but this is acceptable, because the sidecar only needs to reach the allowed clusters.
+
+2. **Upstream Kafka authentication** — the upstream Kafka cluster requires authentication (mTLS client certificates or SASL credentials). The sidecar holds the credentials; the application does not. Even if the application connects directly to the Kafka broker's address, the broker rejects the unauthenticated connection.
```
robobario: on this, we don't have a Proxy SASL Initiation Filter yet, does this design require that as a prerequisite? I'm wondering what the very initial SASL requirements will be, what mechanisms we must support. Should the proxy reject attempted downstream SASL handshakes?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+There is no mechanism for running Kroxylicious as a per-pod sidecar. Users who want per-pod proxying must manually construct the sidecar container spec, generate proxy configuration, and manage the lifecycle themselves.
+
+The proxy already has properties that make it suitable for sidecar use: it can bind to localhost, and runs as non-root with no special capabilities.
+
+## Motivation
+
+A sidecar model is useful when:
+
+- The application should connect to Kafka via `localhost` rather than through a shared proxy tier.
+- Per-pod filter configuration is needed (e.g. different encryption keys per tenant).
```
robobario, line 18: On Filter Configuration. Will we have a mechanism for mounting in the files (secrets, rules files) required for encryption and authz? Will we (can we?) support something similar to the Operator configTemplate?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+When `spec.upstreamTls.trustAnchorSecretRef` is set, the webhook adds a volume mounting the referenced Secret into the sidecar and configures the proxy to use it as a PEM trust store. The Secret must exist in the pod's namespace.
+
+### Delegated annotations
+
+The `delegatedAnnotations` field in `KroxyliciousSidecarConfig` lists which annotations the app owner may set to override sidecar parameters:
+
+| Annotation | Effect |
+|-----------|--------|
+| `kroxylicious.io/sidecar-bootstrap-port` | Override bootstrap port |
+| `kroxylicious.io/sidecar-node-id-range` | Override node ID range (e.g. `"0-5"`) |
```
robobario: what's the reason for having the node-id-range override on the pod? It's a property of the upstream cluster so why would apps need to change it?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+| `kroxylicious.io/sidecar-resources-cpu` | Override CPU request/limit |
+| `kroxylicious.io/sidecar-resources-memory` | Override memory request/limit |
+| `kroxylicious.io/sidecar-plugin-images` | Additional plugin images (JSON array) |
```
robobario: What would this be used for? What's not covered by the CRD plugins array?

proposals/100-sidecar-injection-webhook.md
tombentley: This section got added because I felt that we should not assume that the app dev persona (who owns the application) _wants_ to use the proxy. I suspect that we can't exactly compel them to use the proxy (i.e. this probably cannot be a cast iron security-like guarantee). But at the very least it should be difficult to accidentally _not_ use the proxy.

I don't think this kind of authn this is really a prerequisite: we can make progress without this.

We should be explicit about the goal here though: 
* Should we aim to support a kind of weak trust boundary between policy owner and app owner maybe not immediately but eventually?
* Or should we explicitly rule out such a thing? This might be a better choice if it cannot be a strong boundary, and if having a boundary is not going to be an important requirement.

proposals/100-sidecar-injection-webhook.md
tombentley: If you're asking: "What's the use case for delegating plugin images specifically?" Well, perhaps there's proxy functionality that the app owner wants to use as well. For example record validation. 

That said, actually doing that safely would be a tricky. Perhaps there's a specific point in the filter chain where the app owner's filters can safely go, but they shouldn't be allowed to to change the filter chain entirely.  I'm happy to remove this from the proposal, because it's clearly more complicated than it really warranted at this point.

proposals/100-sidecar-injection-webhook.md
tombentley: TBH, I think we should remove `sidecar-bootstrap-port` and `sidecar-node-id-range`. I don't think that's really needed initially. Later on in this doc there's a mechanism for selecting pre-configured target clusters by name (e.g. `kroxylicious.io/sidecar-upstream-cluster: staging`), which seems like a better way to do it. However, I think we can defer that too initially.

proposals/100-sidecar-injection-webhook.md
tombentley: This is a real pain. To exist as a sidecar we're having to push everything that the proxy configuration depends on (including things like Secrets) into the application namespace as resources. That's the opposite of a nice, clean separation of responsibilities, and is a step away from being able to have a meaningful trust boundary between application and proxy.

The need to reify all the configuration as Kubernetes resources arises because we don't have a proxy control plane. I think that the right fix would really be to address that point directly. The problem is that building a sophisticated enough control plane is a lot of effort. So the question is "Can we build something useful without one, and add a control plane later?" 

I think the answer to that is "yes", because the trust boundary aspect is not an important consideration for all use cases of the side car injection, e.g. when the application owner is the proxy configuration owner then the reification is not really a problem at all.

proposals/100-sidecar-injection-webhook.md
robobario: ah I see, so it was about letting the app owner have some influence over the proxy. Like give them the ability to mixin some filters into the chain per pod. Yeah I think this can be safely removed, we could come up with a mechanism later if required. For now they can at least use multiple KroxyliciousSidecarConfig to give pod X and pod Y in the same namespace different configurations.

k-wall:
Discussed at Community Call 23rd April. Tom is looking for feedback.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+## Proposal
+
+### Trust model
+
+The webhook operates under a strict two-party trust model:
+
+- **Webhook administrator**: controls what gets injected — the proxy image, upstream Kafka address, filter definitions, security context. These are never overridable by the app owner.
+- **Application pod owner**: can opt out of injection, and may override specific settings (bootstrap port, node ID range, resource requests) if the admin explicitly delegates those annotations.
+
+The webhook always overwrites the `kroxylicious.io/proxy-config` annotation on the pod, regardless of any value the app owner may have set. Non-delegated annotations in the `kroxylicious.io/` namespace are silently ignored with a logged warning.
```
k-wall: This is the first mention of kroxylicious.io/proxy-config.  You should introduce what it is for and how it is used.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+Injection is opt-in at the namespace level and opt-out at the pod level, following the Istio/Linkerd convention:
+
+| Mechanism | Key | Effect |
+|-----------|-----|--------|
+| Namespace label | `kroxylicious.io/sidecar-injection: enabled` | Webhook intercepts pod creates in this namespace |
+| Pod label | `kroxylicious.io/inject-sidecar: "false"` | Pod is excluded via `objectSelector` — never reaches the webhook |
+
+The `MutatingWebhookConfiguration` uses `namespaceSelector` to scope interception and `objectSelector` to exclude opted-out pods. The webhook itself is idempotent: if a container named `kroxylicious-proxy` already exists, injection is skipped.
+
+The webhook always returns `allowed: true`, even on internal errors. Errors are logged; the pod is admitted unmodified. This fail-open policy (`failurePolicy: Ignore`) ensures a broken webhook never blocks workloads.
```
k-wall: Is this a recommend best practice?  Or are you following an established convention?  What does Istio do?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+  name: my-config
+spec:
+  upstreamBootstrapServers: kafka-prod.internal:9092
+  bootstrapPort: 19092             # default, configurable
+  nodeIdRange:
+    startInclusive: 0
+    endInclusive: 2
+  managementPort: 9190
+  proxyImage: quay.io/kroxylicious/proxy:0.21.0   # optional override
+  setBootstrapEnvVar: true         # sets KAFKA_BOOTSTRAP_SERVERS on app containers
```
k-wall: question to self.  Does the java kafka client understand this env var itself? Other clients?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+        reference: registry.example.com/my-filter:v1.0@sha256:abc123
+        pullPolicy: IfNotPresent
+  delegatedAnnotations:
+    - kroxylicious.io/sidecar-bootstrap-port
+    - kroxylicious.io/sidecar-node-id-range
+```
+
+**Why a CRD, not a ConfigMap?** Schema validation by the API server, RBAC separation (admin creates, app owners can't modify), status conditions for observability, consistency with the existing Kroxylicious Kubernetes API.
+
+**Why not reuse the operator's CRDs?** The operator CRDs model a shared proxy deployment with ingress networking, multi-cluster support, and cross-resource references. The sidecar use case is fundamentally simpler — one virtual cluster, localhost binding, no ingress. Coupling them would constrain both models.
```
k-wall: How are we going to handle secrets for filters (e.g. record encryption)?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Port allocation
+
+| Port | Purpose | Bind address |
+|------|---------|-------------|
+| 19092 | Kafka bootstrap | `localhost` |
+| 19093+ | Per-broker ports (one per node ID) | `localhost` |
+| 9190 | Management (`/livez`, `/metrics`) | `0.0.0.0` |
+
+Bootstrap defaults to 19092 rather than 9092 to avoid clashing with Kafka client libraries' default port. The webhook sets `KAFKA_BOOTSTRAP_SERVERS=localhost:19092` on application containers (configurable, can be disabled).
```
k-wall: Curious why you didn't chose to use the IANA standard port.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Delegated annotations
+
+The `delegatedAnnotations` field in `KroxyliciousSidecarConfig` lists which annotations the app owner may set to override sidecar parameters:
+
+| Annotation | Effect |
+|-----------|--------|
+| `kroxylicious.io/sidecar-bootstrap-port` | Override bootstrap port |
+| `kroxylicious.io/sidecar-node-id-range` | Override node ID range (e.g. `"0-5"`) |
+| `kroxylicious.io/sidecar-resources-cpu` | Override CPU request/limit |
```
k-wall: I think cpu/memory overrides are sensible.  We are bound to see a situation where one app needs a bit more memory in its proxy than the rest of the fleet (say an app that sends particular large records)

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+### Configuration drift detection
+
+The webhook stamps each injected pod with metadata annotations:
+
+| Annotation | Value |
+|-----------|-------|
+| `kroxylicious.io/config-generation` | The `metadata.generation` of the `KroxyliciousSidecarConfig` at injection time |
+| `kroxylicious.io/injection-timestamp` | ISO-8601 timestamp of when the sidecar was injected |
+
+Because the webhook only mutates pods at creation time, configuration changes to `KroxyliciousSidecarConfig` do not propagate to running pods. This matches how Istio and Linkerd handle sidecar injection. Users must restart pods to pick up new configuration.
```
k-wall: Sounds very sensible

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+
+In a future iteration, a reconciler could watch for pods with outdated generations and surface an `UpToDate` condition on the `KroxyliciousSidecarConfig` status, giving operators visibility into configuration drift without requiring manual queries.
+
+### Third-party plugin support
+
+#### The problem
+
+Users will want to run third-party Kroxylicious plugins (custom filters, KMS providers) in the sidecar. The proxy discovers plugins via `ServiceLoader` from the classpath. Built-in plugins live in `libs/`. Third-party plugin JARs must be delivered separately.
+
+OCI image volumes (KEP-4639) allow mounting an OCI image as a read-only volume in a pod. This is the cleanest delivery mechanism: plugin authors package their JARs in a `FROM scratch` image, and the webhook mounts it into the sidecar at a known path.
```
k-wall, line 305: Yay! I wanted to this feature since the first day I work on the Proxy.   We should add it  support to it to the operator too (obvious separate).

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+#### Init-container fallback
+
+For clusters without OCI image volume support, the webhook supports an init-container fallback:
+
+1. An init container per plugin image copies JARs to an `emptyDir` volume.
+2. The `emptyDir` is mounted at `/opt/kroxylicious/classpath-plugins/<name>` on the sidecar with `readOnly: true`.
+
+This works on any Kubernetes version but adds startup latency and uses writable storage (though the mount itself is read-only on the sidecar).
+
+The webhook auto-detects OCI image volume support via the Kubernetes API server version.
```
k-wall, line 377: Does the server version info reveal which beta features are enabled?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+
+#### The problem
+
+Users will want to run third-party Kroxylicious plugins (custom filters, KMS providers) in the sidecar. The proxy discovers plugins via `ServiceLoader` from the classpath. Built-in plugins live in `libs/`. Third-party plugin JARs must be delivered separately.
+
+OCI image volumes (KEP-4639) allow mounting an OCI image as a read-only volume in a pod. This is the cleanest delivery mechanism: plugin authors package their JARs in a `FROM scratch` image, and the webhook mounts it into the sidecar at a known path.
+
+#### Solution
+
+The proxy startup script already scans `/opt/kroxylicious/classpath-plugins/*/` for subdirectories containing JARs and adds them to the classpath. The webhook mounts each plugin's OCI image at `/opt/kroxylicious/classpath-plugins/<name>/`.
```
k-wall, line 309: I don't think find the classpath plugins feature gives us a defined order for loading jars.  That might lead to inconsistent behaviour, if plugin authors ship different versions of the same JAR file.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+  setBootstrapEnvVar: true         # sets KAFKA_BOOTSTRAP_SERVERS on app containers
+  filterDefinitions:
+    - name: my-filter
+      type: io.example.MyFilterFactory
+      config: { ... }
+  upstreamTls:
+    trustAnchorSecretRef:
+      name: kafka-ca
+      key: ca.crt
+  plugins:
```
k-wall, line 109: We've a slot here where we could add a client cert, if the need arises.

proposals/100-sidecar-injection-webhook.md
k-wall: No, it isn't.   The app will need wire it up. That's fine.

proposals/100-sidecar-injection-webhook.md
k-wall: I think we will need to address this. 

 I can foresee that security folks will have a hard requirement that secrets don't appear in pod annotations (where they are available to kube admins) and potentially flow to [metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md).

We will need something like our templating $(configmap::..)

k-wall:
Thanks @tombentley, looks like a solid proposal.  I've left a few comments to be considered.

proposals/100-sidecar-injection-webhook.md
tombentley: I apologise. Claude was adamant that Istio was fail open, and I was foolish enough to take that at face value (note to self). Istio is configurable, but fail closed by default. I've changed what's proposed to match. I think for an alpha feature we could get away with just implementing one, but it doesn't seem like a lot of work to support both out of the gate.

proposals/100-sidecar-injection-webhook.md
tombentley: Correct. But my understanding is that using non-TLS and binding only to localhost is viewed as acceptable security-wise.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,444 @@
...
+kind: KroxyliciousSidecarConfig
+metadata:
+  name: my-config
+spec:
+  upstreamBootstrapServers: kafka-prod.internal:9092
+  bootstrapPort: 19092             # default, configurable
+  nodeIdRange:
+    startInclusive: 0
+    endInclusive: 2
+  managementPort: 9190
```
tombentley: I think we should make this < 9092, to avoid the problem that Strimzi had when the gap that was originally allocated between the start of the broker listeners and the non-listener ports ended up being too small.

proposals/100-sidecar-injection-webhook.md
tombentley: 9092 is not IANA allocated, it's just a Kafka ecosystem convention. But I've changed to 9092.

proposals/100-sidecar-injection-webhook.md
tombentley: Not really. Because the proxy configuration is built from the `KroxyliciousSidecarConfig` at injection time it means that any subsequent changes to the `KroxyliciousSidecarConfig` are ignored until the next time the pod get created. That creates a lot of problems:

* The user who has RBAC on the `KroxyliciousSidecarConfig` might not have RBAC on the `Pod`, `Deployment` etc. So it assumes some kind of communication between those users. Such communication might not be possible in practice. 
* Erroneous changes to the `KroxyliciousSidecarConfig` are only discovered at some unbounded time later.

The right way to solve this would be for the proxy to phone home to obtain a configuration from a control plane, and to receive updates to that configuration without the need for Pod restarts. But I really want to avoid boiling the ocean here by blocking this **alpha** sidecar feature on having such a control plane. I think the control plane piece is something we could add later (also as an alpha feature) and then maybe they would become stable features together later on.

proposals/100-sidecar-injection-webhook.md
tombentley: Yes, me too.

proposals/100-sidecar-injection-webhook.md
tombentley: Mentioned as a limitation.

proposals/100-sidecar-injection-webhook.md
tombentley: @k-wall I think you misunderstand (but maybe I do). I'm saying that the secrets themselves end up as volume mounts:
* So no risk of accidental disclosure to people who can merely read the annotations.  
* But that that isn't good enough either, in the long run, because those secrets are still visible to the application owner. 

But I'm again trying to avoid proposing a control plane here, in the interests of being able to **experiment** with conveniently deploying the pod as a sidecar.

proposals/100-sidecar-injection-webhook.md
tombentley: Clarified

proposals/100-sidecar-injection-webhook.md
tombentley: They'd be referenced as KeyStores in the proxy config in the annotation projected to the container. 
Then we'd use volume mounts so get those referenced files visible to the container. 
This is a significant weakness (given we want a trust model where the application owner is not necessarily trusted by the proxy administrator), but I think we have to start somewhere.

proposals/100-sidecar-injection-webhook.md
tombentley: Taken these sections on image delegation out entirely for now.

tombentley (commented):
Thanks @k-wall @robobario, I think I've addresses your comments, though with a watering down of some of the initial ambition wrt trust boundaries. PTAL.

proposals/100-sidecar-injection-webhook.md
robobario: are you imagining the `configTemplate` style, where we generate the mounts from the Filter config? Or something more raw, like the `KroxyliciousSidecarConfig` gives the user the ability to mount secrets, more of a pod template style?

proposals/100-sidecar-injection-webhook.md
robobario: Having a look there is a discovery API https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-v3 where we can look for `image` in the Volume properties at webhook startup.

On minikube with k8s v1.33.1

```
kubectl proxy --port=8001 # in another terminal
curl localhost:8001/openapi/v3/api/v1 | jq '.components.schemas."io.k8s.api.core.v1.Volume".properties.image'
{
  "description": "image represents an OCI object (a container image or artifact) pulled and ... snip ...",
  "allOf": [
    {
      "$ref": "#/components/schemas/io.k8s.api.core.v1.ImageVolumeSource"
    }
  ]
}
```

and then on k8s v1.30.14

```
curl -sS localhost:8001/openapi/v3/api/v1 | jq '.components.schemas."io.k8s.api.core.v1.Volume".properties.image' | head
null
```

Gemini is telling me this would likely need RBAC changes to allow the webhook to use the discovery API, by giving it the builtin `system:discovery` Role.

```
roleRef:
  kind: ClusterRole
  name: system:discovery
  apiGroup: rbac.authorization.k8s.io
```

k-wall:

```bash
git mv proposals/100-sidecar-injection-webhook.md proposals/100-injection-webhook.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 100 - /; t; s/^# /# 100 - /}' proposals/100-injection-webhook.md && rm proposals/100-injection-webhook.md.bak
git add proposals/100-injection-webhook.md
git push
```

k-wall:

```bash
git mv proposals/100-sidecar-injection-webhook.md proposals/100-sidecar-injection-webhook.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 100 - /; t; s/^# /# 100 - /}' proposals/100-sidecar-injection-webhook.md && rm proposals/100-sidecar-injection-webhook.md.bak
git add proposals/100-sidecar-injection-webhook.md
git push
```

Sorry for the confusion!

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,375 @@
...
+
+**TLS**: Kubernetes requires HTTPS for admission webhooks. The primary path uses cert-manager with a self-signed issuer. A manual alternative (admin provides cert/key Secret) is documented. The webhook watches cert files for rotation and reloads the SSLContext.
+
+**RBAC**: The webhook needs only `get`, `list`, `watch` on `KroxyliciousSidecarConfig` resources and `get`, `list`, `watch` on namespaces. No ConfigMap or Secret creation permissions are needed.
+
+**HTTP server**: Uses the JDK built-in `HttpsServer` (same pattern as `OperatorMain.java`), serving `POST /mutate` and `GET /livez`. No additional HTTP framework dependencies.
+
+### Independence from the operator
+
+The webhook operates independently of the operator. It does not depend on the operator being deployed, does not use JOSDK, and does not reference operator CRDs.
```
SamBarker, line 429: The proposal states the webhook doesn't use JOSDK but doesn't say why. I'm assuming that's because JOSDK targets the operator/reconciliation-loop pattern which doesn't fit a synchronous admission webhook — is that right? Just want to confirm that the intent is still to use fabric8 directly for interacting with Kubernetes (e.g. reading `KroxyliciousSidecarConfig` at admission time), rather than introducing a second Kubernetes client library into the project.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,375 @@
...
+
+1. **No config in namespace**: the pod is admitted without injection (debug log only). This is the common case for namespaces where the admin has enabled the namespace label but not yet created a config.
+2. **Multiple configs in namespace**: the pod is admitted without injection (warning logged). The pod can select a specific config via the `kroxylicious.io/sidecar-config` annotation; without this annotation the webhook cannot choose and skips injection.
+3. **Config is invalid in a way the webhook can detect** (e.g. malformed delegated annotation values, plugin image without a digest): the webhook logs a warning and admits the pod without injection. Consistent with fail-open semantics.
+4. **Config is invalid in a way only the proxy can detect** (e.g. unreachable upstream Kafka, wrong TLS trust anchor, non-existent filter type): the webhook injects the sidecar normally. The proxy will fail its startup probe and the pod will not become ready, surfacing the problem via standard Kubernetes health-check mechanisms.
+
```
SamBarker: Cases 1–3 all hardcode `allowed: true` regardless of `failurePolicy` — so an administrator who has configured `failurePolicy: Fail` gets fail-open behaviour on configuration errors anyway. I assume this is deliberate, but it means the failure policy the administrator has chosen has no effect on what actually happens when things go wrong.                                                                  
                            
A more consistent approach: for cases 1–3 the webhook returns HTTP 200 with a descriptive error message, but drives the `allowed` value by reading its own `failurePolicy` from the `MutatingWebhookConfiguration` at startup. When `failurePolicy: Fail`, config errors return `allowed: false` — surfacing a clear message in `kubectl` output and controller events, giving the app owner a visible signal to escalate to the platform team. When `failurePolicy: Ignore`, config errors return `allowed: true`, consistent with the fail-open posture. One knob controls everything uniformly.
                                                                                                                                                                                                                                                                                                                                                                                                                       
The "consistent with fail-open semantics" label on case 3 also looks like a remnant of an earlier default rather than a deliberate statement — worth replacing with an explicit acknowledgement of the chosen policy and the rationale for diverging from `failurePolicy`.                                                             
   
Finally, the proposal should acknowledge that logs are a weak observability mechanism for surfacing configuration errors to the platform owner, and note that better surfacing (e.g. status conditions on  `KroxyliciousSidecarConfig`) is deferred.

SamBarker (commented):
Thanks @tombentley.

I think this is a great start, i've left a couple of inline comments but the main thing between me and approval is this:

There seems to be an implicit convention in the annotation naming: the `sidecar-` prefixed annotations (`sidecar-config`, `sidecar-resources-cpu`, etc.) are those set by humans (namespace admins, app owners) to control behaviour, while the non-prefixed ones (`proxy-config`, `config-generation`, `injection-timestamp`) are written by the webhook itself to record state. If that's intentional it's a useful distinction — but it isn't documented anywhere, and `inject-sidecar` (the pod opt-out label) breaks the pattern slightly since it's human-set but doesn't carry the prefix.
                                                                                                                                                                                                                                                                                                                                                                                                                       
It would be worth making the convention explicit in the proposal. On which note — have you considered using `sidecar.kroxylicious.io/` as the annotation prefix rather than `kroxylicious.io/sidecar-*`? That's the approach Istio takes (`sidecar.istio.io/inject`, `sidecar.istio.io/status`, etc.), where the subdomain carries the feature namespace and the name portion is free to express structure. It would also establish a clean model for future feature areas (`operator.kroxylicious.io/`, etc.) and give a natural place to distinguish human-set control annotations from webhook-written state — e.g. `sidecar.kroxylicious.io/config` vs `sidecar.kroxylicious.io/state-proxy-config`.                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                       
Not a blocker, but alpha is the right time to settle on naming conventions since they'll be hard to change later.

tombentley:
@robobario @SamBarker @k-wall I've tried to address your previous comments, and also bring this into live with the implementation PR. Please take another look.

proposals/100-sidecar-injection-webhook.md
robobario: resolved by updates to the proposal to allow the user to configure the webhook with feature flag, removing any need for RBAC or kube API interactions

proposals/100-sidecar-injection-webhook.md
robobario: correct, we have no reconciliation/operator component yet. We intend to use fabric8 for the same CRD-first development style used in the operator, and for its kube client.

proposals/100-sidecar-injection-webhook.md
robobario: the proposal has been updated with a way to supply secrets, there is no templating, instead secrets are written to a documented location following a documented naming scheme.

I'm happy with that for an initial implementation, it would also be nice to support configmaps but I guess users can initially use Secrets for whatever they need mounted even if they are a little fiddlier to work with.

I think this satisfies our initial requirement to keep secrets out of the annotation so will mark resolved (feel free to re-open @k-wall). We could add templating later.

proposals/100-sidecar-injection-webhook.md
robobario: has been removed from proposal

proposals/100-sidecar-injection-webhook.md
robobario: I think that this is covered now by specifying that initially we don't consider it adversarial between admin and application, and these SASL protections are shunted out of scope

proposals/100-sidecar-injection-webhook.md
robobario: proposal updated, there's now a mechanism for mounting secrets, using documentation to inform the user precisely where they will be located and what the naming scheme will be.

proposals/100-sidecar-injection-webhook.md
robobario: this is covered as a future extension now, not part of alpha. Initially they will have the uglier option to install multiple KroxyliciousSidecarConfigs with differing configuration.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+        pullPolicy: IfNotPresent
+```
+
+The `virtualClusters` list contains per-cluster settings (target bootstrap address, bootstrap port, node ID range, target cluster TLS). The alpha enforces exactly one entry (`minItems: 1`, `maxItems: 1`). Top-level fields (`managementPort`, `proxyImage`, `resources`, `filterDefinitions`, `plugins`, `secretMounts`, `setBootstrapEnvVar`) are shared across all virtual clusters.
+
+**Why a CRD, not a ConfigMap?** Schema validation by the API server, RBAC separation (admin creates, app owners can't modify), status conditions for observability, consistency with the existing Kroxylicious Kubernetes API.
+
+**Why not reuse the operator's CRDs?** The operator CRDs model a shared proxy deployment with ingress networking, multi-cluster support, and cross-resource references. The sidecar use case is fundamentally simpler — localhost binding, no ingress, a single virtual cluster in the alpha. Coupling them would constrain both models.
+
+The webhook admin creates one per namespace. The following edge cases are handled:
```
robobario:
```suggestion
The webhook admin creates one or more per namespace. The following edge cases are handled:
```

Sounds a bit like you can only create one currently, but then the edge cases cover the multi case.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+
+The webhook is packaged as a container image and deployed as a single-replica `Deployment` in a dedicated `kroxylicious-webhook` namespace. Install manifests are provided for:
+
+- Namespace, ServiceAccount, ClusterRole, ClusterRoleBinding
+- Deployment (port 8443)
+- Service (port 443 -> 8443)
+- MutatingWebhookConfiguration
+- cert-manager Certificate (optional)
+
+**TLS**: Kubernetes requires HTTPS for admission webhooks. The primary path uses cert-manager with a self-signed issuer. A manual alternative (admin provides cert/key Secret) is documented. The webhook watches cert files for rotation and reloads the SSLContext.
```
robobario, line 416: should we mention that initially we will only support PEMs with PKCS8 RSA/EC private keys, and whatever the cert format is we can parse.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+### Future delegation
+
+Annotation-based delegation could allow the app owner to override specific sidecar parameters on a per-pod basis, with the admin explicitly opting in via the `KroxyliciousSidecarConfig`. Possible future delegation, ordered roughly by blast radius:
+
+1. **Resource overrides** — app owner adjusts CPU/memory requests and limits. Low risk.
+2. **Filter configuration** — app owner adjusts parameters on admin-selected filters. Medium risk: bounded by the filter's config surface.
+
+### Webhook deployment
+
+The webhook is packaged as a container image and deployed as a single-replica `Deployment` in a dedicated `kroxylicious-webhook` namespace. Install manifests are provided for:
```
robobario: should probably mention multi-replica, PDB we've added for availability

proposals/100-sidecar-injection-webhook.md
robobario: @tombentley this still looks to be the case in the latest proposal, we've updated to a fail-closed default but the edge cases still prescribe that we admit Pods even if there is no configuration or a misconfiguration. I guess they should follow the configured failure mode?

robobario: I guess you're imagining some disaster where our webhook prevents all Pod creations on a namespace and that being a bigger problem than we are trying to solve?

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+
+When the webhook skips injection for a reason other than pod opt-out, it labels the pod with `sidecar.kroxylicious.io/injection-skipped` so that operators can find affected pods without grepping logs. The label value indicates the reason:
+
+| Value | Meaning |
+|-------|---------|
+| `no-config` | No `KroxyliciousSidecarConfig` was found for the pod's namespace |
+| `already-injected` | A container named `kroxylicious-proxy` already exists in the pod |
+| `multiple-configs` | Multiple `KroxyliciousSidecarConfig` resources exist in the namespace and no explicit config was selected via annotation |
+
+Pods that opted out via `sidecar.kroxylicious.io/injection: disabled` are not labelled — they already carry a label that identifies them.
```
SamBarker: Three suggestions on the `injection-skipped` label:

**1. Add `opted-out` for explicitly opted-out pods.**
The `sidecar.kroxylicious.io/injection: disabled` label expresses intent (set by
the app owner); `injection-skipped` is observed state (set by the webhook). They
are semantically different: without an ack label there's no way to distinguish
"the webhook honoured the opt-out" from "the webhook hasn't processed this pod
yet." Adding `injection-skipped=opted-out` also means operators get a single
uniform query for all non-injected pods regardless of reason:

    kubectl get pods -l sidecar.kroxylicious.io/injection-skipped

**2. Revisit the label values for clarity without the webhook's context.**
For an SRE encountering these at 2am:

| Current | Suggested | Rationale |
|---------|-----------|-----------|
| `no-config` | `no-sidecar-config` | Makes clear it's the `KroxyliciousSidecarConfig` CR that's absent, not some internal webhook config |
| `multiple-configs` | `ambiguous-sidecar-config` | "Multiple" describes the cause; "ambiguous" describes the problem the operator needs to resolve |
| `already-injected` | `container-name-conflict` | The current name implies a previous injection pass; the actual cause is a pre-existing container with a conflicting name |

**3. Reinvocation should be silent.**
The configuration drift detection section describes a second skip path: the
webhook skips injection if the `sidecar.kroxylicious.io/config-generation`
annotation is already present (idempotency guard on reinvocation). The current
text implies this would set `injection-skipped=already-injected`, but reinvocation
is expected behaviour — adding the label would pollute the signal for operators
querying `-l sidecar.kroxylicious.io/injection-skipped`. Reinvocation should
produce no label.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+
+#### Annotation-based idempotency (future)
+
+The current idempotency check (container name) is weak: it can be circumvented by a pod owner pre-adding a container named `kroxylicious-proxy`, and it does not detect configuration drift during reinvocation. A stronger approach is to compare the `sidecar.kroxylicious.io/proxy-config` annotation on the pod against the configuration the webhook would generate right now. If the annotation is absent or differs, the webhook injects (or re-injects); if it matches, injection is skipped.
+
+This works because `ProxyConfigGenerator.generateConfig()` is deterministic: given the same `KroxyliciousSidecarConfigSpec` inputs, Jackson serialisation produces identical YAML. String equality on the annotation value is therefore a reliable idempotency check within the same webhook version. This also handles the reinvocation case (`reinvocationPolicy: IfNeeded`): if another mutating webhook modifies the pod after initial injection, the webhook can verify that the proxy config annotation is still correct.
```
SamBarker: This paragraph is back-to-front: it cites the current implementation to justify
the design rather than stating the property the design requires. The proposal
should describe the requirement, and leave it to the implementation to satisfy it.

Something like: for annotation comparison to be a reliable idempotency check,
config generation must be deterministic — the same `KroxyliciousSidecarConfigSpec`
must always produce byte-identical serialized output within a given webhook
version. The implementation must guarantee this; how it does so is not a concern
of this proposal.

The version caveat also deserves calling out explicitly as a constraint: a webhook
upgrade that changes serialization output means existing pods' annotations no
longer match what the webhook would generate, making them appear as configuration
drift even though the `KroxyliciousSidecarConfig` itself hasn't changed. The
proposal should state this consequence — and note that remediation requires pods
to cycle, since nothing proactively re-injects running pods.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+| `observedGeneration` | `int64` | The `metadata.generation` most recently observed by the webhook |
+| `conditions` | `[]Condition` | Standard Kubernetes conditions (see below) |
+
+The webhook maintains a single condition type:
+
+| Condition | Meaning |
+|-----------|---------|
+| `Ready` | The webhook has observed and accepted this configuration. `observedGeneration` on the condition tracks which generation was acknowledged. |
+
+The webhook sets `Ready=True` (reason `Accepted`) when it first observes the config via its informer. The condition is only updated when the generation changes, avoiding unnecessary status writes. Status update failures (e.g. conflicts) should be retried; persistent failures should be surfaced via logging so that operators can investigate. A status update failure does not block pod admission, but a `KroxyliciousSidecarConfig` with a stale or missing `Ready` condition indicates a problem that needs attention.
```
SamBarker: The proposal only defines `Ready=True`. If the webhook observes a config and can
determine it is invalid (e.g. missing required fields, cases 1–3 from the edge
cases section), setting `Ready=False` with a descriptive reason and message would
surface the problem directly on the CRD — where an operator would naturally look
— rather than requiring them to find the relevant log line. As it stands, a
misconfigured `KroxyliciousSidecarConfig` is either acknowledged (`Ready=True`)
or invisible; there is no condition state that says "I have seen this and it is
wrong."

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+| `managementPort` (default 9082) | Management (`/livez`, `/metrics`) | `0.0.0.0` |
+
+The webhook sets `KAFKA_BOOTSTRAP_SERVERS=localhost:<bootstrapPort>` on application containers (configurable via `setBootstrapEnvVar`, defaults to `true`).
+
+The management endpoint binds to `0.0.0.0` because kubelet HTTP probes target the pod IP, not loopback. This means the application container can also reach `/livez` and `/metrics`, but neither endpoint exposes sensitive data.
+
+### Native sidecar containers
+
+On Kubernetes 1.29+ (where the `SidecarContainers` feature gate is enabled by default), the webhook injects the proxy as a native sidecar — an init container with `restartPolicy: Always`. This gives proper startup ordering (proxy starts before the application) and shutdown ordering (proxy stops after the application). On older clusters, the webhook falls back to injecting into `spec.containers`.
+
```
SamBarker, line 248: `FEATURE_GATES` is ambiguous — it reads as though it controls the webhook's own
internal feature flags rather than describing which Kubernetes API server feature
gates are active on the cluster. `K8S_FEATURE_GATES` would be clearer about the
subject.

It would also be worth stating explicitly that this is an escape hatch for when
version-based detection is insufficient, not the recommended default. Deployers
who set it are taking on the responsibility of keeping it in sync with their
cluster configuration — the proposal should acknowledge that risk.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+### Sidecar container spec
+
+The injected sidecar follows the same patterns as `ProxyDeploymentDependentResource` in the operator:
+
+- Container-level `securityContext`: `allowPrivilegeEscalation: false`, `capabilities: drop ALL`, `readOnlyRootFilesystem: true`
+- Probes: `startupProbe` (initialDelay 5s, period 2s, failure threshold 30), `livenessProbe` (initialDelay 30s, period 10s, failure threshold 3), `readinessProbe` (initialDelay 5s, period 2s, failure threshold 5) — all HTTP GET `/livez` on the management port (default 9082)
+- `terminationMessagePolicy: FallbackToLogsOnError`
+
+The webhook does not set a pod-level `securityContext`. Pod-level security policies (e.g. `runAsNonRoot`, `seccompProfile`) are the responsibility of the cluster admin via Kubernetes `PodSecurity` admission (`pod-security.kubernetes.io/enforce: restricted` namespace label) or equivalent policy enforcement. Setting pod-level security context from a mutating webhook risks ordering conflicts with other webhooks.
+
```
SamBarker, line 260: The proposal only addresses one direction: the sidecar's security context is never
weakened below what the webhook would set. It doesn't address the reverse — a pod
with a *weaker* pod-level security context (e.g. `seccompProfile: Unconfined` to
support io_uring). Since the webhook doesn't set a container-level seccomp profile
on the sidecar, the sidecar would silently inherit a permissive pod-level one.

Not a blocker for alpha, but worth keeping in mind: the proxy itself is a candidate
for io_uring in the future (for performance), so we may actually want a way to
configure the sidecar's seccomp profile explicitly rather than inheriting from the
pod. The `KroxyliciousSidecarConfig` spec is the natural place for that.

proposals/100-sidecar-injection-webhook.md
tombentley: On (1): opted-out pods are excluded by `objectSelector` on the `MutatingWebhookConfiguration`; they never reach the webhook so it can't label them. Changing the exclusion mechanism (e.g. moving opt-out from `objectSelector` to in-webhook logic) would mean every pod creation in the namespace hits the webhook regardless. I think the current approach is the right trade-off.

On (2): Yes, we can use better values; I'll revise.

On (3): You're right, reinvocation is expected behaviour and shouldn't set the label. I'll make that explicit.

proposals/100-sidecar-injection-webhook.md
tombentley: I'll rewrite this to state the determinism requirement without referencing implementation details, and add an explicit note about the webhook-upgrade consequence: changed serialisation output means existing pods appear as drift, requiring pod cycling to remediate.

proposals/100-sidecar-injection-webhook.md
tombentley: Good point. I'll add Ready=False (reason Invalid, with a descriptive message) for configs that the webhook can determine are invalid at observation time. This naturally covers structural problems detectable by the webhook's own validation beyond what the CRD schema catches (e.g. referencing a non-existent Secret).

For the per-namespace edge cases (no-config, multiple-configs): these aren't properties of a single CRD, so they don't fit naturally into CRD status. They're already covered by the injection-skipped label on affected pods. I think that's the right place for those.

proposals/100-sidecar-injection-webhook.md
tombentley: I'll rename to `K8S_FEATURE_GATES` and add a note that this is an escape hatch for clusters where version-based detection is insufficient, with deployers responsible for keeping it in sync.

proposals/100-sidecar-injection-webhook.md
tombentley: Good catch. For alpha, I think we should set seccompProfile: RuntimeDefault explicitly on the sidecar container, so it doesn't inherit a permissive pod-level profile. Container-level seccomp overrides pod-level, so this won't affect the app container.

Agreed that KroxyliciousSidecarConfig is the right place for explicit seccomp configuration if/when we need it (e.g. for io_uring). I'll add a note about that as a future consideration.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+  filterDefinitions:
+    - name: envelope-encryption
+      type: io.kroxylicious.filter.encryption.EnvelopeEncryptionFilterFactory
+      config:
+        credentialsFile: /opt/kroxylicious/secrets/kms/credentials.json
+```
+
+The mount path is derived automatically from the `name` field — the admin does not specify `mountPath` directly. This keeps the sidecar's filesystem layout under webhook control, consistent with the operator's approach of using fixed base paths for secret volumes.
+
+**Why `secretMounts` over generic `sidecarVolumes`/`sidecarVolumeMounts`?** It signals clear intent (admin-controlled secrets for the sidecar), limits the surface to Secrets rather than arbitrary volumes, and is easier to validate. The field can be generalised later without breaking the existing API.
```
SamBarker, line 289: `secretMounts[].secretName` is a cross-resource reference — without validating
that the referenced Secret exists in the pod's namespace, a
`KroxyliciousSidecarConfig` with a bad reference passes admission silently and
only fails when Kubernetes tries to mount the volume at scheduling time. This
validation needs to be part of the alpha implementation, not deferred. It is also
exactly the case where `Ready=False` would be most valuable — see separate
comment on § Status.

The operator has validation logic for `LocalObjectReference` fields that handles
this; the webhook should follow the same pattern.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+
+**Remaining risks:**
+
+- **Supply chain**: a compromised plugin image contains malicious code with access to Kafka traffic and mounted credentials. Mitigate with image signing and digest-pinned references.
+- **Dependency conflicts**: as described above under flat classpath limitations.
+
+### Virtual clusters
+
+The `virtualClusters` list defines the target Kafka clusters that the sidecar proxy will serve. Each entry specifies a name, target bootstrap address, localhost listening port, node ID range, and optional TLS configuration.
+
```
SamBarker, line 396: Really like the forward-looking list structure here. Active/passive DR is a
compelling use case — the sidecar exposes both live and DR clusters, and failover
is a config change plus pod restart. Could be a great showcase for the project.

proposals/100-sidecar-injection-webhook.md
SamBarker: Good point on (1) — I hadn't appreciated that `objectSelector` excludes pods before they reach the webhook entirely, making the ack label impossible without changing the exclusion mechanism. Happy to go with the simplest solution.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,448 @@
...
+## Affected/not affected projects
+
+| Project | Affected | Nature of change |
+|---------|----------|-----------------|
+| `kroxylicious-kubernetes/kroxylicious-admission` | Yes | New module: webhook implementation |
+| `kroxylicious-kubernetes/kroxylicious-admission-api` | Yes | New module: `KroxyliciousSidecarConfig` CRD |
+| `kroxylicious-app` | Already merged | `classpath-plugins/` directory scanning |
+| `kroxylicious-operator` | No | |
+| `kroxylicious-runtime` | No | Used as a dependency, not modified |
+| `kroxylicious-filters` | No | |
```
SamBarker, line 444: Two permissions are missing from this list:

- `get` (or `list`/`watch` via informer) on `Secrets` — needed to validate `secretMounts[].secretName` references at config-observation time.
- `update` on `KroxyliciousSidecarConfig/status` — needed to write the `Ready` condition described in § Status.

proposals/100-sidecar-injection-webhook.md
tombentley: > `secretMounts[].secretName` is a cross-resource reference — without validating that the referenced Secret exists in the pod's namespace, a `KroxyliciousSidecarConfig` with a bad reference passes admission silently and only fails when Kubernetes tries to mount the volume at scheduling time. 

True.

> This validation needs to be part of the alpha implementation, not deferred.

I don't think that's necessary -- if we were to add this later it would not impact the API or be a breaking change. So I think 'need' is too strong.

> It is also exactly the case where `Ready=False` would be most valuable — see separate comment on § Status.
> 
> The operator has validation logic for `LocalObjectReference` fields that handles this; the webhook should follow the same pattern.

While I agree the concern is valid, but let's consider the implications of doing this:

1. The webhook currently needs zero `Secret` access. Adding validation means `get` on `Secrets` in every namespace where a sidecar config might exist. A webhook service account that can read arbitrary `Secrets` is a high-value target, and turns a compromise of the webhook into a credential exfiltration vector. This directly undermines the "minimal RBAC" design the proposal currently has. So Expecting a kube admin to grant this unknown new webhook access to all the secrets in all the namespaces seems to be asking a heck of a lot.
2. While it provides a way to notify the person configuring the sidecar config at the time the sidecar config is created, there's nothing stopping the secret from being deleted afterwards. It's at the point when the pod's created that the secret needs to exist. But under the two actor design there's no guarantee that the side car admin is paying attention at the point the pod is created by the app owner.    
3. The operator analogy is misleading. The operator has a controller that continuously reconciles, so it can re-check references and update status. The webhook fires once at pod admission. If it sets `Ready=False` because a `Secret` is missing, when does it set `Ready=True`? Only options: watch all Secrets (enormous scope, back to the RBAC problem) or only update on the next pod admission (leaving the status stale and misleading for potentially hours or days).
4. A pod referencing a nonexistent `Secret` via a volume mount won't schedule. `kubectl describe pod` shows exactly what's missing. This is a standard failure mode every Kubernetes operator knows how to debug. Adding a second layer of validation doesn't prevent the problem, it just moves the error message to a different place, while adding complexity and RBAC surface.
5. If you validate `secretMounts`, logical consistency demands validating `trustAnchorSecretRef` too, and any future secret references. This sets a precedent where the webhook must validate all cross-resource references, compounding the RBAC and complexity costs.
6. In progressive delivery workflows (Argo CD, Flux), the `KroxyliciousSidecarConfig` and the `Secret` may be in different apps or kustomizations and deploy in arbitrary order. Rejecting the config because the Secret doesn't exist yet breaks these workflows.

SamBarker: Thanks for the detailed response. The RBAC concern and the progressive delivery point are well taken — I accept that validating at config-creation time is genuinely problematic.

But I want to push back on the framing that `kubectl describe pod` is an adequate UX story here. This isn't injection being *skipped* — the sidecar gets injected, the pod fails to start, and the only signal is a volume mount error buried in pod events. The `KroxyliciousSidecarConfig` shows `Ready=True`. There's no trail back to the config.

This is also the case you named as the prime example of where `Ready=False` would be valuable ("e.g. referencing a non-existent Secret"). If that case can't be covered, the `Ready=False` story is narrower than it appeared.

I'm not insisting on fixing this in alpha — the reconciliation problem is real. But I think the proposal should explicitly acknowledge this as a known UX gap rather than leaving it implicit, and note that a future reconciliation controller is the right fix.

SamBarker: Sketching a future enhancement that addresses the UX gap without changing the webhook's minimal-RBAC design:

**Operator sidecar reconciler (future)**

The operator already has the RBAC, reconciliation loop, and CRD machinery that the webhook deliberately avoids. A future operator enhancement could add a reconciler that watches `KroxyliciousSidecarConfig` resources cluster-wide and provides what the webhook can't:

1. **Referential integrity**: watch referenced Secrets and TLS trust anchor Secrets. Set `Ready=False` (reason: `MissingSecret`, with the name) when a reference is broken; flip back to `Ready=True` when resolved. This handles both the initial-creation case and the "Secret deleted after the fact" case that a one-shot webhook check can't catch.

2. **Drift surfacing**: watch pods with `sidecar.kroxylicious.io/config-generation` annotations. Compare against the current `KroxyliciousSidecarConfig` generation and surface an `UpToDate` condition on the CRD status — "3 pods running stale config" — without requiring manual `jq` queries.

3. **Optional auto-roll** (opt-in): if `KroxyliciousSidecarConfig` carries a `sidecar.kroxylicious.io/auto-roll: enabled` annotation, the reconciler triggers a rolling restart of stale pods by patching the owning `Deployment`/`StatefulSet` pod template annotation (the standard `kubectl rollout restart` mechanism). Opt-in so teams that manage their own rollouts aren't surprised.

The webhook stays unchanged — inject, set `Ready=True`, done. The operator adds observability and optionally remediation on top. Neither component needs to compromise its design to support the other.

This is the right vehicle for the "future reconciler" the proposal already gestures at in § Configuration drift detection.

tombentley: > the sidecar gets injected, the pod fails to start, and the only signal is a volume mount error buried in pod events

But that is something that anyone familiar with kube will know how to deal with.

> There's no trail back to the config.

That's not really true. If you describe the pod you'll see the secret volume, the volume mount, that the volume mount is in an injected sidecar. I think that's _enough_ for a competant person or AI to figure out that they need to involve the kube administrator (if they don't directly know about the kroxylicious sidecar already). It's not exemplary UX, but it's enough.

> If that case can't be covered, the Ready=False story is narrower than it appeared.

I get your disappointment. I'd also like to have a better UX here. But that just doesn't seem to fit within the role of a webhook, which is what this proposal is about.

> But I think the proposal should explicitly acknowledge this as a known UX gap rather than leaving it implicit, 

I'm happy to do that.

> note that a future reconciliation controller is the right fix.

I don't think either of us really know the extent to which this is a real problem for people in practice. So I'd prefer to avoid making something that looks like a commitment to future work in this proposal.

> **Operator sidecar reconciler (future)**

We've been following the path that the admission webhook is strictly orthogonal to the current operator. That's because :

* we think there will be people who are interested in one but not the other, because
* they're fundamentally different deployment styles for using a proxy, with different pros and cons.

So I think deciding to extend the scope of the existing operator to include the sidecar CR is premature. 

Sure, it's one option. Another option is to have a dedicated operator for sidecar configs. Another option is to do nothing. How we weigh these options depends on feedback we get from users.

proposals/100-sidecar-injection-webhook.md
robobario: On reflection, the proposal doesn't need to go down that low a level on key/cert formats. We've alreay declared we need TLS and prefer cert-man integration

proposals/100-sidecar-injection-webhook.md
robobario: On discussion offline I now see the distinction. There are two places where we need to think about an admission policy:

1. Infrastructure Failure. When the webhook is unavailable or is unable to cleanly make a decision (unhandled exceptions etc)
2. Logical Exclusion. When the webhook is available, and targeting a namespace, and the environment is in a predictable state, but we do not want to inject the pod due to missing configuration, misconfiguration, etc.

For 1. we can rely on the kubernetes webhook failure policy, which already implements a well-known way to control admission when the webhook is unavailable or returning error responses. Fail-closed is a good policy here as the webhook is totally out of the loop or non-functional, admitting the pod would silently avoid all webhook logic.

For 2., when the webhook decides not to inject due to it's own logic/rules, then we need a policy about whether to admit or deny. At this point fail open is more tolerable because the webhook is in control and able to stamp pod with observable evidence pods have been intercepted by the webhook and injection has been skipped.

tombentley: @robobario yes, that's what's now implemented in the feature branch. I think your point 1 is described in this proposal on line 56 and your point 2 is addressed by line 58.

proposals/100-sidecar-injection-webhook.md
tombentley: Just pushed a fix for this.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,469 @@
...
+4. **Config is invalid in a way only the proxy can detect** (e.g. unreachable target Kafka cluster, wrong TLS trust anchor, non-existent filter type): the webhook injects the sidecar normally. The proxy will fail its startup probe and the pod will not become ready, surfacing the problem via standard Kubernetes health-check mechanisms.
+
+When the webhook skips injection for a reason other than pod opt-out, it labels the pod with `sidecar.kroxylicious.io/injection-skipped` so that operators can find affected pods without grepping logs. The label value indicates the reason:
+
+| Value | Meaning |
+|-------|---------|
+| `no-KroxyliciousSidecarConfig` | No `KroxyliciousSidecarConfig` was found for the pod's namespace |
+| `ambiguous-KroxyliciousSidecarConfig` | Multiple `KroxyliciousSidecarConfig` resources exist in the namespace and no explicit config was selected via annotation |
+| `container-name-conflict` | A container named `kroxylicious-proxy` already exists in the pod |
```
SamBarker, line 135: The `injection-skipped` label table covers three values, but case 3 from the edge-case list — "config is invalid in a way the webhook can detect" — is missing. Under `UNINJECTED_POD_POLICY=Admit` that case also results in skipped injection, so the label should be set. An operator running:

    kubectl get pods -l sidecar.kroxylicious.io/injection-skipped

would find those pods with no machine-readable way to distinguish "config invalid" from "no config" or "ambiguous config".

Suggest adding an `invalid-KroxyliciousSidecarConfig` entry to the table.

proposals/100-sidecar-injection-webhook.md
```diff
@@ -0,0 +1,469 @@
...
+
+```
+kubectl get pods -n my-ns -o json | jq '[.items[] |
+  select(.metadata.annotations["sidecar.kroxylicious.io/config-generation"] != null) |
+  {name: .metadata.name, generation: .metadata.annotations["sidecar.kroxylicious.io/config-generation"]}]'
+```
+
+In a future iteration, a reconciler could watch for pods with outdated generations and surface an `UpToDate` condition on the `KroxyliciousSidecarConfig` status, giving operators visibility into configuration drift without requiring manual queries.
+
+#### Annotation-based idempotency (future)
```
SamBarker, line 173: Two statements read as contradictory:

- Lines 156–158 describe the `config-generation` annotation presence as a current idempotency guard: "its presence indicates that the sidecar has already been injected, preventing re-injection when the webhook is reinvoked."
- Line 173's section heading "Annotation-based idempotency (future)" calls the container name check "the current idempotency check", implying the annotation guard is not yet in place.

I suspect things have evolved while this proposal was in progress and the text hasn't quite kept up. Could you check these distinctions are still accurate and tighten the section heading at 173 if the generation guard is already in place?

SamBarker (approved):
LGTM

A couple of housekeeping niggles to resolve but this represents a substantial improvement to the project. Thanks for all the hard work her @tombentley & @robobario

robobario (approved):
LGTM, thanks @tombentley

PR #101, opened 2026-04-17
New routing strategy: proxyProtocolAuthorityIdentifiesNode

hrishabhg:
---

## ⚠️ Proposal Numbering

This PR contains proposal files that don't follow the expected format:

**Issues found:**

- <code>proposals/101-proxy-protocol-authority-routing.md</code>

**Fix for <code>proposals/101-proxy-protocol-authority-routing.md</code>:**

```bash
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 101 - /; t; s/^# /# 101 - /}' proposals/101-proxy-protocol-authority-routing.md && rm proposals/101-proxy-protocol-authority-routing.md.bak
git add proposals/101-proxy-protocol-authority-routing.md
git commit -m "Rename proposal to use PR number 101"
git push
```

**Note:** The commands above can be copied and pasted directly into your terminal. The sed command will automatically preserve your title text.

See [proposals/README.md](https://github.com/kroxylicious/design/blob/main/proposals/README.md) for the complete workflow.


proposals/017-proxy-protocol-authority-routing.md
```diff
@@ -0,0 +1,165 @@
...
+This strategy builds on the existing PROXY protocol support. When `proxyProtocolAuthorityIdentifiesNode` is configured:
+
+- `proxyProtocol.mode` is implicitly `required` for gateways using this strategy.
+- The existing `HaProxyContext` TLV capture is extended to make the authority value available to the routing layer.
+
+### Validation
+
+The following conditions should be validated at configuration time:
+
+- `proxyProtocol.mode` must not be `disabled` (or it is set to `required` implicitly).
```
tombentley: I think there's an argument to force the configuration to be explicit. I.e. it must be `proxyProtocol.mode` must be present and have value `required`. I know that's more verbose, but it means there's less possibility for a reader to misunderstand what's going on.

tombentley (approved):
@hrishabhg thanks for this. The use case makes sense, and I think we should do this. 

I would like to see some words calling out that when terminating TLS in NLB/HAProxy and using plain TCP for the connection to the Kroxylicious proxy instance(s) there is an assumption that the intermediate network is trusted. It's obvious, but should not be overlooked by people deploying the proxy, so we'll need to ensure that we have similar words in the docs too.

proposals/017-proxy-protocol-authority-routing.md
```diff
@@ -0,0 +1,165 @@
...
+```yaml
+virtualClusters:
+  - name: demo
+    targetCluster:
+      bootstrapServers: upstream-kafka:9092
+    gateways:
+      - name: via-nlb
+        proxyProtocolAuthorityIdentifiesNode:
+          bootstrapAddress: 0.0.0.0:9192
+          advertisedBrokerAddressPattern: broker-$(nodeId).kafka.example.com
```
robobario: Would we need to support advertised port override the same as the SNI strategy? Like supporting:

```
        proxyProtocolAuthorityIdentifiesNode:
          bootstrapAddress: 0.0.0.0:9192
          advertisedBrokerAddressPattern: broker-$(nodeId).kafka.example.com:442
```
so that the Proxy binds to 9192, but can offer 442 in the Metadata response in case the loadbalancer uses a different port?

proposals/017-proxy-protocol-authority-routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+```yaml
+virtualClusters:
+  - name: demo
+    targetCluster:
+      bootstrapServers: upstream-kafka:9092
+    gateways:
+      - name: via-nlb
+        proxyProtocolAuthorityIdentifiesNode:
+          bootstrapAddress: 0.0.0.0:9192
```
robobario: For the other strategies we support a convenience bootstrapAddress like:
```
  sniHostIdentifiesNode:
        bootstrapAddress: bootstrap.mycluster.example.com:9192 
        advertisedBrokerAddressPattern: mybroker-$(nodeId).mycluster.example.com:443
```
Kafka does not require this, but it can make client configuration more convenient.

Do we also want to support a convenience bootstrapAddress in this strategy? So that users might connect to the NLB with `bootstrap.kafka.example.com` and the Proxy can identify which gateway the incoming connection is associated with?

robobario: Can multiple proxyProtocolAuthorityIdentifiesNode gateways share a port the same way we do for SNI? I imagine that they can.

robobario: ah, yes in your POC example you have:

```
Kafka clients connecting via TLS to HAProxy with bootstrap server `bootstrap.kafka.example.com:49192` produced the following in Kroxylicious logs:
```

so I think the proxy will need to be aware of that host so it can route it to the right place. Maybe something like:

```
- name: via-nlb
        proxyProtocolAuthorityIdentifiesNode:
        bindAddress: 0.0.0.0:9192
        bootstrapHostname: bootstrap.kafka.example.com
        advertisedBrokerAddressPattern: broker-$(nodeId).kafka.example.com
```

proposals/017-proxy-protocol-authority-routing.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+The test setup (HAProxy + docker-compose + cert generation) is available in `dev/envoy-proxy-protocol/` of the Kroxylicious repository.
+
+## Affected/not affected projects
+
+| Project | Affected | Notes |
+|---------|----------|-------|
+| kroxylicious-proxy | Yes | New routing strategy implementation, extends `HaProxyContext` to expose authority TLV to routing layer |
+| kroxylicious-operator | Yes | CRD update to support the new gateway type |
```
robobario: Maybe we should set this to not affected and cover the operator in a separate proposal, going into detail here will add complexity to this design.

k-wall:
Discussed at Community Call 23rd April.  General feeling is that this is good idea.
@hrishabhg  is looking for feedback.

k-wall:

```bash
git mv proposals/017-proxy-protocol-authority-routing.md proposals/101-protocol-authority-routing.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 101 - /; t; s/^# /# 101 - /}' proposals/101-protocol-authority-routing.md && rm proposals/101-protocol-authority-routing.md.bak
git add proposals/101-protocol-authority-routing.md
git push
```

k-wall:

```bash
git mv proposals/017-proxy-protocol-authority-routing.md proposals/101-proxy-protocol-authority-routing.md
sed -i.bak '0,/^# /{s/^# \([0-9]\{3\}\|xxx\|nnn\|000\) - /# 101 - /; t; s/^# /# 101 - /}' proposals/101-proxy-protocol-authority-routing.md && rm proposals/101-proxy-protocol-authority-routing.md.bak
git add proposals/101-proxy-protocol-authority-routing.md
git push
```

Sorry for the confusion!

PR #104, opened 2026-04-24
Implement PR-number-based proposal workflow

k-wall:
This PR implements Rob's suggestion from #95 to use PR numbers as proposal identifiers.

## Changes

- **Simplified proposals/README.md**: Removed index tables (directory listing serves as the index), streamlined author workflow
- **Updated proposals/000-template.md**: Added workflow instructions
- **Added GitHub workflow**: Automatically checks proposal numbering
  - Updates PR description when files don't match PR number
  - Provides exact commands to fix naming
  - Removes warning once corrected
- **Added notify-open-prs.sh**: One-time script to notify existing PR authors to rebase

## Proposal Numbering

Proposals now use PR numbers as identifiers (e.g., PR #105 → proposals/105-my-proposal.md).

**Why PR numbers?**
- Natural chronological ordering
- No number collisions (GitHub guarantees unique PR numbers)
- GitHub references (`#105`) work automatically
- Simpler workflow (no separate allocation PR needed)

## Migration for Existing Proposals

After merge, run `./notify-open-prs.sh` to notify authors of open proposal PRs. When they rebase, the workflow will automatically guide them through any needed renaming.

Proposals 001-019 retain their original numbers.

---

Separately, [kroxylicious#3794](https://github.com/kroxylicious/kroxylicious/pull/3794) has been raised to ensure that a Claude session running the main repo finds the proposals.

**Note:** This PR replaces #95 with a cleaner implementation focused on automation rather than manual index maintenance.

### DIFF
--- a/.github/workflows/proposal-numbering.yml
+++ b/.github/workflows/proposal-numbering.yml
@@ -1,0 +1,209 @@
+#
+# Copyright Kroxylicious Authors.
+#
+# Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+#
+
+# Automatically checks that proposal files follow the PR-number naming convention (e.g., proposals/092-my-proposal.md for PR #92).
+# When a proposal file doesn't match its PR number, this workflow updates the PR description with instructions to rename it.
+# Once the author fixes the naming, the warning is automatically removed from the PR description.
+
+name: Check Proposal Numbering
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  check-numbering:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          fetch-depth: 0
+
+      - name: Check proposal file numbering
+        id: check
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          # Find all .md files directly in proposals directory (not subdirectories)
+          # Include both Added (A) and Renamed (R) files
+          ALL_PROPOSAL_FILES=$(git diff --name-only --diff-filter=AR origin/${{ github.base_ref }}...HEAD | grep '^proposals/[^/]*\.md$' || true)
+
+          # Filter out non-proposal files (README.md and template)
+          PROPOSAL_FILES=""
+          for file in $ALL_PROPOSAL_FILES; do
+            # Skip README.md and the template
+            if [[ "$file" == "proposals/README.md" ]] || [[ "$file" == "proposals/000-template.md" ]]; then
+              continue
+            fi
+            PROPOSAL_FILES="$PROPOSAL_FILES $file"
+          done
+
+          if [ -z "$PROPOSAL_FILES" ]; then
+            echo "No proposal files found in this PR"
+            echo "has_proposal=false" >> $GITHUB_OUTPUT
+            exit 0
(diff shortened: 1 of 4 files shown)


proposals/README.md
```diff
@@ -0,0 +1,38 @@
+# Kroxylicious Proposals
+
+This directory contains proposals for the Kroxylicious project.
+
+## Creating a New Proposal
+
+1. **Create your proposal file** using a placeholder name based on the [template](./000-template.md):
+   ```
+   proposals/000-<descriptive-name>.md
```
robobario:
```suggestion
   cp proposals/000-template.md proposals/000-<descriptive-name>.md
```

.github/workflows/proposal-numbering.yml
```diff
@@ -0,0 +1,171 @@
...
+              'This PR contains proposal files that don\'t follow the expected numbering scheme:',
+              '',
+              filesList,
+              '',
+              `**Expected naming:** <code>${expectedPrefix}&lt;descriptive-name&gt;.md</code>`,
+              '',
+              `Please rename your proposal file to use PR #${prNumber}:`,
+              '',
+              '```bash',
+              `git mv proposals/000-<name>.md ${expectedPrefix}<name>.md`,
```
robobario:
```suggestion
              `git mv proposals/000-<name>.md proposals/${expectedPrefix}-<name>.md`,
```

.github/workflows/proposal-numbering.yml
```diff
@@ -0,0 +1,171 @@
...
+# When a proposal file doesn't match its PR number, this workflow updates the PR description with instructions to rename it.
+# Once the author fixes the naming, the warning is automatically removed from the PR description.
+
+name: Check Proposal Numbering
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'proposals/*.md'
```
robobario: Are we able to make this a required check in the branch rules if its conditional? I think it will be a no-op if we run it all the time for PRs with no `proposals/*` changes.

.github/workflows/proposal-numbering.yml
```diff
@@ -0,0 +1,171 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Check proposal file numbering
+        id: check
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          # Find all .md files directly in proposals directory (not subdirectories)
+          ALL_PROPOSAL_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep '^proposals/[^/]*\.md$' || true)
```
robobario:
```suggestion
          ALL_PROPOSAL_FILES=$(git diff --name-only --diff-filter=A origin/${{ github.base_ref }}...HEAD | grep '^proposals/[^/]*\.md$' || true)
```

Maybe we should filter to additions only, anticipating follow-up PRs modifying established proposal fires which I think would fail the automation.

robobario (approved):
LGTM. Had a play around with it on my fork and it updated/cleaned the PR description as expected.

I wonder if we should require the number to be mentioned in the proposal too, like the title should be `XYZ - My Description Title`, or include a PR link so that users can nav to the PR easily if they are browsing the merged files.

One thing I found is that PRs aren't google-able, this seems to be pretty standard, github offers up headers/robots to make sure PRs aren't crawled/indexed since they are in flux. So unfortunately if you Google "Kroxylicious Hot Reload" nothing useful comes up. Maybe we could add automation to automatically create an Issue for the new PR since Issues are indexed.

.github/workflows/proposal-numbering.yml
k-wall: Good idea.

PR #107, opened 2026-04-28
fix(actions): proposal numbering body update failed

robobario:
PRs from forks do not have permissions to write the PR

https://github.com/kroxylicious/design/actions/runs/25076591595

Switching to pull_request_target will enable the write Note that this is safe for the design repo as we don't execute any code from the PR.


robobario:
tested on my fork, I PRed from the main repo and it updated the PR body https://github.com/robobario/design/pull/2

SamBarker (approved):
LGTM

PR #110, opened 2026-05-14
110 - Adopt plugin versioning scheme

robobario:
This is a smaller idea for adopting plugin versioning without tackling all of the other ideas like contract-first Configuration based on JSON schemas, kubernetes sympathy by decomposing into multiple configuration files which are covered in #96 . It only makes versioning explicit and adopts k8s CRD versioning conventions to try and make our stability guarantees explicit.

Proposal generated with the help of Claude Sonnet 4.5

### DIFF
--- a/proposals/110-versioned-plugin-configuration.md
+++ b/proposals/110-versioned-plugin-configuration.md
@@ -1,90 +1,165 @@
-# 110 - Versioned Plugin Configuration
+# 110 - Plugin Configuration Versioning
 
-This proposal introduces explicit API versioning for plugins through a new `@Version` annotation and changes to the plugin type resolution mechanism.
-The changes enable plugin implementations to declare their API stability level independently of the project's semantic version and allow users to explicitly reference plugin versions in configuration.
-Version information also serves as a disambiguation mechanism when multiple plugin implementations share the same simple class name.
+This proposal introduces explicit configuration schema versioning for plugins by extending the existing `@Plugin` annotation with a `configVersion` attribute.
+Plugin implementations can declare support for multiple configuration schema versions simultaneously, enabling controlled API evolution and migration periods.
+Users explicitly specify which configuration schema version they are using via a `version` field in their YAML configuration.
+This establishes a foundation for Kubernetes-style API versioning (v1alpha1, v1beta1, v1) that will be fully realized in PR #96's multi-file configuration system.
 
 ## Current situation
 
 Plugins are referenced in configuration using either their simple class name or fully qualified name.
-The proxy resolves these references at startup through ServiceLoader discovery, registering each plugin under both its fully qualified name and simple name.
-When multiple plugins share the same simple name, the proxy logs an ambiguity warning and requires users to reference the plugin by its fully qualified name.
-Plugin API stability is coupled to the project's semantic version, making it difficult to ship experimental or unstable plugin APIs alongside stable ones as the project approaches 1.0.0.
+The proxy resolves these references at startup through ServiceLoader discovery.
+Each plugin has a single configuration schema tied to its `@Plugin(configType = ...)` annotation.
+When a plugin's configuration needs to evolve, the plugin author must choose between breaking existing users or maintaining backward compatibility through complex conditional logic in a single configuration class.
+Plugin API stability is implicitly coupled to the project's semantic version, making it difficult to ship experimental or unstable plugin configuration schemas alongside stable ones as the project approaches 1.0.0.
 
 ## Motivation
 
 Kroxylicious is approaching its 1.0.0 release, which will establish backward compatibility guarantees for public APIs.
-However, different plugins may have different maturity levels and API stability guarantees.
-Coupling all plugin APIs to a single project version creates tension between shipping experimental features and maintaining stability.
-The Record Encryption filter, for example, may need multiple major revisions before its API stabilizes, but tying those revisions to the project version would either delay 1.0.0 or force premature API commitments.
+However, different plugins may have different maturity levels and configuration schema stability guarantees.
+Coupling all plugin configuration schemas to a single project version creates tension between shipping experimental features and maintaining stability.
+The Record Encryption filter, for example, may need multiple configuration schema revisions before it stabilizes, but tying those revisions to the project version would either delay 1.0.0 or force premature API commitments.
 
-Kubernetes-style API versioning (v1alpha1, v1beta1, v1) provides a well-understood model for expressing API maturity.
-Adopting this model for individual plugins allows the project to reach 1.0.0 while clearly communicating that specific plugins remain experimental.
-The versioning scheme also enables controlled migration paths where plugin authors can ship both alpha and beta implementations in the same JAR, allowing users to migrate gradually.
+Kubernetes-style API versioning (v1alpha1, v1beta1, v1) provides a well-understood model for expressing configuration schema maturity.
+Adopting this model for individual plugin configurations allows the project to reach 1.0.0 while clearly communicating that specific plugin configurations remain experimental.
+The versioning scheme also enables controlled migration paths where plugin authors can support multiple configuration schema versions in the same plugin implementation, allowing users to migrate gradually.
 
-Version-based disambiguation additionally solves a practical problem for plugin authors who wish to maintain multiple implementations during API transitions.
-Currently, two classes named RecordEncryption in different packages create an ambiguity that can only be resolved through fully qualified names.
-Adding version as part of the plugin identity allows RecordEncryption/v1alpha1 and RecordEncryption/v1beta1 to coexist without requiring users to remember package structures.
+This proposal establishes the versioning mechanism that PR #96 will build upon.
+By making configuration version an explicit concept in 1.0.0, we ensure users understand that plugin configurations can evolve independently of the project version.
+PR #96 will then add the multi-file layout, dependency graphs, and JSON Schema validation that make versioned configurations more powerful.
 
 ## Proposal
 
 ### Functional changes
 
-Plugin authors will annotate their implementations with `@Version("v1alpha1")` or similar version identifiers following Kubernetes conventions.
-Users will reference these versioned plugins in configuration using the format `PluginName/version`.
-The proxy will validate at configuration parse time that the referenced version matches the plugin's declared version, failing fast with a clear error message if they diverge.
-Unversioned plugins and unversioned references will continue to work to maintain backward compatibility with existing deployments.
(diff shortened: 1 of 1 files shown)


robobario:
Prototype implementation https://github.com/kroxylicious/kroxylicious/pull/3952/

proposals/110-versioned-plugin-configuration.md
```diff
@@ -0,0 +1,218 @@
...
+
+Plugin authors will annotate their implementations with `@Version("v1alpha1")` or similar version identifiers following Kubernetes conventions.
+Users will reference these versioned plugins in configuration using the format `PluginName/version`.
+The proxy will validate at configuration parse time that the referenced version matches the plugin's declared version, failing fast with a clear error message if they diverge.
+Unversioned plugins and unversioned references will continue to work to maintain backward compatibility with existing deployments.
+
+For proxy users, configuration syntax extends from `type: RecordEncryption` to support `type: RecordEncryption/v1alpha1`.
+When a version is present in the type string, the proxy uses it both for validation and for name resolution.
+A plugin named RecordEncryption annotated with `@Version("v1alpha1")` can be referenced as either `RecordEncryption/v1alpha1` (explicit version) or `RecordEncryption` (implicit, triggers warning).
+If multiple RecordEncryption implementations exist with different versions (v1alpha1 and v1beta1), users disambiguate by including the version rather than switching to fully qualified class names.
```
robobario: that should say `can disambiguate` it's optional and they could still disambiguate using fully qualified name, backwards compat is preserved in phase 1.0.0

proposals/110-versioned-plugin-configuration.md
```diff
@@ -0,0 +1,218 @@
...
+The proxy resolves these references at startup through ServiceLoader discovery, registering each plugin under both its fully qualified name and simple name.
+When multiple plugins share the same simple name, the proxy logs an ambiguity warning and requires users to reference the plugin by its fully qualified name.
+Plugin API stability is coupled to the project's semantic version, making it difficult to ship experimental or unstable plugin APIs alongside stable ones as the project approaches 1.0.0.
+
+## Motivation
+
+Kroxylicious is approaching its 1.0.0 release, which will establish backward compatibility guarantees for public APIs.
+However, different plugins may have different maturity levels and API stability guarantees.
+Coupling all plugin APIs to a single project version creates tension between shipping experimental features and maintaining stability.
+The Record Encryption filter, for example, may need multiple major revisions before its API stabilizes, but tying those revisions to the project version would either delay 1.0.0 or force premature API commitments.
```
robobario: claude has picked our most stable plugin for the example 🤷

proposals/110-versioned-plugin-configuration.md
```diff
@@ -0,0 +1,218 @@
...
+- Plugin authors can ship multiple API versions in the same JAR for migration purposes.
+- Users are encouraged to update configurations to include explicit versions to prepare for 1.0.0.
+- All existing configurations continue to work without modification.
+- The `KROXYLICIOUS_REQUIRE_PLUGIN_VERSIONS` environment variable allows operators to opt into strict enforcement before 1.0.0 by setting it to `true`, treating missing versions as errors rather than warnings for testing purposes (particularly useful in CI pipelines and staging environments).
+
+#### Phase 2: Enforcement (1.0.0+)
+
+Phase 2 begins with the 1.0.0 release.
+
+- The enforcement policy changes from warning to error when a plugin declares a version but the configuration omits it.
```
robobario: needs a bullet to say that the annotation on the plugin becomes mandatory too

proposals/110-versioned-plugin-configuration.md
tombentley: If two unrelated plugins share a simple class name then they can be disambiguated via version number, but logically that's not the right way to fix it: The correct fix is to use an FQ name. 

It's nasty and surprising to have two different disambiguation mechanisms.

proposals/110-versioned-plugin-configuration.md
```diff
@@ -0,0 +1,218 @@
...
+An environment variable `KROXYLICIOUS_REQUIRE_PLUGIN_VERSIONS` allows operators to opt into strict enforcement before 1.0.0 for testing purposes.
+
+For filter authors and plugin developers, the `@Version` annotation becomes part of the public API contract.
+Placing `@Version("v1alpha1")` on a filter implementation signals to users that the API is experimental and subject to breaking changes.
+Transitioning from v1alpha1 to v1beta1 may involve incompatible configuration changes, but users explicitly opt into the new version by updating their configuration.
+Plugin authors can ship both versions in the same JAR during migration periods, allowing gradual rollout.
+When a plugin reaches stability, annotating it with `@Version("v1")` communicates that the API will remain backward compatible within the v1 series.
+
+Version-based disambiguation enables new packaging strategies.
+A plugin author maintaining both alpha and beta implementations can ship `io.kroxylicious.filter.encryption.alpha.RecordEncryption` annotated with `@Version("v1alpha1")` and `io.kroxylicious.filter.encryption.beta.RecordEncryption` annotated with `@Version("v1beta1")` in the same JAR.
```
tombentley: This again highlights the weirdness of having two different ways of referring to the same thing. You need the alpha/beta in the FQ class name, but you also have it in the annotation. 

This can make for a poor UX: It's easy to imagine a user meaning to update from alpha (specified as `io.kroxylicious.filter.encryption.alpha.RecordEncryption/v1alpha1`) to beta, but not really paying attention to the alpha in the package name and being confused when they change it to `io.kroxylicious.filter.encryption.alpha.RecordEncryption/v1beta1`.

The reality of this situation is there there are two different plugins masquerading as being one plugin which accepts two different configuration formats, and it's only when you use simple class naming that that pretence works. To my mind, this is pushing people towards using simple class names. However, I think simple class names are a mistake, because it means the interpretation of a configuration is relative to the plugins available to the runtime. Ultimately a configuration is about communicating intent. A configuration schema which results in configurations that are open to interpretation is a poor configuration schema. This is why #96 adopts FQ names.

proposals/110-versioned-plugin-configuration.md
```diff
@@ -0,0 +1,218 @@
...
+The enforcement policy evolves across Kroxylicious releases.
+In versions before 1.0.0, version annotations are optional and version references in configuration are optional.
+When a plugin declares a version but the configuration omits it, the proxy logs a warning but continues.
+This allows plugin authors to begin adopting version annotations while maintaining compatibility with existing configurations.
+In 1.0.0 and later, when a plugin declares a version, configurations must include that version.
+The proxy rejects configurations that omit required versions.
+An environment variable `KROXYLICIOUS_REQUIRE_PLUGIN_VERSIONS` allows operators to opt into strict enforcement before 1.0.0 for testing purposes.
+
+For filter authors and plugin developers, the `@Version` annotation becomes part of the public API contract.
+Placing `@Version("v1alpha1")` on a filter implementation signals to users that the API is experimental and subject to breaking changes.
```
tombentley: It's very odd that the thing that the version is intended to describe (the schema for the configuration) ends up in an annotation on the plugin itself.

tombentley (commented):
Thanks @robobario. I really appreciate that someone is trying to see if there's a simpler alternative to #96 which might satisfy some of the things we agreed in #82 are important to have for Kroxylicious 1.0. 

I'll admit that 96 doesn't do a great job of calling out all of its goals, but let me try to compare the two:

| Goal | #96 | #100 |
| -------|-------|------|
|Explicit plugin config version | Yes | Yes |
|Support config schema evolution | Yes | Yes |
|Plugin identification | Yes | No |
|Deeper runtime understanding of nested plugins | Yes | No |
|Fine grained dependency analysys | Yes | No |
|Better documentation of schema configs | Yes | No |
|External validation of schema configs | Yes | No |
|Editor support for writing configs | Yes | No |
|Globally shared interpretation of proxy config | Yes | No |
|Synergy with K8S | Stronger | Weaker |

It's clear that there's a very different level of ambition between the two proposals. My fear is that, out of a desire to see 1.0 happen sooner, we make decisions which undermine our ability to address some of the concerns in the left hand column. 

I asked Claude to look had how compatible there two proposals are and  [this is the result](https://gist.github.com/tombentley/e9dfb5cbd5d1216bf3048761252d7954). Reading that I think it's difficult to see 110 as an outright good move that's compatible with the ideas in 96 as a solution of those goals. Instead while there are those structural differences between the two (like the annotation, the YAML syntax and the difference in terms of simple vs FQCNs) 110 looks more like a temporary thing which would cause disruption to plugin developers.

So I think it would be helpful to hear if you think any of those goals are not actually worth striving for. If they are worth solving then I'd appreciate your review on 96: is it a good way of solving them (maybe there are better ways)? In the light of that we can figure out if 110 is the best way of doing something that's "good enough for 1.0", or if we can tweak 110 to be more compatible with 96.

I also think it's worth considering that https://github.com/kroxylicious/kroxylicious/pull/3691, which is a POC of the changes proposed in 96, is "only" a +7k change. Sure that's a lot, and it needs work on the tests and so on, so would likely grow a bit. But in the light of some of the recent big things we've merged, I don't think we should immediately be thinking that 96 is a huge piece of work that would take many months to implement.

tombentley:
> if we can tweak 110 to be more compatible with 96.

I asked Claude about this and it's the [second file in the gist](https://gist.github.com/tombentley/e9dfb5cbd5d1216bf3048761252d7954#file-revised-110-md)

robobario:
I've incorporated the feedback so that this is aligned with #96, so a step 1 that introduces explicit plugin versioning using the annotation modifications from 96 and allows for multiple configuration versions to co-exist using instanceof dispatch.

I see this as a pre-step for 96, as mentioned this is far less ambitious but hits the key requirement for me for 1.0.0 which is making our (and third party) Plugin stability guarantees visible. Accepting this wouldn't prevent us from also accepting more of 96 or other configuration changes before 1.0.0 (though I accept it would be messier to have a new design come in and make part of another design redundant).

proposals/110-versioned-plugin-configuration.md
robobario: moved to #96 style plugin identification by `version` field

robobario: updated to use #96 style plugin configuration version identification

proposals/110-versioned-plugin-configuration.md
robobario: resolved by aligning with #96, the version is still contained in the `@Plugin` annotation, but that annotation is associated with exactly one config class, so similar outcome to annotating the config class.

PR #111, opened 2026-05-20
feat: system test framework proposal

SamBarker:
## Summary

- Introduces a layered abstraction for system tests separating test intent (`ProxyScenario`) from deployment mechanism (`ProxyFixture`) with convergence gating (`ProxyHandle`)
- Organises tests into four modules by concern: `systemtest-feature`, `systemtest-operator`, `systemtest-webhook`, `systemtest-installer`
- Defines `Installer` as the primary downstream extension point — downstream varies by installation method, not proxy deployment
- Feature tests are portable across all fixtures (CRD, manifest, sidecar, standalone) with no Kubernetes dependency

## Test plan

- [ ] Review proposal for internal consistency across all sections
- [ ] Verify the abstraction model covers existing system test patterns
- [ ] Confirm TCK model works for downstream distributors

### DIFF
--- a/proposals/111-system-test-framework.md
+++ b/proposals/111-system-test-framework.md
@@ -4,7 +4,7 @@
 
 Introduce a layered abstraction for system tests that separates test intent from deployment mechanism, and organise tests into modules by what they cover: feature behaviour, operator reconciliation, webhook behaviour, and installation validation.
 
-A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. An `Installer` — the primary downstream extension point — handles getting the operator, CRDs, and RBAC into the cluster independently of how proxies are deployed.
+A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. An `Installer` — the primary downstream extension point — handles getting project components (operator, webhook, CRDs, RBAC) into the cluster independently of how proxies are deployed.
 
 Feature tests become portable across deployment mechanisms — the same test runs against a CRD-deployed proxy, a manifest-managed proxy, a standalone process, or a downstream distribution — and cheap enough to write before the production code, as a specification.
 
@@ -47,7 +47,7 @@ The framework needs one organising question answered for every test class: **wha
 
 Four categories of system test have fundamentally different concerns:
 
-- **Feature tests** — does record encryption work? Does authorisation enforce ACL rules? These tests care only that a correctly-configured proxy exists and is serving traffic. They must not care how the proxy was deployed. They have no Kubernetes dependency.
+- **Feature tests** — does record encryption work? Does authorisation enforce ACL rules? These tests care only that a proxy has been started with a given configuration. They must not care how the proxy was deployed. They have no Kubernetes dependency.
 
 - **Operator tests** — does the operator detect a configuration change and trigger a rolling restart? These tests are explicitly about the operator's reconciliation behaviour. They interact with Kubernetes resources directly and depend on the Kubernetes client.
 
@@ -117,11 +117,11 @@ interface ProxyFixture {
 
 This is an explicit call — not magic JUnit injection — because the test author needs to understand that `apply()` is a blocking operation that includes convergence waiting. Hiding it behind injection would obscure the framework's most important contract.
 
-Fixture implementations span two independent concerns: how the infrastructure is installed (CRDs, RBAC, operator Deployment) and how proxy instances are deployed. These concerns are separated by composing a `ProxyFixture` with an `Installer`.
+Fixture implementations span two independent concerns: how the infrastructure is installed (operator, webhook, CRDs, RBAC) and how proxy instances are deployed. These concerns are separated by composing a `ProxyFixture` with an `Installer`.
 
 ### `Installer` — Infrastructure Installation
 
-`Installer` handles getting the operator, CRDs, RBAC rules, and ServiceAccounts into the cluster. It is a **public interface** — the primary extension point for downstream distributors, who typically vary only by installation method (their own OLM catalog, their own Helm chart) and not by how proxies are deployed.
+`Installer` handles getting project components — operator, webhook, CRDs, RBAC rules, ServiceAccounts — into the cluster. It is a **public interface** — the primary extension point for downstream distributors, who typically vary only by installation method (their own OLM catalog, their own Helm chart) and not by how proxies are deployed.
 
 ```java
 interface Installer {
@@ -150,7 +150,7 @@ Four fixture implementations cover the deployment mechanisms:
 
 **`ManifestProxyFixture`**: translates `ProxyScenario` into a proxy configuration file and a Kubernetes Deployment, applies them, then waits for the Deployment to reach stable state. No operator or installer required.
 
-**`SidecarProxyFixture`**: takes an `Installer` as a constructor dependency. The installer puts the operator and admission webhook into the cluster; the fixture creates a pod with the injection annotation, waits for the webhook to mutate it and the sidecar to be ready, then returns a `ProxyHandle`. Feature tests run against it unchanged — the proxy happens to be a sidecar rather than a standalone Deployment.
+**`SidecarProxyFixture`**: takes an `Installer` as a constructor dependency. The installer puts the admission webhook into the cluster; the fixture creates a pod with the injection annotation, waits for the webhook to mutate it and the sidecar to be ready, then returns a `ProxyHandle`. Feature tests run against it unchanged — the proxy happens to be a sidecar rather than a standalone Deployment.
 
 **`StandaloneProxyFixture`**: starts the proxy as a local Java process with a generated configuration file, waits for the port to be ready, and returns a `ProxyHandle` with a localhost bootstrap. No Kubernetes, no installer, no namespaces. `KubernetesCapability` is not available for tests running against this fixture.
 
--- a/proposals/111-system-test-framework.md
+++ b/proposals/111-system-test-framework.md
@@ -37,7 +37,7 @@ Addressing these enables:
 - **Test-first development**: a developer writing a new filter can write a failing system test as the first commit of their feature branch, without reading framework documentation or asking QE for help.
 - **Deployment-agnostic feature tests**: the same test runs against an operator-managed proxy, a manifest-managed proxy, or a Helm installation, with no changes to the test body.
 - **Reliable convergence**: `proxyFixture.apply()` is a blocking call with a defined contract — when it returns, the proxy is serving the requested configuration. Manual polling disappears from test classes.
-- **A TCK for downstream distributions**: downstream distributors implement `Installer` for their distribution and run upstream's test modules — feature, operator, installer, and webhook — without forking.
+- **A Technology Compatibility Kit (TCK) for downstream distributions**: downstream distributors implement `Installer` for their distribution and run upstream's test modules — feature, operator, installer, and webhook — without forking. The test modules define what "correct" means; the distributor proves their packaging satisfies it.
 
 ## Proposal
 
--- a/proposals/111-system-test-framework.md
+++ b/proposals/111-system-test-framework.md
@@ -4,7 +4,7 @@
 
 Introduce a layered abstraction for system tests that separates test intent from deployment mechanism, and organise tests into modules by what they cover: feature behaviour, operator reconciliation, webhook behaviour, and installation validation.
 
-A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. An `Installer` — the primary downstream extension point — handles getting project components (operator, webhook, CRDs, RBAC) into the cluster independently of how proxies are deployed.
+A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. A `KafkaClusterFixture` provisions the upstream Kafka cluster and returns a `KafkaClusterHandle` — the same handle-based convergence pattern, making cluster provisioning pluggable across implementations (TestContainers, Strimzi, in-VM). An `Installer` — the primary downstream extension point — handles getting project components (operator, webhook, CRDs, RBAC) into the cluster independently of how proxies are deployed.
 
 Feature tests become portable across deployment mechanisms — the same test runs against a CRD-deployed proxy, a manifest-managed proxy, a standalone process, or a downstream distribution — and cheap enough to write before the production code, as a specification.
 
@@ -59,10 +59,10 @@ These are not tags on a single test suite — they are separate modules with dif
 
 | Module | Depends on | Kubernetes dependency | Portable across fixtures |
 |---|---|---|---|
-| `systemtest/feature` | `ProxyFixture`, `ProxyScenario`, `ProxyHandle`, `FilterSpec` | None | Yes — runs against any `ProxyFixture` |
-| `systemtest/operator` | `OperatorFixture`, `KubernetesCapability`, CRD types, K8s client | Yes | No — requires operator installed |
+| `systemtest/feature` | `ProxyFixture`, `KafkaClusterFixture`, `ProxyScenario`, `ProxyHandle`, `KafkaClusterHandle`, `FilterSpec` | None | Yes — runs against any `ProxyFixture` |
+| `systemtest/operator` | `OperatorFixture`, `KafkaClusterFixture`, `KubernetesCapability`, CRD types, K8s client | Yes | No — requires operator installed |
 | `systemtest/webhook` | `WebhookFixture`, `KubernetesCapability`, K8s client | Yes | No — requires webhook installed |
-| `systemtest/installer` | `ProxyFixture`, `KubernetesCapability` | Yes (except standalone) | No — one test per installer |
+| `systemtest/installer` | `ProxyFixture`, `KafkaClusterFixture`, `KubernetesCapability` | Yes (except standalone) | No — one test per installer |
 
 Feature tests do not import Kubernetes types. They cannot accidentally depend on CRDs, namespaces, or client libraries. The module boundary enforces this at compile time, not by convention.
 
@@ -74,17 +74,19 @@ A plain Java value object describing what configuration the proxy should have. N
 
 ```java
 ProxyScenario scenario = ProxyScenario.builder()
-        .withUpstream(clusterName)
+        .withUpstream(upstream)
         .withFilter(new RecordEncryptionFilterSpec(testKmsFacade))
         .build();
 
 ProxyScenario scenario = ProxyScenario.builder()
-        .withUpstream(clusterName)
+        .withUpstream(upstream)
         .withFilter(new RecordEncryptionFilterSpec(testKmsFacade)
                               .withExperimentalConfig(config))
         .withDownstreamTls(tls)
         .build();
 ```
+
+Where `upstream` is a `KafkaClusterHandle` obtained from `kafkaClusterFixture.provision()` — see [KafkaClusterFixture](#kafkaclusterfixture--upstream-cluster-provisioning).
 
 ### `FilterSpec` — The Filter DSL
 
@@ -181,15 +183,40 @@ interface ProxyHandle {
 
 `waitForRestart()` is on `ProxyHandle` rather than on any capability because restarting the proxy is meaningful across all fixture types — on Kubernetes the fixture observes the Deployment rollout; on bare metal the fixture manages the process restart directly. The concept is universal; the mechanism is fixture-specific.
 
-### Injection Model and Tags
+### `KafkaClusterFixture` — Upstream Cluster Provisioning
 
-The appropriate fixture (`ProxyFixture`, `OperatorFixture`, or `WebhookFixture`) is injected by the JUnit extension at class scope — it is an environment configuration concern, long-lived, with no timing implications. For feature tests, `ProxyHandle` is always obtained explicitly by calling `proxyFixture.apply()` in the test body. This call is blocking and includes convergence waiting; making it explicit ensures the test author understands the contract.
+Tests need an upstream Kafka cluster to proxy. Provisioning that cluster is a pluggable concern with meaningfully different implementations: an in-VM broker for fast local iteration, TestContainers for a real broker without a Kubernetes dependency, Strimzi for Kubernetes-native clusters, or a managed service.
+
+`KafkaClusterFixture` follows the same pattern as the other fixtures: the extension injects it, the test author calls `provision()` explicitly, and the returned `KafkaClusterHandle` is a token proving the cluster is ready.
 
 ```java
-ProxyHandle proxy = proxyFixture.apply(scenario);  // explicit — convergence visible
+interface KafkaClusterFixture {
+    KafkaClusterHandle provision();
+}
+
+interface KafkaClusterHandle {
+    String bootstrap();
+}
 ```
 
-**Module boundaries as compile-time separation**: `systemtest/feature` tests depend only on `ProxyFixture` and have no Kubernetes dependency. `systemtest/operator` tests depend on `OperatorFixture` and the Kubernetes client. `systemtest/webhook` tests depend on `WebhookFixture` and the Kubernetes client. These are compile-time boundaries that control what a test *can* depend on.
+`KafkaClusterHandle` serves the same role as `ProxyHandle` — you cannot interact with the cluster before provisioning has completed. `ProxyScenario.withUpstream()` takes a `KafkaClusterHandle` rather than a raw string, making the dependency on a provisioned cluster explicit in the type system.
+
+**Lifecycle is the test author's choice.** A `KafkaClusterHandle` provisioned in `@BeforeAll` is shared across all tests in the class — sensible when tests don't mutate cluster state and the cluster is expensive to provision. A handle provisioned in the test body is per-test. The extension manages cleanup at the appropriate scope. This is the same model as `ProxyFixture` — the framework provides the mechanism; the test author decides the scope.
+
+**Selection**: the `KafkaClusterFixture` implementation is selected via system property (`-Dkafka.cluster=testcontainers|strimzi|invm`), consistent with fixture and installer selection.
+
+Specifying the full set of `KafkaClusterFixture` implementations and their configuration surface is out of scope for this proposal. The key contribution is establishing that upstream cluster provisioning is a pluggable, fixture-managed concern with a handle-based convergence contract — not an ambient assumption.
+
+### Injection Model and Tags
+
+The appropriate fixtures (`ProxyFixture`, `OperatorFixture`, `WebhookFixture`, and `KafkaClusterFixture`) are injected by the JUnit extension at class scope — they are environment configuration concerns, long-lived, with no timing implications. `ProxyHandle` and `KafkaClusterHandle` are always obtained explicitly by calling `apply()` or `provision()` in the test body. These calls are blocking and include convergence waiting; making them explicit ensures the test author understands the contract.
+
+```java
+KafkaClusterHandle upstream = kafkaClusterFixture.provision();  // explicit — provisioning visible
+ProxyHandle proxy = proxyFixture.apply(scenario);               // explicit — convergence visible
+```
+
+**Module boundaries as compile-time separation**: `systemtest/feature` tests depend only on `ProxyFixture`, `KafkaClusterFixture`, and their handle types, and have no Kubernetes dependency. `systemtest/operator` tests depend on `OperatorFixture` and the Kubernetes client. `systemtest/webhook` tests depend on `WebhookFixture` and the Kubernetes client. These are compile-time boundaries that control what a test *can* depend on.
 
 **Tags as runtime skip conditions**: module boundaries cannot account for runtime environment availability. Tags like `@Kubernetes` declare that a test requires a specific runtime environment and cause the extension to skip the test when that environment is not available — e.g. skipping `systemtest/operator` tests when no cluster is reachable. Fixture and installer selection is by system property (see [Fixture and Installer Selection](#fixture-and-installer-selection)).
 
@@ -198,7 +225,9 @@ ProxyHandle proxy = proxyFixture.apply(scenario);  // explicit — convergence v
 | Concept | How obtained | Reason |
 |---|---|---|
 | `ProxyFixture` / `OperatorFixture` / `WebhookFixture` | Injected (class-scoped) | Environment config, no timing implications |
+| `KafkaClusterFixture` | Injected (class-scoped) | Environment config, no timing implications |
 | `KubernetesCapability` | Injected for Kubernetes-deployed tests | Namespace and client access for resource observation |
+| `KafkaClusterHandle` | Always explicit via `provision()` | Provisioning is a blocking operation; must be visible |
 | `ProxyHandle` | Always explicit via `apply()` | Convergence is a blocking operation; must be visible |
 
 ### `KubernetesCapability` — Cluster Environment Access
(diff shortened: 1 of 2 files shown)


proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,494 @@
...
+
+This model has already been validated in the codebase: the checksum annotation mechanism in the operator was built test-first. The system test described what the operator should do; the implementation followed.
+
+Three problems prevent this from being the norm:
+
+1. **Setup cost**: expressing "a proxy with this filter" requires navigating CRD builders, template classes, namespace management, and ingress configuration. The ceremony dwarfs the test.
+
+2. **No convergence contract**: the framework does not define when the proxy is ready. Each test class independently polls for readiness, with varying strategies and varying reliability.
+
+3. **Operator coupling**: every test implicitly requires the operator. Feature tests — which care only that a correctly-configured proxy is serving traffic — cannot run without the full operator installation. This conflates feature correctness with operator correctness and prevents fast local iteration.
```
franvila: The purpose of a System Tests is having all installed as you want on a real environment. We decided to go only for Kubernetes installation as we didn't have enough resources and time to cover everything. Ideally, we should have all kind of installations, such as bare-metal, Kubernetes... 

Totally agree with system tests decoupling with the installation. The test cases should not care about how Kroxylicious has been installed, just check what is needed. Having a quicker installation will easy the Test-first development as we don't need to depend on the operator.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,494 @@
...
+
+Addressing these enables:
+
+- **Test-first development**: a developer writing a new filter can write a failing system test as the first commit of their feature branch, without reading framework documentation or asking QE for help.
+- **Deployment-agnostic feature tests**: the same test runs against an operator-managed proxy, a manifest-managed proxy, or a Helm installation, with no changes to the test body.
+- **Reliable convergence**: `proxyFixture.apply()` is a blocking call with a defined contract — when it returns, the proxy is serving the requested configuration. Manual polling disappears from test classes.
+- **A TCK for downstream distributions**: downstream distributors implement `Installer` for their distribution and run upstream's test modules — feature, operator, installer, and webhook — without forking.
```
franvila: What does TCK mean?

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,494 @@
+# 111 - System Test Framework
+
+## Summary
+
+Introduce a layered abstraction for system tests that separates test intent from deployment mechanism, and organise tests into modules by what they cover: feature behaviour, operator reconciliation, webhook behaviour, and installation validation.
+
+A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. An `Installer` — the primary downstream extension point — handles getting the operator, CRDs, and RBAC into the cluster independently of how proxies are deployed.
+
+Feature tests become portable across deployment mechanisms — the same test runs against a CRD-deployed proxy, a manifest-managed proxy, a standalone process, or a downstream distribution — and cheap enough to write before the production code, as a specification.
```
k-wall, line 9: Aside: I'd like to get Kroxylicious into Operatorhub soon.  So having test that can are agnostic to their install methodology will be an enabler.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,494 @@
...
+
+## Current Situation
+
+The system test suite covers the right things at the right level. Assertions in `RecordEncryptionST`, `AuthorizationST`, and `EntityIsolationST` test feature behaviour cleanly. `OperatorChangeDetectionST` tests operator reconciliation behaviour. The abstraction breaks down only in setup.
+
+Every feature test class has a private `deployXxx()` method that reimplements the same builder/template pattern against the operator's CRD types. Adding a new optional parameter (e.g. `ExperimentalKmsConfig`) requires touching every one of them. Timing workarounds are scattered across test classes with comments pointing at unresolved issues. The convergence question — "is the proxy actually serving the configuration I just applied?" — is answered by ad hoc polling in each test class rather than by a framework-level contract.
+
+This setup cost has a second-order effect: system tests are written after features merge, delegated to QE because they are too expensive for a developer to include in a feature PR. The test framework is the bottleneck, not the assertions.
```
k-wall: Let's describe the problem without the Red Hat terminology.

proposals/111-system-test-framework.md
k-wall: Sam's borrowing the Java parlance https://en.wikipedia.org/wiki/Technology_Compatibility_Kit

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,494 @@
...
+```java
+KafkaClient inNamespace(String namespace);
+String getImage();
+void preloadImage();
+```
+
+These move to a `KubernetesClientCapability`:
+
+```java
+interface KafkaClient {
```
k-wall, line 297: One thing I dislike about the system tests is that all our client interactions are via a client CLI spawned in a separate process. 

CLIs expose pretty basic operators: they boil down to produce or fetch from a topic.  Consumer groups are possible but testing things like produce transactions + commit offsets is pretty much impossible.  This will become a real impediment when we come to test things like Router implementations.

I think we need to be able to program the KafkaClient from the test so we are freer to express the fully richness of the Kafka.

A secondary issue is that relying on a CLI is slow: you pay the process startup time.  

I have had an itch to scratch for a while which I think I've expressed to you both.   Could we expose  a common Java interface for Librdkafka and Sarama using Java Foreign Functions?  We could even _partially_ implement the Kafka Consumer, Producer and Admins client interfaces.  The system tests could then be written in term of the Java Kafka API and exercise java/go/librdkafka using the same test.  
Now we've got AI assistance I think writing a POC for this is in reach.

k-wall:
I like the ideas you are expressing in the proposal.  I think it is the right way to go.

henryZrncik (approved):
Looks Great to me. 
I think most of the proposed solution suits perfectly. I think to to gret extent current situation of System tests addressed mainly testing assuming operator and subsequently using resource manager mostly on test level instead of inside function calls. Because there is also mention of different enviroments (minikube, ocp etc) we need to also address the need to make http calls and client calls regardless of acessing local cluster or remote, but this represents minimal changes that are to be part of this. 👍

proposals/111-system-test-framework.md
SamBarker: Expanded at first use in 4a8abb3.

proposals/111-system-test-framework.md
SamBarker: Addressed in 9e7748b.

proposals/111-system-test-framework.md
SamBarker: Agreed — the CLI-based approach is too limiting, especially as we get into transactions, consumer groups, and richer Kafka interactions. The KafkaClient interface in the proposal is intended as the seam where a better client strategy plugs in, but I think working out what that strategy is (FFI, direct Java clients, something else) is a separate concern from the framework layering here. Happy to pick it up as follow-on work once this lands.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+
+1. **Setup cost**: expressing "a proxy with this filter" requires navigating CRD builders, template classes, namespace management, and ingress configuration. The ceremony dwarfs the test.
+
+2. **No convergence contract**: the framework does not define when the proxy is ready. Each test class independently polls for readiness, with varying strategies and varying reliability.
+
+3. **Operator coupling**: every test implicitly requires the operator. Feature tests — which care only that a correctly-configured proxy is serving traffic — cannot run without the full operator installation. This conflates feature correctness with operator correctness and prevents fast local iteration.
+
+Addressing these enables:
+
+- **Test-first development**: a developer writing a new filter can write a failing system test as the first commit of their feature branch, without reading framework documentation or asking QE for help.
```
k-wall: We should avoid Red Hat specific acronyms

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+- **Deployment-agnostic feature tests**: the same test runs against an operator-managed proxy, a manifest-managed proxy, or a Helm installation, with no changes to the test body.
+- **Reliable convergence**: `proxyFixture.apply()` is a blocking call with a defined contract — when it returns, the proxy is serving the requested configuration. Manual polling disappears from test classes.
+- **A Technology Compatibility Kit (TCK) for downstream distributions**: downstream distributors implement `Installer` for their distribution and run upstream's test modules — feature, operator, installer, and webhook — without forking. The test modules define what "correct" means; the distributor proves their packaging satisfies it.
```
k-wall, line 40: So this means, say, an organisation makes a derivative version of Kroxylicious, distributed say, via a private OLM catalogue would be catered for?  They'd implement Installer to know about how to install from the private OLM catalogue

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Proposal
+
+### The Primary Seam
+
+The framework needs one organising question answered for every test class: **what is this test covering?** The answer determines which module the test belongs to — not which tag it carries, but which compile-time dependencies it has.
+
+Four categories of system test have fundamentally different concerns:
+
+- **Feature tests** — does record encryption work? Does authorisation enforce ACL rules? These tests care only that a correctly-configured proxy exists and is serving traffic. They must not care how the proxy was deployed. They have no Kubernetes dependency.
```
k-wall: Where do transport tests fit into this hierarchy? Upstream/Downstream TLS, certificate auth etc.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Fixture Implementations
+
+Tests never instantiate fixtures or installers — the JUnit extension reads system properties (`-Dfixture`, `-Dinstaller`) and composes them. The composition is extension-internal; the test sees only an injected `ProxyFixture`.
+
+Four fixture implementations cover the deployment mechanisms:
+
+**`CrdProxyFixture`**: takes an `Installer` as a constructor dependency. The installer puts the operator into the cluster; the fixture applies Kroxylicious CRDs (`KafkaProxy`, `VirtualKafkaCluster`, `KafkaProtocolFilter`) via Server-Side Apply, then waits for observable convergence signals — the controller has reconciled the resources and the Deployment has reached stable state with updated replicas ready and serving. The fixture knows the CRD schema and the convergence protocol, not the operator's internals.
+
+**`ManifestProxyFixture`**: translates `ProxyScenario` into a proxy configuration file and a Kubernetes Deployment, applies them, then waits for the Deployment to reach stable state. No operator or installer required.
```
k-wall: Is this opinionated about ingress?

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+
+A downstream distributor implements `Installer` for their distribution and composes it with upstream's fixture — no need to reimplement proxy deployment or convergence logic.
+
+### Fixture Implementations
+
+Tests never instantiate fixtures or installers — the JUnit extension reads system properties (`-Dfixture`, `-Dinstaller`) and composes them. The composition is extension-internal; the test sees only an injected `ProxyFixture`.
+
+Four fixture implementations cover the deployment mechanisms:
+
+**`CrdProxyFixture`**: takes an `Installer` as a constructor dependency. The installer puts the operator into the cluster; the fixture applies Kroxylicious CRDs (`KafkaProxy`, `VirtualKafkaCluster`, `KafkaProtocolFilter`) via Server-Side Apply, then waits for observable convergence signals — the controller has reconciled the resources and the Deployment has reached stable state with updated replicas ready and serving. The fixture knows the CRD schema and the convergence protocol, not the operator's internals.
```
k-wall: So is CrdFixture converting from ProxyScenario/RecordEncryptionFilterSpec objects etc into CRDs?

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+    kroxylicious.createOrUpdateResources();
+}
+```
+
+**After**:
+
+```java
+@Test
+void ensureClusterHasEncryptedMessage() {
+    testKmsFacade.getTestKekManager().generateKek(KEK_PREFIX + topicName);
```
k-wall, line 415: I'm not getting what's responsible for providing the KafkaCluster.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+The proposed model separates them. `KafkaClient` becomes a richer test-facing API — pure intent, expressed in terms of Kafka operations rather than CLI invocations:
+
+```java
+interface KafkaClient {
+    void produce(String topic, String bootstrap, List<ProducerRecord> records);
+    List<ConsumerRecord> consume(String topic, String bootstrap, int count);
+    void createTopic(String topic, String bootstrap, int partitions, int replicas);
+
+    // Richer operations — enabled by in-process drivers
+    void produceInTransaction(String topic, String bootstrap, List<ProducerRecord> records);
```
k-wall, line 303: I'm not sure this API is sufficiently rich to model use cases like sendOffsetsToTransaction.

proposals/111-system-test-framework.md
k-wall: >  The KafkaClient interface in the proposal is intended as the seam where a better client strategy plugs in, but I think working out what that strategy is (FFI, direct Java clients, something else) is a separate concern from the framework layering here. Happy to pick it up as follow-on work once this lands.

Makes sense.

k-wall: Probably makes sense to spin this out as a separate proposal.  I'd quite like to try some prototyping with FFI.   I could volunteer to push on that proposal separately.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
...
+Feature tests do not import Kubernetes types. They cannot accidentally depend on CRDs, namespaces, or client libraries. The module boundary enforces this at compile time, not by convention.
+
+All four modules are consumable as a TCK. A downstream distributor runs feature tests to prove their distribution satisfies the proxy's behavioural contract, operator tests to prove reconciliation works with their installation, webhook tests to prove admission webhook behaviour, and installer tests to prove their installation method works.
+
+### `ProxyScenario` — Intent Without Deployment
+
+A plain Java value object describing what configuration the proxy should have. No knowledge of namespaces, CRD templates, or deployment mechanism.
+
+```java
+ProxyScenario scenario = ProxyScenario.builder()
```
k-wall: Is the intention here to have this builder compose the Proxy's own existing configuration types?  I'm concerned about introducing another configuration model into the proxy (core config model + crds + a new one).

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,505 @@
+# 111 - System Test Framework
+
+## Summary
+
+Introduce a layered abstraction for system tests that separates test intent from deployment mechanism, and organise tests into modules by what they cover: feature behaviour, operator reconciliation, webhook behaviour, and installation validation.
+
+A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. An `Installer` — the primary downstream extension point — handles getting the operator, CRDs, and RBAC into the cluster independently of how proxies are deployed.
```
k-wall: What's actually providing the upstream kafka and how's it configured?  I think ideally we'd like this to be pluggable too.  It could be an InVM or TestContainer cluster coming from the test extension, a cluster stood up by Strimzi, a Cloud Kafka Service (Confluent).

proposals/111-system-test-framework.md
SamBarker: Yes exactly.

proposals/111-system-test-framework.md
SamBarker: Not entirely sold it needs a proposal but very keen to keep it separate to this, and to see it done. 😁

proposals/111-system-test-framework.md
SamBarker: I think the transport level is hopefully just a feature test. I can see it might spill into the builders a bit. I've tweaked the definition a little to clarify things.

proposals/111-system-test-framework.md
SamBarker: The intent is for tests to be agnostic of ingress. The test calls `proxy.bootstrap()` and gets the right address — it doesn't know or care how that was set up.

There is real tension here though: the proxy's advertised addresses have to match the ingress configuration, so the fixture implementation needs to solve that coupling. That's non-trivial — but it's an implementation concern, not a test-facing one. As long as `ProxyScenario` and `ProxyHandle` keep the test abstracted from the mechanics of config generation, we're in the best place we can be.

proposals/111-system-test-framework.md
SamBarker: Yes. Each `ProxyFixture` implementation translates the same `ProxyScenario` into whatever the deployment mechanism needs. `CrdProxyFixture` generates CRD resources (`KafkaProxy`, `VirtualKafkaCluster`, `KafkaProtocolFilter`). `ManifestProxyFixture` generates a proxy configuration file and a Kubernetes Deployment. `StandaloneProxyFixture` generates a configuration file and starts a local process. The test provides the same `ProxyScenario` regardless — the fixture decides what to do with it.

proposals/111-system-test-framework.md
SamBarker: The intent is for ProxyScenario to be a thin, purpose-built DSL for system tests — not a general configuration model. It describes high-level intent: "a proxy with filter X and upstream Y" without caring whether that manifests as a YAML blob in a config file or a set of CRD resources.

The bet is that there are enough common high-level actions to make this worthwhile. For example, a record encryption test might say:

```java
withFilter(new RecordEncryptionSpec(Kms.VAULT, Set.of("topicA", "topicB")))
```

The test doesn't care about KEK naming patterns, key rotation config, or how the KMS is connected — it just says "these topics get encrypted" and asserts that other topics don't. KMS provisioning and configuration sits in the installer layer, not the test. That's the level of gesture the DSL aims at.

The tension is real — this is the hardest seam to extract cleanly, and there are good questions to work through: where does `RecordEncryptionSpec` come from? How do test authors discover available gestures? How does this work for BYO filters? I've seen purpose-built test DSLs work very well in the past but they're expensive to build and iterate on. Answering those questions now feels like over-specifying — this proposal is trying to answer the bigger-picture question of how we structure system tests and make downstream packages easier to validate. The DSL shape will need its own iteration once the overall structure lands.

proposals/111-system-test-framework.md
SamBarker: Addressed in 9d9b831 — introduced KafkaClusterFixture and KafkaClusterHandle.

proposals/111-system-test-framework.md
SamBarker: Addressed in 9d9b831 — introduced KafkaClusterFixture and KafkaClusterHandle as the pluggable seam for upstream cluster provisioning.

proposals/111-system-test-framework.md
SamBarker: Agreed — the API as shown isn't rich enough for things like `sendOffsetsToTransaction`. The current high-level gestures (produce, consume, create topic) map to what the CLI-based clients can do today, and that's the starting point. For system tests, the goal is high-level gestures, not covering every possible Kafka operation — that's for lower-level tests.

I'm 100% on board that the CLI isn't good enough long-term, but what the right answer looks like depends heavily on what the FFI work throws up and what we can sensibly offer from that. I'd rather let that inform the API shape than lock it in now.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
...
+    void install(Component component);
+    void uninstall(Component component);
+}
+```
+
+We recognise there is likely to be a set of standard installers which need customising only by simple parameters such as a Helm chart URI or an OLM bundle image. Specifying those implementations and their configuration surface is out of scope for this proposal. A downstream distributor could configure an upstream-provided implementation or implement the interface directly.
+
+### `ProxyFixture` Implementations
+
+Tests never instantiate fixtures or installers — the JUnit extension reads system properties (`-Dfixture`, `-Dinstaller`) and composes them. The composition is extension-internal; the test sees only an injected fixture.
```
robobario: worth having an env var style and precedence rules? env vars may drop more easily into things like github CI, where system props give you more control within an environment.

So we could have `KROXYLICIOUS_TEST_FIXTURE` and `KROXYLICIOUS_TEST_INSTALLER` env vars but prefer the system props. I guess a namespacing on the system props could also be worthwhile.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
...
+
+The `ProxyFixture` and `Installer` interfaces are designed for extension. Downstream distributors typically implement `Installer` and compose it with upstream fixtures; distributors with fundamentally different deployment models implement `ProxyFixture` directly. Once published, `ProxyFixture`, `KafkaClusterFixture`, `Installer`, `ProxyScenario`, `ProxyHandle`, and `KafkaClusterHandle` become API surface for downstream consumers — their signatures should be treated as a compatibility commitment.
+
+## Rejected Alternatives
+
+### Magic JUnit injection of `ProxyHandle`
+
+We considered having the JUnit extension inject `ProxyHandle` directly as a test parameter (similar to how `ProxyFixture` is injected), with convergence waiting happening transparently during parameter resolution. This hides the most important contract in the framework — that `apply()` is a blocking operation — behind invisible lifecycle callbacks. The test author would not see when convergence happens, making it harder to reason about test timing and harder to debug when convergence fails. The explicit `proxyFixture.apply()` call keeps the blocking operation visible.
```
robobario, line 509: good call, I think it'll be a lot better knowing exactly when we manifest what resources. you can more easily reason about when test setup occurs relative to the resources being created and so on.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
...
+
+Specifying the full set of `KafkaClusterFixture` implementations and their configuration surface is out of scope for this proposal. The key contribution is establishing that upstream cluster provisioning is a pluggable, fixture-managed concern with a handle-based convergence contract — not an ambient assumption.
+
+### Injection Model and Tags
+
+The appropriate fixtures (`ProxyFixture`, `OperatorFixture`, `WebhookFixture`, and `KafkaClusterFixture`) are injected by the JUnit extension at class scope — they are environment configuration concerns, long-lived, with no timing implications. `ProxyHandle` and `KafkaClusterHandle` are always obtained explicitly by calling `apply()` or `provision()` in the test body. These calls are blocking and include convergence waiting; making them explicit ensures the test author understands the contract.
+
+```java
+KafkaClusterHandle upstream = kafkaClusterFixture.provision();  // explicit — provisioning visible
+ProxyHandle proxy = proxyFixture.apply(scenario);               // explicit — convergence visible
```
robobario, line 218: maybe they should both use the `provision` language? `apply` is awkwardly a functional method name and kubernetes concept.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
+# 111 - System Test Framework
+
+## Summary
+
+Introduce a layered abstraction for system tests that separates test intent from deployment mechanism, and organise tests into modules by what they cover: feature behaviour, operator reconciliation, webhook behaviour, and installation validation.
+
+A `ProxyScenario` describes the desired proxy configuration in deployment-agnostic terms; a `ProxyFixture` translates that into running infrastructure and blocks until convergence; the resulting `ProxyHandle` is a token of convergence that gates all subsequent interaction. A `KafkaClusterFixture` provisions the upstream Kafka cluster and returns a `KafkaClusterHandle` — the same handle-based convergence pattern, making cluster provisioning pluggable across implementations (TestContainers, Strimzi, in-VM). An `Installer` — the primary downstream extension point — handles getting project components (operator, webhook, CRDs, RBAC) into the cluster independently of how proxies are deployed.
```
robobario: might be a bit nitpicky but the scenario name might be a bit overloaded since we're in a testing context, I read it and thing of an entire given/when/then BDD style test. but other options are also a bit crusty in their own ways ProxyConfig, ProxyDefinition, ProxyDescriptor

proposals/111-system-test-framework.md
robobario: starting with passing around jackson JsonObject blobs or HashMaps for filter config would be no worse than [what's implemented](https://github.com/kroxylicious/kroxylicious/blob/23bfbf97adfa5f652bd9f07db2d5168bee66dda4/kroxylicious-systemtests/src/main/java/io/kroxylicious/systemtests/templates/kroxylicious/KroxyliciousFilterTemplates.java#L103). Agree that we can decide on that separately. (should keep in mind this is something we can iterate on rapidly as opposed to public API)

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
...
+Webhook tests are about the admission webhook's mutation behaviour — they need to create pods and verify that the webhook mutates them correctly. `WebhookFixture` calls `install(WEBHOOK)` internally and exposes the webhook's surface: creating annotated pods, inspecting the resulting pod spec, and verifying that injection produces valid sidecar configurations.
+
+Tests in `systemtest/webhook` receive a `WebhookFixture` rather than a `ProxyFixture`. The test drives the webhook through pod creation and asserts the mutation outcome.
+
+### `ProxyHandle` — A Token of Convergence
+
+The only way to obtain a `ProxyHandle` is through `ProxyFixture.apply()`. This means a test cannot accidentally interact with the proxy before convergence has been waited for.
+
+```java
+interface ProxyHandle {
```
robobario, line 174: who disposes of a provisioned thing? should Fixtures have a symmetrical deprovision? Or Handles a dispose/close method? Or is it done automatically at end of test by the framework?

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
...
+    String bootstrap();
+    String bootstrap(ClientLocation location);
+    ProxyHandle reconfigure(ProxyScenario scenario);
+    void waitForRestart();
```
robobario, line 178: is this triggering a restart, or waiting for one that was caused by a different operation like reconfigure? Shouldn't reconfigure converge like apply?

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,530 @@
...
+
+**Selection**: the `KafkaClusterFixture` implementation is selected via system property (`-Dkafka.cluster=testcontainers|strimzi|invm`), consistent with fixture and installer selection.
+
+Specifying the full set of `KafkaClusterFixture` implementations and their configuration surface is out of scope for this proposal. The key contribution is establishing that upstream cluster provisioning is a pluggable, fixture-managed concern with a handle-based convergence contract — not an ambient assumption.
+
+### Injection Model and Tags
+
+The appropriate fixtures (`ProxyFixture`, `OperatorFixture`, `WebhookFixture`, and `KafkaClusterFixture`) are injected by the JUnit extension at class scope — they are environment configuration concerns, long-lived, with no timing implications. `ProxyHandle` and `KafkaClusterHandle` are always obtained explicitly by calling `apply()` or `provision()` in the test body. These calls are blocking and include convergence waiting; making them explicit ensures the test author understands the contract.
```
robobario, line 214: just wondering, are the fixtures themselves stateless? I'm wondering why we inject vs provide a static factory method.

robobario (commented):
Thanks @SamBarker sounds like a nice separation of concerns,the main thing I'd like added is some details about resource teardown, who's responsible, when/where is it executed?

proposals/111-system-test-framework.md
SamBarker: Good question — I've added a "Symmetrical cleanup" paragraph to the injection model section making this explicit. Every install, provision, and apply action has a corresponding teardown managed by the extension at the appropriate JUnit lifecycle event. No explicit `deprovision()` or `close()` on handles.

I can imagine a future need for explicit mid-test teardown (e.g. simulating a cluster disappearing), but I'd rather add that when a concrete use case materialises than design it in eagerly. Keeping disposal implicit for now is as much a scope-management exercise as anything.

proposals/111-system-test-framework.md
SamBarker: I'm open to changing the name — `apply` is very generic and I take the point about overloading. That said, I think the two verbs communicate genuinely different semantics: `provision()` is "create this thing from nothing", while `apply()` is "converge to this desired state" — closer to `kubectl apply` than to provisioning. It's also re-entrant (`reconfigure` returns a new handle), which fits the declarative model.

Happy to bike-shed the name, but I'd want the replacement to carry that declarative/convergence connotation rather than flattening both operations to the same verb.

proposals/111-system-test-framework.md
SamBarker: Agreed — `JsonObject` blobs give us a good starting point rather than trying to spec out the DSL now. That's still my ultimate goal: a test author expressing "these topics get encrypted with this KMS" without caring about filter config shape. But iterating toward that from a working blob-based foundation is a lot more tractable than speccing it up front.

proposals/111-system-test-framework.md
SamBarker: Fair point — and on reflection the asymmetry is a design smell I hadn't noticed. `OperatorFixture` and `WebhookFixture` have no `*Scenario` siblings because their "input" is the raw surface they're testing directly (CRDs, pod specs). `ProxyScenario` isn't one of a family — it's a one-off concept, and the name implies siblings that don't exist and shouldn't.

`ProxyDefinition` works better: it reads as "the definition of what the proxy should be" without the BDD connotation, and it's a natural extension point — we can start with JSON blobs and evolve toward a richer DSL without changing the type name or the fixture API.

proposals/111-system-test-framework.md
SamBarker: `waitForRestart()` is for tests where the restart itself is the scenario — e.g. asserting that clients reconnect seamlessly and keep processing after a proxy pod or process restart. It's a high-level gesture: "restart the proxy and gate until it's back", not a low-level handle on the pod lifecycle.

`reconfigure()` is a separate concern — it converges the proxy to a new configuration and returns a new handle when done. Whether that involves a rolling restart internally is the fixture's problem, not the test's.

proposals/111-system-test-framework.md
SamBarker: Fixtures hold some state — `CrdProxyFixture` carries an `Installer` reference for example — but no per-test state. They're long-lived environment handles scoped to the test class.

The reason for injection over a static factory is that the extension owns the selection and composition logic: it reads environment variables, picks the right implementation, and wires dependencies (e.g. the `Installer` into `CrdProxyFixture`). A static factory would either need to duplicate that logic or force the test to call something like `ProxyFixture.fromEnv()`, which is framework plumbing leaking into test code.

proposals/111-system-test-framework.md
SamBarker: Good suggestion — the whole point of fixture selection is to make automation easy, and env vars compose more naturally with CI pipelines than JVM flags. Less idiomatic Java, but the right tradeoff here.

On reflection though, env vars alone might be sufficient: they work in Maven via Surefire `<environmentVariables>`, they work in Gradle, and they work in whatever a downstream distributor's CI looks like. System properties are Java convention, but they don't buy anything that env vars don't already cover. Inclined to go env vars only and keep the API surface simpler — `KROXYLICIOUS_TEST_FIXTURE`, `KROXYLICIOUS_TEST_INSTALLER`, `KROXYLICIOUS_TEST_KAFKA_CLUSTER`. Thoughts?

proposals/111-system-test-framework.md
robobario: okay probably doesn't need the naming perfected here, starting with `wait` makes its sound like the waiting is the command. if it's responsible for restarting it might be better named as just `restart()` (our other converging methods don't have anything about wait in their method names)

proposals/111-system-test-framework.md
robobario: automatically cleaning up sounds good to me for a start

proposals/111-system-test-framework.md
robobario: cool, hadn't clocked the difference. mm maybe apply is best then.

proposals/111-system-test-framework.md
```diff
@@ -0,0 +1,534 @@
...
+# Manifest-managed proxy (no operator)
+KROXYLICIOUS_TEST_FIXTURE=manifest mvn test
+
+# Standalone with in-VM Kafka (fast local iteration)
+KROXYLICIOUS_TEST_FIXTURE=standalone KROXYLICIOUS_TEST_KAFKA_CLUSTER=invm mvn test
+```
+
+Environment variables work uniformly across Maven (via Surefire `<environmentVariables>`), Gradle, and any CI tooling a downstream distributor uses. The extension composes them: it instantiates the installer, passes it to the fixture constructor, and manages the lifecycle. When `KROXYLICIOUS_TEST_INSTALLER` is not set, the fixture uses its default (manifest-based for fixtures that require an installer). Standalone and manifest fixtures do not take an installer. When `KROXYLICIOUS_TEST_KAFKA_CLUSTER` is not set, the extension uses a sensible default for the fixture type.
+
+### `KafkaClient` Abstraction
```
robobario, line 274: This section looks a bit light compared to the rest, but I agree in principal that it makes sense to abstract like this along the client/environment lines. Do you want to shuffle it to a new proposal or take it all the way here, or we could vote on the principle since this is system tests and not public API? I assume we'd need to define environment variables for picking the client driver and environment and so on?

Should probably check, do you imagine this TCK idea being public API with the same strictures about modifications, or could we operate on a best-effort compatibility here?

robobario (approved):
LGTM, thanks @SamBarker

proposals/111-system-test-framework.md
henryZrncik: I am down for using environment variables here: all our STs use this approach so  `KROXYLICIOUS_TEST_FIXTURE` etc. seems as good way to go for me

franvila (approved):
Thanks Sam for this great proposal.

This will be a huge advance in our system test framework to cover all the installation mechanisms and decoupling all that are non-test related from the test cases.

System tests have been grown fast in recent time so we would need something like this to stop growing without a clear view of where we want to go.

When all of this implementation is finished, we will need to decide what to include in our current PR pipelines (installation mechanisms supported on Github Actions, kafka clients...). Maybe we would need to add some kind of `REGRESSION` annotation for those unskippable tests and include nightly builds to cover all the matrix tests, but this will be a discussion for the future.

PR #112, opened 2026-05-22
Add `DataVirtualization` router 

tombentley:
The proposal for https://github.com/kroxylicious/kroxylicious/issues/4392

### DIFF
--- a/proposals/112-topic-based-routing.md
+++ b/proposals/112-topic-based-routing.md
@@ -1,0 +1,387 @@
+# 112 - Add `TopicPartitionRouter`
+
+This proposal describes a `Router` implementation that presents multiple upstream Kafka clusters to clients as a single virtual cluster, routing requests based on topic ownership.
+
+## Current situation
+
+[Proposal 070](070-routing-api.md) introduced the `Router` plugin API, which enables the proxy to route requests to multiple upstream clusters.
+No concrete `Router` implementation exists yet.
+
+## Motivation
+
+Organisations often operate multiple Kafka clusters — for scaling, isolation, geography, or organisational reasons.
+A client is locked in to accessing only the topics available on the single cluster it connects to.
+This creates operational headaches when an application needs to process data that lives across multiple clusters.
+
+The options available today have stark compromises:
+
+* **Multiple clients.** The application code can be changed to use multiple producer and/or consumer instances, one for each cluster. But this is not viable for all applications — transactions and consumer groups are scoped to a single cluster, so workflows that need cross-cluster transactions or coordinated consumption cannot use this approach.
+* **Topic replication.** Tools like MirrorMaker 2 can replicate topics between clusters so that all the data an application needs lives in a single cluster. But this introduces operational burden (managing replication topologies), asynchrony (replicated data lags behind the source), and offset divergence (the same record has different offsets on source and replica, complicating failover and exactly-once semantics).
+
+Data virtualization can address this by presenting a virtual cluster that is composed of multiple real clusters.
+Clients connect to the virtual cluster and interact with it as if it were a single Kafka cluster.
+The proxy transparently routes each request to the correct upstream cluster based on which cluster owns the topics involved.
+
+The Kafka protocol makes this non-trivial.
+Many request types batch operations across multiple topics in a single request (e.g. a `PRODUCE` request may contain records for topics owned by different clusters).
+The router must decompose such requests, fan out the sub-requests in parallel, and recompose a single response.
+Beyond basic produce and consume, stateful protocol features — idempotent producers, transactions, consumer groups, fetch sessions — each add further complexity.
+This proposal addresses these features incrementally, with the simplest cases (produce, metadata) first and the more complex cases (transactions, consumer groups) building on that foundation.
+
+
+## Proposal
+
+### Overview
+
+The data virtualization router partitions Kafka traffic across multiple upstream clusters using two complementary strategies:
+
+**Topic-based partitioning.** 
+Each topic is owned by exactly one upstream cluster.
+The router determines the owning cluster from the topic name (via an unambiguous mapping) and routes data-plane operations accordingly.
+When a single client request references topics on different clusters, the router decomposes the request, fans out the sub-requests, fans-in the sub-responses and recomposes a single response to be returned to the client.
+
+Whether this decomposition is needed for a given API depends on the `NodeIdMapping` strategy in use (see proposal 070).
+With a _dedicated_ (one-to-one) mapping, each virtual node belongs to exactly one route, so a well-behaved Kafka client's leader-directed requests (e.g. `PRODUCE`, `FETCH`) will naturally contain only topics from one route — no decomposition is needed for these APIs.
+With a _shared_ (many-to-one) mapping — for example, if proxy instances are presented to clients as broker nodes — a single virtual node may serve brokers from multiple routes, and leader-directed requests _can_ contain topics from different routes.
+The router does not know which mapping is in use, so it must handle the general case.
+Decomposition is always required for non-leader-directed APIs such as `METADATA` and admin operations (`CREATE_TOPICS`, `DELETE_TOPICS`, etc.), regardless of the mapping strategy.
+
+**Subject-based partitioning.**
+Some Kafka protocol operations cannot be partitioned by topic because the topic is not known at the point where the routing decision must be made.
+
+The clearest example is transactions.
+The transactional protocol begins with `INIT_PRODUCER_ID` (which names only a `transactional.id`, not a topic) and proceeds through `ADD_PARTITIONS_TO_TXN` and `END_TXN`.
+The router does not learn which topics are involved until `ADD_PARTITIONS_TO_TXN`, but by that point the transaction coordinator has already been chosen and the producer ID has been allocated on a specific cluster.
+In principle the router could cache the mapping between the client's producer ID and each cluster's producer ID, but this mapping is not crash-safe: if the proxy restarts, the mapping is lost, and the client's in-flight transaction cannot be recovered.
+Building a crash-safe mapping would require the proxy to maintain its own replicated state — effectively implementing a distributed transaction coordinator.
+
+Subject-based partitioning avoids this problem entirely.
+An authenticated user (a _subject_) is pinned to a specific route for all coordinator-bound operations (transactions, consumer groups).
+Because the subject is determined at connection time (via SASL authentication), the router always knows which cluster should handle coordinator operations, without needing to inspect the request body.
+Topic-addressed operations (`PRODUCE`, `FETCH`, etc.) continue to be decomposed by topic ownership as normal.
+
+Consumer groups face a similar challenge: the group coordinator manages offsets and membership for the group as a whole, and these operations cannot be split across clusters.
+Subject-based partitioning ensures that a given user's group operations always go to the same cluster.
+
+Another way to think about this: The end user is forced to declare ahead-of-time which cluster each given subject's topics will all be in, at least for subjects corresponding to applications that use transactions or groups.
+
+The following sections describe the detailed design of both strategies.
+
+### Routing table
+
+Topics are assigned to routes using a _routing table_ with three tiers:
+
+1. **Explicit topic names** — a topic name listed directly in a route's configuration.
+2. **Topic name prefixes** — a topic whose name starts with a prefix defined on a route. This is a convenience to save enumerating large numbers of topic names when all the topics with a prefix live in the same cluster.
+3. **Default route** — a fallback route for topics that match no explicit name or prefix.
+
+Precedence follows the order above: explicit name takes priority over prefix, prefix over default.
+
+Prefixes must be _disjoint_ across routes: no route's prefix may be a prefix of another route's prefix.
+This is validated at startup.
+Without this constraint, a topic name could match multiple prefixes, making the routing decision ambiguous.
+
+### Request decomposition
+
+Each Kafka API has its own request and response schema, and many APIs batch operations across multiple topics in a single request.
+A `PRODUCE` request, for example, may contain records for several topics, each potentially owned by a different cluster.
+The router must split such requests by topic ownership, dispatch the sub-requests to the correct routes, and merge the responses.
+
+This is handled by per-API _decomposers_ implementing a `RequestDecomposer<Req, Resp>` interface with two operations:
+
+* **Decompose**: given a request, produce a map from route name to the sub-request for that route.
+* **Recompose**: given the per-route responses, produce the merged response for the client.
+
+Bespoke per-API handling is unavoidable because each API has a different request/response shape and different merging semantics.
+For example, merging `METADATA` responses requires unioning broker lists by virtual node ID and taking the maximum of the response throttle times, 
+whereas merging `PRODUCE` responses is a straightforward concatenation of per-topic-partition results.
+
+Not every Kafka API key needs decomposition.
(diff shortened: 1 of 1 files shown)


proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,383 @@
...
+## Proposal
+
+### Overview
+
+The data virtualization router partitions Kafka traffic across multiple upstream clusters using two complementary strategies:
+
+**Topic-based partitioning.** 
+Each topic is owned by exactly one upstream cluster.
+The router determines the owning cluster from the topic name (via an unambiguous mapping) and routes data-plane operations accordingly.
+When a single client request references topics on different clusters (e.g. a `PRODUCE` batch containing records for topics on two clusters), the router decomposes the request, fans out the sub-requests, fans-in the sub-responses and the recomposes a single response to be returned to the client.
```
k-wall: "`PRODUCE` batch containing records for topics on two clusters".  When would this happen in Kafka?

Something is confused here.  With the virtualised node id scheme described in the Route API proposal every broker had a unique node id.  The client knows all the node ids. Right?

Producing Kafka clients work by connecting to the broker that is the topic leader.  So in the in the case where a client needs to send to two topic on different clusters, the client knows this and will be making two broker connections.

It will send a produce request for topic A to the first, and a produce request for topic B in the second.

The produce request that the router sees won't need to be decomposed because it will only contain topic partition request for the 'correct' broker.

Am I missing something?

tombentley: You're right that with the current dedicated (one-to-one) `BijectiveNodeIdMapping`, a well-behaved Kafka producer will not send a PRODUCE request spanning routes. Each virtual node belongs to exactly one route, the client batches by partition leader, and so each ProduceRequest naturally contains only topics from one route.
  
However, the Router API is designed to support different `NodeIdMapping` strategies (see proposal 070's "Node ID mapping implementation" section). The current dedicated mapping is one strategy (the first one we'd implement).

I wanted to have an API which didn't lock router implementation into _only_ working like that. I think it would be a strategic mistake to pour a lot of effort into implementing routers that could only work with a single kind of deployment if there's a relatively simple way to do something much more general. 

For example: We may want to support future strategies where a single virtual node may serve brokers from multiple routes — for example, if proxy instances were presented to clients as broker nodes. With such a shared (many-to-one) mapping, a single virtual node fronts brokers from different clusters, and a client's leader-directed PRODUCE request can contain topics from different routes.

The router doesn't know which mapping is in use — that's a runtime implementation detail. So it must handle the general case. Decomposition logic would ensure correctness regardless of the mapping strategy.

That said, your observation prompted a clarification this in the proposal — the overview now explains when decomposition is and isn't needed depending on the mapping, and proposal 070 now describes the dedicated and shared strategies concretely.

k-wall: Thank you. The explanation make sense. Having the API shaped in such a way to support the alternative pattern is useful.

It worries me that the Data Virtualisation Router PR resulting from this proposal will have code paths that we can't hit in production.  I mean it will only ever see requests Produce/Fetch requests that don't need to be decomposed - so it we'll have redundant code.  YAGNI tells us that is a the wrong approach.  We can enhance Data Virtualisation to support other NodeIdMapping later if the need arrives.

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+
+The following API keys are dynamically routed:
+
+| Category | API keys | Decomposition strategy |
+|----------|----------|----------------------|
+| **Data** | `PRODUCE` | Split by topic ownership. Records for each topic sent to the owning route. |
+| **Data** | `FETCH` | Split by partition leader (see _Leader caching_ below). |
+| **Data** | `LIST_OFFSETS` | Split by partition leader. |
+| **Metadata** | `METADATA` | Three variants: all-topics (fan out to all routes, filter responses to owned topics), empty-topics/broker-info-only (fan out to all routes for broker discovery), specific topics (split by ownership). Broker lists unioned by node ID. |
+| **Metadata** | `API_VERSIONS` | Forwarded to the specific broker via `sendRequestToNode(route, virtualNodeId, ...)`; response capped (see _Version capping_). |
```
k-wall: I don't understand how the Router implementation knows the node id in order to call RoutingContext#sendRequest

Neither the [Router](https://github.com/tombentley/kroxylicious-design/blob/84ef8af76667015048a3c62bae561aa16fed1fc0/proposals/070-routing-api.md#router) or [RouterContext](https://github.com/tombentley/kroxylicious-design/blob/84ef8af76667015048a3c62bae561aa16fed1fc0/proposals/070-routing-api.md#router) have sight of it. 

On the branch you have a version of the io.kroxylicious.proxy.routing.RoutingContext#sendRequest that doesn't require a node id.

k-wall: If the Router implementation is agnostic to the NodeIdMapping strategy, then surely the ApiVersions response needs to be a logical AND of the all the brokers.  I say this because we need to make sure the client sends requests that the can are intelligible to any of the brokers and the router can't predict in advance which broker(s) the channel will end up needing to talk to.

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+3. The router uses the names for routing decisions (which route owns each topic).
+4. The router forwards the request (with the original topic IDs) to the correct route's broker.
+
+The returned `CompletionStage<TopicNameMapping>` may contain failures (e.g. `UNKNOWN_TOPIC_ID` for an ID the cache has never seen and the upstream cluster doesn't recognise).
+The router must handle these gracefully, typically by returning an appropriate error response to the client.
+
+**Security consideration.**
+The topic ID resolution cache does not perform authorisation checks.
+A client that knows a topic ID can discover the corresponding topic name (and confirm the topic's existence) by submitting a request containing that ID, regardless of whether the client is authorised to access the topic.
+In practice we do not consider this a significant weakness: topic IDs are UUIDs with a vast keyspace that is neither guessable nor enumerable (for pratical purposes), so an attacker would need to have obtained a valid topic ID from another source (e.g. logs or a compromised client), at which point they likely already know the topic name.
```
k-wall:
```suggestion
In practice we do not consider this a significant weakness: topic IDs are UUIDs with a vast keyspace that is neither guessable nor enumerable (for practical purposes), so an attacker would need to have obtained a valid topic ID from another source (e.g. logs or a compromised client), at which point they likely already know the topic name.
```

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+2. The PID from the default route is returned to the client as _its_ PID.
+3. The per-route PID mappings are stored in a shared in-memory `ProducerIdManager` (shared across all connections to the same virtual cluster, thread-safe).
+4. On `PRODUCE`, the router rewrites the record batch headers (producer ID and epoch) for non-default routes before forwarding.
+
+Fanning out `INIT_PRODUCER_ID` to all routes means that PIDs are allocated on clusters that may never receive any records from that producer.
+This is a deliberate trade-off: it wastes PIDs on those clusters, but it allows the router to support idempotent producers (which are enabled by default since Kafka 3.0) using topic-based partitioning alone, without requiring subject-based routing.
+Subject-based routing is only needed for transactional producers, which have the stronger coordinator-affinity requirements described in the _Overview_.
+
+PID mappings are evicted after a configurable TTL (default: 7 days).
+If a mapping is evicted and a subsequent `PRODUCE` arrives, the router returns `UNKNOWN_PRODUCER_ID`, which causes the Kafka producer to re-initialise.
```
k-wall, line 169: I think this strategy works even if proxy replicas are scaled, right.  If a producing client reconnects any lands on a different proxy, it will follow the `UNKNOWN_PRODUCER_ID` path.  The flow is the same as for eviction case.

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+
+
+### KIP-227 fetch session management
```
k-wall, line 172: I think this section could use a bit more details.

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+
+The router implements _bidirectional session management_:
+
+* **Server-side** (facing the client): the router acts as a session server, tracking which topic-partitions the client has registered.
+* **Client-side** (facing the broker): the router acts as a session client for each backend route, independently managing sessions with each upstream cluster.
+
+These two sides are independent: a pre-v7 client (no session support) can still benefit from server-side sessions with backends, and vice versa.
+
+When a backend sends an _incremental_ response (containing only changed partitions), the router reconstructs a _full_ response using its cached session state before merging across routes.
+This ensures the client always receives a complete picture.
```
k-wall, line 185: Are you saying the client will always receive a full-response when using this Router?  Isn't that negaing KIP-227?

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+As described in the _Overview_, subject-based partitioning pins an authenticated user to a specific route for coordinator-bound operations.
+The configuration declares which subjects belong to which route (see _Configuration_ below).
+
+The following operations are forwarded to the subject's assigned route without decomposition:
+* `FIND_COORDINATOR` (group and transaction coordinators)
+* `ADD_PARTITIONS_TO_TXN`, `END_TXN`, `ADD_OFFSETS_TO_TXN`, `TXN_OFFSET_COMMIT`
+* `CONSUMER_GROUP_HEARTBEAT`, `CONSUMER_GROUP_DESCRIBE`
+* `OFFSET_COMMIT`, `OFFSET_FETCH` (when the subject is assigned to a route)
+
+**Prerequisite**: subject routing requires that the client's identity is known, which in practice means SASL termination must occur before the router in the topology.
```
k-wall, line 213: Why doesn't downstream TLS client auth work for Subject routing?

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+This avoids duplicating potentially large caches (the partition leader cache may have entries for every topic-partition) across every connection.
+The caches must be thread-safe since connections run on different event loop threads; a `ConcurrentHashMap` is sufficient given that reads dominate and writes (on leader change or coordinator migration) are infrequent.
+This follows the same pattern as the `ProducerIdManager` (see _Producer ID management_).
+
+Fetch session state, by contrast, is genuinely per-connection and lives in the `Router` instance (see _KIP-227 fetch session management_).
+
+
+### Assignment rejection
+
+`CREATE_TOPICS` and `CREATE_PARTITIONS` accept optional explicit broker assignments (replica placement lists that name specific broker node IDs).
```
k-wall, line 243: Shouldn't create_topic need reject topic creations where topic name would be unroutable?  Otherwise, the client would be allowed to create topics that would be invisible.

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+| **Metadata** | `METADATA` | Three variants: all-topics (fan out to all routes, filter responses to owned topics), empty-topics/broker-info-only (fan out to all routes for broker discovery), specific topics (split by ownership). Broker lists unioned by node ID. |
+| **Metadata** | `API_VERSIONS` | Forwarded to the specific broker via `sendRequestToNode(route, virtualNodeId, ...)`; response capped (see _Version capping_). |
+| **Metadata** | `FIND_COORDINATOR` | Routed to the appropriate route based on coordinator type and subject routing configuration. |
+| **Metadata** | `DESCRIBE_CLUSTER` | Fan out to all routes; merge broker lists. |
+| **Offsets** | `OFFSET_COMMIT`, `OFFSET_FETCH` | Split by topic ownership (non-subject-routed) or forwarded to assigned route (subject-routed). |
+| **Offsets** | `OFFSET_FOR_LEADER_EPOCH` | Split by partition leader. |
+| **Admin** | `CREATE_TOPICS` | Split by topic ownership; reject explicit broker assignments (see _Assignment rejection_). |
+| **Admin** | `DELETE_TOPICS` | Split by topic ownership. |
+| **Admin** | `CREATE_PARTITIONS` | Split by topic ownership; reject explicit broker assignments. |
```
k-wall, line 121: ALTER_PARTITION_REASSIGNMENTS should split by topic ownership. Same for LIST_PARTITION_REASSIGNMENTS  It also needs to validate that the request does not try to move partitions between clusters.  To be honest, I think it might be better to deliberately leave these out of scope for v1.

proposals/112-topic-based-routing.md
```diff
@@ -0,0 +1,389 @@
...
+
+Data virtualization can address this by presenting a virtual cluster that is composed of multiple real clusters.
+Clients connect to the virtual cluster and interact with it as if it were a single Kafka cluster.
+The proxy transparently routes each request to the correct upstream cluster based on which cluster owns the topics involved.
+
+The Kafka protocol makes this non-trivial.
+Many request types batch operations across multiple topics in a single request (e.g. a `PRODUCE` request may contain records for topics owned by different clusters).
+The router must decompose such requests, fan out the sub-requests in parallel, and recompose a single response.
+Beyond basic produce and consume, stateful protocol features — idempotent producers, transactions, consumer groups, fetch sessions — each add further complexity.
+This proposal addresses these features incrementally, with the simplest cases (produce, metadata) first and the more complex cases (transactions, consumer groups) building on that foundation.
```
k-wall, line 29: I presume we are deferring things like Share Consumers for another proposal.

k-wall:
Couple more comments, but otherwise heading in the right direction.

k-wall:
@tombentley please add a TL;DR to the PR's description.  Also fix up the PR's title to reflect the proposal's name.  Thank you.

proposals/112-topic-based-routing.md
tombentley: There are a lot of Kafka features which were omitted from the POC in the name of time, including share groups. They use topic ids, so there's a prereq to solve that. But otherwise I'm not aware of any showstoppers (which doesn't mean there aren't any). We could try to bite off more features as part of this proposal and its implementation (aiming for completeness), or we could prioritise timely delivery. The bottleneck will likely be the code review which we'd all need to feel comfortable that we had an understanding of the code.

k-wall: > They use topic ids, so there's a prereq to solve that. But otherwise I'm not aware of any showstoppers (which doesn't mean there aren't any). We could try to bite off more features as part of this proposal and its implementation (aiming for completeness), or we could prioritise timely delivery.

I'd favour the latter.  The sooner we get a release out with the feature, the sooner we get feedback on real use-cases.  There's also a chance we can decompose work on the missing parts into separate issues, that could be worked in parallel, perhaps by community members.

The proposal just need to enumerate what's out of scope.

proposals/112-topic-based-routing.md
tombentley: I think this has been address via the revisions to #70. Resolving this, but reopen if it's still unclear.

tombentley:
A big question for the topic router is whether it should continue to do the subject-based routing internally, or whether we should factor out a real subject-based router (currently there's one for use in the integration tests), and use a multi-router topology?

The background is that:
* Subject-based routing was what was used as the workaround for the lack of cross cluster transactions and consumer groups. 
*  the POC code was written before the API or runtime supported multiple routers. But the runtime now supports multiple routers.

Having subject-based routing built into the topic router feels a bit more conceptually complete and coherent, in terms of the functionality it's trying to provide. I think I lean more towards this view. But if the consensus is to force the end user to configure two separate routers to get the more complete functionality then I can revise the proposal.

PR #114, opened 2026-06-11
docs: scatter-gather routing API — an alternative to #70

SamBarker:
## Summary

An alternative routing proposal that builds on the foundation @tombentley laid in #70. This wouldn't exist without the thinking, experimentation, and POC work behind that proposal — thank you Tom for the effort and rigour you've put into it.

This proposal explores a different decomposition of the same problem. Where #70 gives the Router full ownership of the request lifecycle via `onRequest()`, this alternative narrows the Router to scatter-gather coordination only, with protocol-level concerns (namespace translation, identifier mapping, partition routing) handled by per-route filter chains and the runtime.

### Key differences from #70

- **Narrower Router API**: `scatter()` handles "which routes, how to compose" — protocol decomposition lives in exit filters written once by the runtime, not reimplemented in every Router
- **Router-scoped entry/exit filter positions**: entry filters resolve identifiers before scatter; exit filters handle per-route namespace translation after scatter. Each Router in a DAG gets its own entry and exit filters
- **Design principle**: each component operates in exactly one namespace and knows only its own concern. The Router is the scope boundary
- **`sendToRoute` + `sendToNode`**: two dispatch methods with architecturally distinct roles — `sendToRoute` for discovery (any broker), `sendToNode` for targeted dispatch (specific broker)
- **`BoundaryAwareRouterFactory`**: escape hatch for routers that need explicit control over their filter chains, motivated by external control-plane integration rather than common-case routing

### What's in the proposal

- The model: 1:1 as degenerate case, 1:N routing, DAG composition
- Design principle: one component, one namespace, one concern
- Plugin API: `Router`, `RouterFactory`, `BoundaryAwareRouterFactory`, `ScatterContext`, `ScatterGatherResult`
- Worked examples: AZ-aware fetch router (simple), union cluster router (full scatter-gather)
- Protocol identifier mapping as a uniform category (topicId, PIDs, fetch sessions)
- Configuration with router-targeting-router for DAG composition
- Contrast table with #70
- Rejected alternatives

### DIFF
--- a/proposals/000-scatter-gather-routing-api.md
+++ b/proposals/000-scatter-gather-routing-api.md
@@ -94,6 +94,16 @@ Each route gets its own filter chain: user-configured filters (e.g. topic name r
 SASL initiator for upstream authentication) followed by runtime-managed implicit filters
 (node ID mapping, broker address rewriting, API version intersection).
 
+Each route has two distinct ID spaces: **downstream** and **upstream**. The downstream
+space is what the layer above (client or parent router) sees. The upstream space is what
+the next layer down sees — which may be a backing cluster's real IDs or another router's
+virtual IDs. In a nested topology (`Client → Router A → route ra1 → Router B → route rb2
+→ Cluster X`), ra1's upstream IDs are Router B's downstream IDs, not Cluster X's real
+broker IDs. The mapping is always between adjacent layers. Each route maintains its own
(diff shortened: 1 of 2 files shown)


SamBarker:
I think 70 allows us to build a lot of the things this proposal was outlining.

PR #115, opened 2026-06-19
All-in-one Install YAML Manifests for GitOps

k-wall:
This proposal addresses kroxylicious/kroxylicious#4016 (option 1) by publishing fully-rendered Kubernetes installation manifests as GitHub release assets.

## Summary

Enable GitOps workflows by shipping single-file YAML manifests that can be directly referenced via URL, following patterns established by Strimzi and cert-manager.

## New Artifacts

All artifacts follow Maven classifier naming conventions (`artifactId-version-classifier.extension`):

**Per component (operator and admission webhook):**
- `{component}-dist-{version}-install.yaml` - Full install (CRDs + resources)
- `{component}-dist-{version}-crds.yaml` - CRDs only (for separate lifecycle)
- `{component}-dist-{version}-examples.tar.gz` - Examples (separate from install)
- `{component}-dist-{version}-examples.zip` - Examples (ZIP format)

**Maven coordinates:**
- Install: `io.kroxylicious:{component}-dist:install:yaml`
- CRDs: `io.kroxylicious:{component}-dist:crds:yaml`
- Examples: `io.kroxylicious:{component}-dist:examples:{tar.gz,zip}`

## Key Features

- All template variables fully substituted
- Direct `kubectl apply -f https://...` installation
- Two-stage install option (CRDs first, then operator)
- Automatic GPG signing via maven-gpg-plugin
- Proper Maven coordinates for deployment

## Migration Path

Existing `.tar.gz`/`.zip` archives (install + examples) will be deprecated and removed in a future release (at v1.0.0 or earlier).

Users migrate to separate install manifests + examples archives.

Relates-to: kroxylicious/kroxylicious#4016

### DIFF
--- a/proposals/115-all-in-one-install-yaml.md
+++ b/proposals/115-all-in-one-install-yaml.md
@@ -14,18 +14,18 @@ Publish all-in-one Kubernetes installation manifests as GitHub release assets to
 ## Current situation
 
 Kroxylicious publishes Kubernetes operator and admission webhook as `.tar.gz` and `.zip` archives containing:
-- Installation manifests with unsubstituted template variables (e.g., `$[io.kroxylicious.operator.image.name]`)
+- Installation manifests (multiple files in a directory)
 - Example configurations
 - Documentation
 
 Users must:
 1. Download the release archive
 2. Extract the contents
-3. Locate the install directory with rendered manifests
+3. Locate the install directory with manifests
 4. Manually commit these to their GitOps repositories
 5. Repeat this process for every release
 
-This workflow is incompatible with GitOps tools (Flux CD, Argo CD) that pull manifests directly from Git or URLs.
+The archive format creates friction and is incompatible with GitOps tools (Flux CD, Argo CD) that expect to pull manifests directly from URLs or Git repositories.
 
 ## Motivation
 
--- a/proposals/115-all-in-one-install-yaml.md
+++ b/proposals/115-all-in-one-install-yaml.md
@@ -142,11 +142,11 @@ Examples move to dedicated archives separate from install manifests:
 
 ### Backwards Compatibility
 
-**Preserved:** Existing `.tar.gz` and `.zip` archives continue to be published for 2 releases (v0.22.0, v0.23.0).
+**Preserved:** Existing `.tar.gz` and `.zip` archives continue to be published alongside the new manifest artifacts.
 
-**Deprecated:** These combined archives marked as deprecated in release notes.
+**Deprecated:** The deprecation of these combined archives will be announced in the CHANGELOG of the next release (at time of writing v0.23.0).
 
-**Removed:** Archives removed in v0.24.0 (gives users 2 releases to migrate).
+**Removed:** Archives will be removed following the project's [standard deprecation policy](https://github.com/kroxylicious/kroxylicious/blob/main/DEV_GUIDE.md#deprecation-policy) unless preempted by the v1.0.0 release.
 
 **Migration path:** Documentation provides clear instructions for switching from archive-based to manifest-based installation.
 
@@ -164,13 +164,7 @@ Examples move to dedicated archives separate from install manifests:
 
 **Timeline:**
 
-Combined archives will be deprecated immediately in v0.22.0 and removed at the **v1.0.0 milestone** (whichever comes first: v0.24.0 or v1.0.0).
-
-| Release | Archive Status | Manifest Status | Notes |
-|---------|---------------|-----------------|-------|
-| v0.22.0 | ⚠️ **Deprecated** - Published with deprecation notice | ✅ **New** - Full support | Introduction release. Release notes highlight new manifests as recommended approach. |
-| v0.23.0 | ⚠️ **Deprecated** - Published | ✅ **Supported** | Deprecation notice continues in release notes. |
-| v0.24.0 or v1.0.0 | ❌ **Removed** - No longer published | ✅ **Supported** | Archives removed at v1.0.0 milestone or v0.24.0, whichever comes first. |
+Combined archives will be deprecated in the next release (at time of writing v0.23.0). Removal follows the project's [standard deprecation policy](https://github.com/kroxylicious/kroxylicious/blob/main/DEV_GUIDE.md#deprecation-policy): eligible for removal in the third minor release following deprecation, with at least three months elapsed, unless preempted by the v1.0.0 release.
 
 **What gets deprecated:**
 - `kroxylicious-operator-{version}.tar.gz` (combined install + examples)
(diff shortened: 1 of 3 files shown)


MisterTimn:
This would definitely meet our needs with regards to GitOps deployment, we already deploy cert manager and strimzi operator in the same manner.

proposals/115-all-in-one-install-yaml.md
```diff
@@ -0,0 +1,262 @@
...
+No template markers remain in published manifests.
+
+### Installation Workflows
+
+**Single-step installation:**
+```bash
+kubectl apply -f https://github.com/kroxylicious/kroxylicious/releases/download/v0.22.0/kroxylicious-operator-dist-0.22.0-install.yaml
+```
+
+**Two-stage installation (CRDs first):**
```
SamBarker, line 101: The two-stage installation workflow implies a split-privilege scenario (cluster-admin installs CRDs, namespace-admin installs the operator). But `install.yaml` is defined as containing CRDs, so the namespace-admin would be re-applying cluster-scoped resources they likely don't have permissions for.

If two-stage is genuinely for split-privilege, I think the artifact split needs to follow the privilege boundary rather than the content type:

- `install.yaml` — everything (for the simple single-admin case, as proposed)
- `cluster-admin.yaml` — CRDs, ClusterRoles, ClusterRoleBindings, and any other cluster-scoped resources
- `namespace-admin.yaml` — Deployment, Services, namespace-scoped RBAC, etc.

This diverges from cert-manager's `crds.yaml` naming, but it's clearer about intent and forces us to think about where RBAC resources land — not just "CRDs vs the rest" but "cluster-scoped vs namespace-scoped."

If two-stage isn't intended for split-privilege, what's the use case? Ordering alone doesn't seem to justify it since `install.yaml` already orders CRDs first.

proposals/115-all-in-one-install-yaml.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Compatibility
+
+### Backwards Compatibility
+
+**Preserved:** Existing `.tar.gz` and `.zip` archives continue to be published for 2 releases (v0.22.0, v0.23.0).
+
+**Deprecated:** These combined archives marked as deprecated in release notes.
+
+**Removed:** Archives removed in v0.24.0 (gives users 2 releases to migrate).
```
SamBarker: nit: The deprecation timeline is stated slightly differently in three places:

- Here: "removed in v0.24.0 (gives users 2 releases to migrate)"
- Line 167: "removed at the v1.0.0 milestone (whichever comes first: v0.24.0 or v1.0.0)"
- Line 225: "Rationale for 3-release timeline"

The thrust is right — deprecate in v0.22.0, remove at v0.24.0 or v1.0.0 (whichever is first). But the three descriptions don't quite agree on whether it's 2 or 3 releases of notice, and this section doesn't mention the v1.0.0 accelerator. Worth a consistency pass.

proposals/115-all-in-one-install-yaml.md
```diff
@@ -0,0 +1,262 @@
+<!--
+To submit this proposal:
+1. Copy to kroxylicious/design repo as proposals/000-rendered-install-manifests.md
+2. Open a PR
+3. Rename file using PR number: proposals/<PR#>-rendered-install-manifests.md
+4. Update heading to include PR number
+5. Push changes
```
SamBarker: nit: Template submission instructions still present — should be removed before merge.

SamBarker (approved):
LGTM

Thanks @k-wall I think the basic premiss is good, the two stage install is a wrinkle but I've come round to the idea of a single `kubectl apply` for install

proposals/115-all-in-one-install-yaml.md
k-wall: The truthful answer is I included a separate CRDs artefact as that's what CertMgr and Strimzi do already.   I followed the herd.  I didn't give it too much thought.

Your comment prompted me to look a bit deeper.   The motivation seems to be not one of split privilege.  Operators often require cluster wide resources (ClusterRoles, for instance) and the installation of these requires cluster admin privileges.

The real motivation appears to be the fact that CRDs are really tricky.  

https://helm.sh/community/hips/hip-0011/#the-core-problem says:

> The core problem is that CRDs (being a globally shared resource) are fragile. Once a CRD is installed, we typically have to assume (all other things being equal) that it is shared across namespaces and groups of users.
>For that reason, installing, modifying, and deleting CRDs is a process that has ramifications for all users and systems of that cluster.
>  In spite of this, the Kubernetes API server is permissive about these CRDs. CRDs are mutable (even without a version change). When they are deleted, all instances of them are deleted without warning. They can be upgraded to be entirely incompatible with previous versions. And there is no way to programmatically inspect the CRDs in a cluster and determine whether they are used, how they are used, and by what.

and then goes on to explain why Helm treats CRDs separately https://helm.sh/community/hips/hip-0011/#installing-crds

ArgoCD [discusses the same issues](https://kubernetes.recipes/recipes/deployments/argocd-sync-waves-crd-operators/).  Its solution is `sync waves` that deploy Custom Resource Definitions before operators and custom resources that aim to prevent CRD race conditions in GitOps pipelines. 

One specific use-case for separate CRDs bundles is that it allow the deployment of a complex system of operators. Imagine a deployment of Operator B creates an instance of Operator A's CRD.  If you try to install (`kubectl apply`) Operator B before A, it will partially fail.   If A and B ship separate CRDs bundles then a Kube admin can apply all the CRD bundles first, and then install the operators in any order.

So whilst I don't claim to have a good grasp of all the issues, I do think that shipping yaml in the same manner as often used adjacent Operators is likely to make our uses life easier.

proposals/115-all-in-one-install-yaml.md
k-wall: Agreed. Claude got all wordy.  I'll simply things to reference our current deprecation policy and talk about how the arrival of 1.0.0 would trump things.

k-wall:
@SamBarker I responded to your comments. I wanted to give you chance to see my responses.  Please merge if you are happy.

proposals/115-all-in-one-install-yaml.md
SamBarker: Thanks for digging into this. I'll be honest — I still don't fully see the advantage of the two-stage install for our case, but if that's what Strimzi and cert-manager ship then matching the pattern makes sense for users who already have workflows built around it.

For posterity: the comment on line 106 (`# Namespace-admin installs operator`) is misleading since `install.yaml` contains CRDs and other cluster-scoped resources too. If we revisit this, the install workflow docs should reflect the actual intent (CRD ordering, not split privilege).

SamBarker (approved):

PR #116, opened 2026-06-23
[proposal] RequestData|ResponseData API migration  

piotrpdev:
## Summary

Eliminate Kroxylicious's exposure to Kafka internal package re-organisations by taking full source ownership of the Kafka protocol classes, delivering a stable public API surface before 1.0.

## Approach

Kafka's `MessageGenerator` (~35 files) and the non-generated support classes (`protocol.*`, `record.*`, scattered `common.*` types) are copied into the Kroxylicious repo as source under `io.kroxylicious.kafka.*`, retaining their original Apache 2.0 licence headers. The generator is run at build time against the upstream IDL specs (bundled in `kafka-clients`) to produce the `*Data` classes:

- **Source ownership from day one:** every upstream change is a deliberate choice - review, absorb, or skip
- **Generator enhancements** (type-safe request-response pairing, `aliases`, Javadoc from IDL) are first-class commits, not patches
- **`japicmp` drift detection:** `kafka-clients` is shaded into a throwaway reference artifact with matching package names; `japicmp` then reports any API divergence between our classes and upstream
- **Automated upgrades:** whenever dependabot opens a PR bumping the `kafka-clients` version; a new GitHub Workflow regenerates `*Data` classes and runs fidelity tests and `japicmp` on the PR, reporting any incompatibility immediately

## Key Benefits

- KAFKA-20128-style re-organisations are handled by updating the copied source - filter developers see no change                                                                                                                                                                               
- Single breaking change (one import migration) rather than repeated breakage per Kafka release
- Wire compatibility proven continuously by fidelity tests across all 198 specs and every supported protocol version
- No bytecode transformation dependency to eventually unwind - the source is ours to replace incrementally with Netty-native implementations

## Migration

Filter developers update all `org.apache.kafka.*` imports to `io.kroxylicious.kafka.*`. A `sed`-based migration script and package mapping guide will be provided alongside the change.

Relates-to: kroxylicious/kroxylicious#3996


proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+For a 1.0 release, Kroxylicious needs a stable public API surface that the project controls. The Kafka protocol wire format itself is stable (governed by KIPs), but the Java classes representing that format are not.
+
+## Proposal
+
+### Scope: own the full API surface in one move
+
+Every type that appears in Kroxylicious's public API should live in a Kroxylicious namespace. This means a single breaking change for filter developers - one import migration - rather than a drip of repeated breakage as Kafka reorganises its internals.
+
+"Own the surface" does not mean "rewrite everything from scratch." The shaded and generated artifacts redistribute `kafka-clients` bytecode under the same Apache 2.0 licence. The original Apache Software Foundation copyright notices are preserved in the bytecode. The underlying implementations can be replaced with Kroxylicious-native ones incrementally, gated by the verification harness proving each replacement is correct.
```
k-wall: nit: the licence is comment in the source, so does not end up in the bytecode.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+- **`ApiKeys`** - the enum that maps API key numbers to names and version ranges.
+- **Scattered `common.*` types** - `Uuid`, `ApiException` subclasses, ACL types, SASL classes, utilities.
+
+Kroxylicious's `ByteBufAccessor` currently `implements Readable, Writable` from Kafka, bridging Netty's `ByteBuf` to Kafka's I/O abstraction. Any namespace change requires updating this bridge.
+
+The `kroxylicious-krpc-plugin` already parses the JSON IDL specs independently using Jackson (without depending on Kafka's generator) and drives FreeMarker templates to generate filter interfaces, invokers, and decoders. Its only Kafka dependency is `ApiKeys`, used in 3 places.
+
+## Motivation
+
+Kafka 4.3.0 moved 37 classes from `org.apache.kafka.common.record.*` to `org.apache.kafka.common.record.internal.*` ([KAFKA-20128](https://issues.apache.org/jira/browse/KAFKA-20128)). Upgrading Kroxylicious to 4.3.0 required import changes across 80 Java files in 9 modules. This is a symptom of a deeper problem: Kroxylicious depends on Kafka internal APIs that Kafka is free to move or remove at any time.
```
k-wall: It is not just moves and removals that are problematic.  Changes in methods/method signatures API would be equally impactful.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+Kafka does not consider these packages to be stable public API. The project has no obligation to maintain backward compatibility, and further reorganisations are likely. Each one forces a breaking change on every Kroxylicious filter developer - internal and external.
```
k-wall: By **internal** I think you you mean this project itself and the filters it ships as part of Kroxylicious.  By **external**, I think you mean the **Filter authors** i.e. users writing their own Filters against the Filter API implementing their own business logic.
 
I don't really think "internal" is really the concern.  If we didn't have a public Filter API, we could just swallow the Kafka changes and it would all be beneath the abstraction.  If we cared at all, we wouldn't need a proposal because no public API would be changd.

I think the the motivation for this change is the second group you called external.  We've defined the persona name as 'Filter Authors'.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Scope: own the full API surface in one move
+
+Every type that appears in Kroxylicious's public API should live in a Kroxylicious namespace. This means a single breaking change for filter developers - one import migration - rather than a drip of repeated breakage as Kafka reorganises its internals.
+
+"Own the surface" does not mean "rewrite everything from scratch." The shaded and generated artifacts redistribute `kafka-clients` bytecode under the same Apache 2.0 licence. The original Apache Software Foundation copyright notices are preserved in the bytecode. The underlying implementations can be replaced with Kroxylicious-native ones incrementally, gated by the verification harness proving each replacement is correct.
+
+### Approach: shade `kafka-clients` via a dedicated repo
+
+The Maven Shade Plugin takes compiled classes from `kafka-clients` and republishes them under a Kroxylicious namespace. All cross-references between relocated classes are rewritten automatically at the bytecode level. This means Kafka package reorganisations - including moves to an `internal` subpackage like KAFKA-20128 - are completely invisible.
```
k-wall: Shading can handle repackaging, but AFAIK, can't handle things like method renames or method signature changes.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+A dedicated repo in the Kroxylicious GitHub org (`kroxylicious/kafka-patches`) stores the shading configuration and any generator patches needed for future enhancements. When generator enhancements are implemented, CI will apply the patches against the Kafka generator source, run the patched generator to produce enhanced `*Data` source, compile, and shade the result alongside the rest of `kafka-clients`.
```
k-wall: Not sure if I have understood this right.  I think the kafka-patches exist to allow us to add features to the Kafka code generator (field aliasing, for example).  Have I groked this right?  


Patching can get tricky and I'd worry about having it  in our foundations.  

I prefer the approach of forking the generator once and owning the forked code going forward.  We'd rely on the serialisation/deserialisation fidelity tests to catch regressions between ourselves and Kafka.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+Published artifact versions use the Kafka version as a base with a qualifier suffix (e.g. `4.3.0-kroxy-0`, `4.3.0-kroxy-1` for Kroxylicious-only fixes, `4.4.0-kroxy-0` when bumping to a new Kafka release). This makes the Kafka baseline immediately visible in any dependency declaration.
+
+The `kroxylicious-krpc-plugin` currently depends on `kafka-clients` solely for `ApiKeys`. After this change it depends on the new artifact instead. Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.
```
k-wall: Presumably, we could codegen a ApiKeys from the RPC definitions themselves.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+#### Wire-level round-trip tests
+
+For every spec and every supported protocol version: serialise a message with our generated classes, deserialise with Kafka's, and assert byte-level equality - then reverse the direction. This is the ground truth. If bytes we produce are consumed correctly by Kafka's classes and vice versa, we are wire-compatible.
+
+Exploratory work on the [`piotrpdev/kroxylicious#refactor/generate-data-class`](https://github.com/piotrpdev/kroxylicious/tree/refactor/generate-data-class) branch built a prototype of this infrastructure covering all 198 specs: parameterised fidelity tests at every valid version, randomised field-value tests, jqwik property-based tests for full-depth message instances, and a compile-time spec count guard. That prototype was designed around the FreeMarker template approach (see ["Rejected Alternatives"](#rejected-alternatives)), where generated classes were compiled at test time via `ToolProvider.getSystemJavaCompiler()` and loaded via `URLClassLoader`. With the shading approach the classes are already on the classpath at build time, so only the class-loading mechanism changes - the verification logic itself carries over unchanged.
```
k-wall: Where a field is nullable, we'd need to test the null case too.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+A dedicated repo in the Kroxylicious GitHub org (`kroxylicious/kafka-patches`) stores the shading configuration and any generator patches needed for future enhancements. When generator enhancements are implemented, CI will apply the patches against the Kafka generator source, run the patched generator to produce enhanced `*Data` source, compile, and shade the result alongside the rest of `kafka-clients`.
+
+Published artifact versions use the Kafka version as a base with a qualifier suffix (e.g. `4.3.0-kroxy-0`, `4.3.0-kroxy-1` for Kroxylicious-only fixes, `4.4.0-kroxy-0` when bumping to a new Kafka release). This makes the Kafka baseline immediately visible in any dependency declaration.
+
+The `kroxylicious-krpc-plugin` currently depends on `kafka-clients` solely for `ApiKeys`. After this change it depends on the new artifact instead. Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.
+
+### Shading configuration
+
+The shading configuration in `kroxylicious/kafka-patches` uses a single broad relocation rule:
```
k-wall: So would we release kafka-patches in lock step with kafka releases.  If a kafka release is made which doesn't change wire level API (e.g. 4.3.1) would we still release it?

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+1. Queries Maven Central's search API for the latest release of `org.apache.kafka:kafka-clients`.
+2. Compares the result against the current `kafka-clients` version in `kroxylicious/kafka-patches`.
+3. If a newer version exists and no open PR already targets it, opens a PR bumping the version.
+4. CI runs on the PR - rebuilding the shaded artifact and running the fidelity tests against the new `kafka-clients`.
+
+If CI passes, the PR can be reviewed and merged with minimal effort. If CI fails, the PR surfaces exactly what broke - a fidelity test failure identifies a wire incompatibility; an `EXPECTED_SPEC_COUNT` failure flags a spec change requiring review. Either way, the problem is visible and actionable immediately, before anyone has to manually discover the new release.
+
+#### Drift in shaded classes
+
+With shading, upstream changes to hand-coded classes like `MemoryRecords` and `RecordBatch` are absorbed automatically when `kafka-clients` is bumped. Drift is detected through two complementary mechanisms:
```
k-wall: If Kafka changed its MemoryRecord API, we wouldn't necessarily want to ripple the change into our public API surface.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Fork the `apache/kafka` repository into the Kroxylicious GitHub org
+
+Maintaining a `kroxylicious/kafka` fork with Kroxylicious-specific commits on top of upstream. This gives a standard Git workflow (edit files, commit, push) and IDE support, and Git's merge/rebase machinery handles context drift more gracefully than patches do. However, it adds a repository to the org that must be kept in sync, the fork's diff from upstream is implicit rather than inspectable, and there is a risk of the fork accumulating unintentional drift. Storing generator patches in `kroxylicious/kafka-patches` achieves the same generator-modification capability with less overhead.
```
k-wall: When we were talking about forking, I understood that to mean 'just' the generator and the support classes (MemoryRecords, ApiMessage etc).  I don't think any of us imagined forking the whole kafka repo.

proposals/116-kafka-api-migration.md
piotrpdev:
```suggestion
"Own the surface" does not mean "rewrite everything from scratch." The shaded and generated artifacts redistribute `kafka-clients` bytecode under the same Apache 2.0 licence. The original Apache Software Foundation copyright notices are preserved. The underlying implementations can be replaced with Kroxylicious-native ones incrementally, gated by the verification harness proving each replacement is correct.
```

proposals/116-kafka-api-migration.md
piotrpdev:
```suggestion
Kafka 4.3.0 moved 37 classes from `org.apache.kafka.common.record.*` to `org.apache.kafka.common.record.internal.*` ([KAFKA-20128](https://issues.apache.org/jira/browse/KAFKA-20128)). Upgrading Kroxylicious to 4.3.0 required import changes across 80 Java files in 9 modules. This is a symptom of a deeper problem: Kroxylicious depends on Kafka internal APIs that Kafka is free to change, move or remove at any time.
```

k-wall:
Thanks @piotrpdev. The shading idea is certainly an interesting approach, but I don't think we should go that way.  Having the project's foundation relying on shading would worry me.

I would:

1. fork the generator and support classes (ApiMessage, MemoryRecords etc) into kroxylicious.  Literally copy the sources into the kroxylicious repo.  Maintain the original licence on the sources so people understand the heritage.  
2. The APIs for the imported classes (ApiMessage, MemoryRecords) and generated Data becomes our public API.  At this point, we've severed the link to the Kafka internal classes completely.   It is now our code.
3. Implement serialisation/deserialisation tests proving wire level interop with that written by Kafka Client.

If Kafka choose to refactor say MemoryRecords, we probably don't care.  We don't necessarily need to take the changes into our version.  We might, but we don't have to.  The important thing is our MemoryRecords has a stable public API and it is interoperable.

@SamBarker @robobario WDYT?

proposals/116-kafka-api-migration.md
piotrpdev: That's where we could use the patches to rename methods or add overloads for them

proposals/116-kafka-api-migration.md
piotrpdev: By "forking" I thought we meant GitHub forking, apologies. Wouldn't forking the whole repo on GitHub allow us to track changes in the files we want more easily?

proposals/116-kafka-api-migration.md
piotrpdev: I imagine most changes would be enhancements and fixes, no? I thought we would want to absorb most of the stuff, and we could add patches for reverting what we don't want.

proposals/116-kafka-api-migration.md
piotrpdev: Presumably, the proposed GitHub workflow creates a PR bumping the version, and all the tests pass straight away, so we just merge it and have another workflow automatically create the patch release. If there are conflicts, I suppose we could optionally defer until a minor release if we're busy?

proposals/116-kafka-api-migration.md
piotrpdev:
```suggestion
Exploratory work on the [`piotrpdev/kroxylicious#refactor/generate-data-class`](https://github.com/piotrpdev/kroxylicious/tree/refactor/generate-data-class) branch built a prototype of this infrastructure covering all 198 specs: parameterised fidelity tests at every valid version, randomised field-value tests (including explicit coverage of `null` for nullable fields), jqwik property-based tests for full-depth message instances, and a compile-time spec count guard. That prototype was designed around the FreeMarker template approach (see ["Rejected Alternatives"](#rejected-alternatives)), where generated classes were compiled at test time via `ToolProvider.getSystemJavaCompiler()` and loaded via `URLClassLoader`. With the shading approach the classes are already on the classpath at build time, so only the class-loading mechanism changes - the verification logic itself carries over unchanged.
```

proposals/116-kafka-api-migration.md
piotrpdev: > I think the kafka-patches exist to allow us to add features to the Kafka code generator...

Yes, they would also allow us to bypass the upstream changes we don't like (e.g. we add overloads for functions where the signature has changed).

> I prefer the approach of forking the generator...

I worry that if we keep excluding upstream changes that don't affect the (de)serialized output (e.g. fixes/refactorings), eventually there may be an upstream change that both does affect the output and depends on a series of previous non-output-affecting changes, which would make the update significantly harder. I believe the proposed shading + patches approach helps mitigate this (and drift) in the best way.

proposals/116-kafka-api-migration.md
piotrpdev:
```suggestion
The `kroxylicious-krpc-plugin` currently depends on `kafka-clients` solely for `ApiKeys`. After this change it depends on the new artifact instead (or we could codegen an `ApiKeys` from the RPC definitions themselves). Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.
```

proposals/116-kafka-api-migration.md
piotrpdev:
```suggestion
Kafka does not consider these packages to be stable public APIs. The project has no obligation to maintain backward compatibility, and further reorganisations are likely. Because these unstable Kafka classes are exposed through our public API, we cannot simply absorb these changes beneath our abstraction layer. Instead, every upstream reorganization forces a breaking change directly onto **Filter Authors** who rely on the Filter API to implement their own custom business logic.
```

proposals/116-kafka-api-migration.md
piotrpdev:
```suggestion
The Maven Shade Plugin takes compiled classes from `kafka-clients` and republishes them under a Kroxylicious namespace. All cross-references between relocated classes are rewritten automatically at the bytecode level. This means Kafka package reorganisations - including moves to an `internal` subpackage like KAFKA-20128 - are completely invisible. Breaking code changes such as method renames or signature changes are addressed by adding patches that overload the methods or wrap them appropriately.
```

piotrpdev:
> Literally copy the sources into the kroxylicious repo. 

Just some food for thought, one of my comment responses:

> I worry that if we keep excluding upstream changes that don't affect the (de)serialized output (e.g. fixes/refactorings), eventually there may be an upstream change that both does affect the output and depends on a series of previous non-output-affecting changes, which would make the update significantly harder. I believe the proposed shading + patches approach helps mitigate this (and drift) in the best way.

@tombentley any thoughts :eyes: ?

SamBarker:
Thanks for the thorough proposal @piotrpdev, and for the prototype work backing it up. The problem statement is spot on — we need to own this namespace before 1.0, and the verification harness you've designed is valuable regardless of which approach we take.

Reading through the discussion with @k-wall, I find myself coming from a similar space to Keith — my instinct is toward owning the source rather than owning a bytecode transformation. But I also see genuine value in the automatic upstream tracking that shading provides, and I don't think either of the two models I'm about to describe is exactly what's proposed in the document or what Keith suggested. I want to set them out side by side so we can have a clear conversation about the trade-offs and figure out which direction fits us best.

## On patches

You'll notice neither model below includes the patch-based workflow from the proposal. Maintaining patches against Kafka's generator is a cunning solution to avoid forking it, but I don't think it pulls its weight. If we can detect drift with tooling (japicmp + fidelity tests), we don't need to stay mechanically coupled to upstream's source tree. When we need to change the generator, we fork it and own it — cleanly, not through a patch queue.

## The destination

I think we all agree that Kroxylicious needs to own the namespace and own the source. Personally, I'd also like us to eventually be in a position to replace key classes (like `MemoryRecords` and `RecordBatch`) with implementations that properly integrate with Netty's `ByteBuf` — but that's my longer-term aspiration. Either way, the question is about the journey — what are the intermediate states, and what does each step cost?

Both models below reach full ownership. They differ in what order they get there and what you're carrying at each stage.

## Model A: Shade first, own incrementally

1. **Now:** Shade `kafka-clients` in-tree (mono-repo module, not a separate repo) under `io.kroxylicious.kafka.*`. One shade rule, no source to maintain. Automatic absorption of upstream changes on each Kafka bump.
2. **When we need generator changes:** Fork the generator into the mono-repo — the same thing Model B does in step 1, just deferred. The ~50 non-generated classes can remain shaded or be copied in too. The question is whether deferring this fork actually saves us anything, or whether we've just delayed the same work.
3. **When we're ready:** Replace the remaining shaded classes with Kroxylicious-native implementations (Netty `ByteBuf` integration, etc.). The shaded jar shrinks and eventually disappears.

**What it gives us:** The cheapest possible step 1 — namespace ownership with almost no source to maintain. Automatic upstream tracking until we choose to take over each class.

**What it costs:** We carry a bytecode transformation as our foundation until step 3. And step 2 is the same work as Model B's step 1 — deferred, not avoided.

**What it gets right:** Non-wire behavioural changes (bug fixes, validation improvements, etc.) are absorbed automatically. We're not trying to fork Kafka — we want to be additive to upstream. We own the namespace and the wire contract, and we layer our own enhancements on top (e.g. field aliasing to maintain API compatibility when upstream renames things).

## Model B: Own the source from the start

1. **Now:** Fork Kafka's `MessageGenerator` into the mono-repo. Run it against the upstream IDL specs to produce `*Data` source under our namespace. Copy in the non-generated classes (`protocol.*`, `record.*`, scattered `common.*` types) as source.
2. **When we need generator changes:** They're just commits — type-safe pairing, Javadoc, aliases. No patches, no layering.
3. **When we're ready:** Replace buffer/record classes with Netty-native implementations. Same as Model A's step 3, but there's no shading to unwind — we already own the source.

**What it gives us:** Full source ownership from day one. Every upstream change is a deliberate choice — review, cherry-pick, skip. Generator enhancements and buffer replacements are the same kind of work (modify source we own). Uniform handling of all change types.

**What it costs:** More manual effort to absorb upstream changes. Risk of drift if we neglect upstream tracking — though tooling mitigates this (see below).

## Drift detection for either model

`japicmp-maven-plugin` compares by fully-qualified class name, so we can't point it directly at our jar and `kafka-clients` — the package names won't match. But we can shade `kafka-clients` with the same `org.apache.kafka` → `io.kroxylicious.kafka` relocation into a throwaway reference artifact, then run japicmp against that. Both jars have matching package names and japicmp tells us exactly which APIs have diverged from upstream.

Combined with the wire-level fidelity tests from the proposal, this gives us two complementary signals: japicmp catches API-level divergence, fidelity tests catch wire-format divergence. We already use japicmp in the main build, so the plugin infrastructure is there — it's just a matter of configuring a comparison module. As long as we have wire compatibility, we're insulated from the need to absorb every API change as it happens.

## The sequencing question

The real question isn't which model is "right" — it's which sequence of intermediate states we're comfortable carrying. Model A gives the cheapest step 1 but means step 2 is a transition (from shading to source ownership for the generated classes). Model B front-loads more work but every subsequent step is the same kind of work: modify source we own.

I'm also not sure how much step 1 of Model A actually saves over Model B. The non-generated classes (~50 or so) are relatively stable — copying them in as source with japicmp tracking drift might be nearly as cheap as shading them, without the bytecode-transformation dependency.

What do people think? I'm genuinely open — I see the appeal of both paths and I'd like to hear where others land.

SamBarker:
We've been sitting without any progress for a few weeks, so I want to give a clearer steer rather than leaving the question open.

My previous comment deliberately laid out Model A and Model B as neutral alternatives to start a debate. However, since there's been no further input from anyone, I'm going to take a position.

Model A's real advantage is speed: less upfront work, faster to absorb a new Kafka release. That's a genuine benefit and I wanted to give space for others to argue for it. But in the absence of that argument, I think Model B is the stronger foundation. It gives us uniform handling of all change types from day one, avoids carrying a bytecode transformation we'll eventually unwind anyway, and generator enhancements are enhancements we can now make rather than patches layered on top.

**My recommendation: Model B — own the source from the start.**

Concretely, that means:
- Fork the `MessageGenerator` into the mono-repo
- Copy in the non-generated support classes as source under our namespace
- Use `japicmp` + wire-level fidelity tests for drift detection
- Generator enhancements (type-safe pairing, field aliases, Javadoc) are enhancements we can now make

I realise this is a significant change to the proposal's direction. Would you be open to reworking it along these lines, @piotrpdev? Happy to talk through it on a call if that would help — reshaping a proposal after review is normal and this has been a really valuable exploration of the problem space.

piotrpdev:
@SamBarker done in [`79a1796` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/79a179672c4e9c68924be9237e4ef6cabf60666c) :+1:

tombentley:
@piotrpdev can you update the PR description to describe the current approach?

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+For a 1.0 release, Kroxylicious needs a stable public API surface that the project controls. The Kafka protocol wire format itself is stable (governed by KIPs), but the Java classes representing that format are not.
+
+## Proposal
+
+### Scope: own the full API surface in one move
+
+Every type that appears in Kroxylicious's public API should live in a Kroxylicious namespace. This means a single breaking change for filter developers - one import migration - rather than a drip of repeated breakage as Kafka reorganises its internals.
+
+"Own the surface" does not mean "rewrite everything from scratch." The copied source files retain their original Apache Software Foundation copyright headers - the ASF remains the author of the underlying code, and these files are redistributed under the same Apache 2.0 licence. The underlying implementations can be replaced with Kroxylicious-native ones incrementally, gated by the verification harness proving each replacement is correct.
```
tombentley: "The underlying implementations can be replaced with Kroxylicious-native ones incrementally" they _can_ be, but we should be clear that that's done something that's necessary or part of this proposal.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+The `kroxylicious-krpc-plugin` currently depends on `kafka-clients` solely for `ApiKeys`. After this change it depends on the new artifact instead (or we could codegen an `ApiKeys` from the RPC definitions themselves). Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.
+
+### What gets copied and generated
+
+| Component                                  | Source                                    | Approach                                               |
+|--------------------------------------------|-------------------------------------------|--------------------------------------------------------|
+| `MessageGenerator` + friends (~35 files)   | `kafka/generator/src/`                    | Copied into Kroxylicious repo as a new module          |
```
tombentley: "as a new module": let's decide what it's called.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+2. **Wire-level fidelity tests:** Any change to serialisation behaviour surfaces as a test failure. These tests serialise with our classes and deserialise with the original `kafka-clients` classes (and vice versa), so any divergence in the binary format is caught immediately.
+
+Together, `japicmp` catches API-level divergence and the fidelity tests catch wire-format divergence. Non-wire behavioural changes - validation logic, error handling - require reviewing the `kafka-clients` changelog when bumping the version, which is part of the standard upgrade process.
+
+### Enhancements to the generator
+
+#### Type-safe request-response pairing
```
tombentley, line 105: Personally I think this is just about worth doing. On the one hand it's a bit cute and doesn't offer a lot of extra type safety. But it also doesn't seem particularly hard or risky.

tombentley (commented):
A few nits, but over all I'm happy with this.

tombentley (approved):

tombentley:
I'm currently working on https://github.com/kroxylicious/kroxylicious/issues/4391. The code which I have currently uses some of Apache Kafka's internal classes in `org.apache.kafka.common.security.***.internals`. I'm trying to figure out whether that means my work has a dependency on this proposal as it stands.

* One option would be for the authentication related classes I care about to be included in the scope of 116. I suppose that means they would end up in our API module, even though they're not really implementation details (rather than API) and are specifically about SASL.
* Another option would be for me to just fork the classes I care about into my own module. But we already have the OAUTHBEAR validation filter, which also has a dependency on some of the same classes. It would be nice to work towards solving the kafka clients dependency in that filter as well as the one I'm intended to propose and implement. 
* A 3rd option: we create a new module, which will be a shared dependency for both OAUTHBEAR validation filter, and my work on 4391, but without doing that in the scope of this proposal or adding those classes to the API module. 

Thoughts?

proposals/116-kafka-api-migration.md
piotrpdev: Done in [`7df5748` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/7df5748122205c32bece1d9b0ce6aaa2cc6622ae)

proposals/116-kafka-api-migration.md
piotrpdev: Decided on `kroxylicious-kafka-message-generator` for the generator and `kroxylicious-kafka-common` for the rest (we copy Kafka's package structure), open to suggestions though :smiley:.

[`40a6db8` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/40a6db8176d6ac1f16edaca3559817c93467636c)

piotrpdev:
> The code which I have currently uses some of Apache Kafka's internal classes in `org.apache.kafka.common.security.***.internals`
> -- @tombentley 

Hmm, is that not a similar case to `OAuthBearerValidatorCallbackHandler` in https://github.com/kroxylicious/kroxylicious/issues/3996#issuecomment-4508381975, where we decided to [keep the dependency on `kafka-clients`](https://github.com/kroxylicious/kroxylicious/issues/3996#issuecomment-4631260304)?

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Approach: own the source from the start
+
+Kafka's `MessageGenerator` (~35 files in `generator/src/main/java/org/apache/kafka/message/`) is copied into the Kroxylicious repo as a new module (`kroxylicious-kafka-message-generator`). It is invoked during the build against the upstream JSON IDL spec files (bundled inside `kafka-clients`) to produce `*Data` source files under the Kroxylicious namespace (`io.kroxylicious.kafka.*`). The non-generated classes - protocol infrastructure, record classes, and scattered `common.*` types - are similarly copied into the Kroxylicious repo as source under the same namespace, retaining their original Apache Software Foundation copyright headers. All of these classes live in a new `kroxylicious-kafka-common` module, mirroring the structure of Kafka's own `org.apache.kafka.common` package.
```
k-wall: nit - naming: you say namespace in this sentence when you mean java package.
also this paragraph would be clearer if:
a) make the forking of MessageGenerator into a module kroxylicious-kafka-message-generator its own paragraph.
b) mention the role of the new module `kroxylicious-kafka-common` and make it clear that it will invoke the MessageGenerator

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+This gives full source ownership from day one. Every upstream change is a deliberate choice - review it, absorb it, or skip it. Generator enhancements (type-safe request-response pairing, Javadoc from IDL, `aliases`) are first-class commits to the in-repo generator, not patches layered on top. And when the time comes to replace buffer/record classes with Netty-native implementations, there is no bytecode transformation to unwind - we already own the source.
+
+The `kroxylicious-krpc-plugin` currently depends on `kafka-clients` solely for `ApiKeys`. After this change it depends on the new artifact instead (or we could codegen an `ApiKeys` from the RPC definitions themselves). Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.
```
k-wall: please disambiguate "new artefact".  just naming the module is simplest.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+### What gets copied and generated
+
+| Component                                   | Source                                    | Approach                                               |
+|---------------------------------------------|-------------------------------------------|--------------------------------------------------------|
+| `MessageGenerator` + friends (~35 files)    | `kafka/generator/src/`                    | Copied into `kroxylicious-kafka-message-generator`     |
+| `*Data` classes (~198)                      | IDL specs + copied `MessageGenerator`     | Generated into `kroxylicious-kafka-common`             |
```
k-wall: So `kroxylicious-api` is going to depend on `kroxylicious-kafka-common`.   Curious - did you consider running the generator in kroxylicious-api instead?  I'm wondering if we really need the kroxylicious-kafka-common module?

proposals/116-kafka-api-migration.md
k-wall: > After this change [kroxylicious-krpc-plugin] depends on the new artifact instead (or we could codegen an `ApiKeys` from the RPC definitions themselves). Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.

Either way it sounds a bit ugly.  

 I *think* eliminating the use of `ApiKeys` class from `KrpcGenerator` mechanism itself will simplify the problem.  From a quick look, I think its only use-case is mapping id to name.  The Kafka RPC definitions include the ids and names, so I think it should be an easy refactor.

```json
{
  "apiKey": 1,
  "type": "request",
  "listeners": ["broker", "controller"],
  "name": "FetchRequest"
}
```

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+The `kroxylicious-krpc-plugin` currently depends on `kafka-clients` solely for `ApiKeys`. After this change it depends on the new artifact instead (or we could codegen an `ApiKeys` from the RPC definitions themselves). Its role in generating filter interfaces, invokers, and decoders is otherwise unchanged.
+
+### What gets copied and generated
+
+| Component                                   | Source                                    | Approach                                               |
+|---------------------------------------------|-------------------------------------------|--------------------------------------------------------|
+| `MessageGenerator` + friends (~35 files)    | `kafka/generator/src/`                    | Copied into `kroxylicious-kafka-message-generator`     |
```
k-wall, line 49: The Kafka classes won't adhere to our coding conventions, so I'd bet money checkstyle and errorprone are going to light up.   I suggest we relax the rules for the copied file locations and have a separate task to make them adhere to our own conventions.

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+Exploratory work on the [`piotrpdev/kroxylicious#refactor/generate-data-class`](https://github.com/piotrpdev/kroxylicious/tree/refactor/generate-data-class) branch built a prototype of this infrastructure covering all 198 specs: parameterised fidelity tests at every valid version, randomised field-value tests (including explicit coverage of `null` for nullable fields), jqwik property-based tests for full-depth message instances, and a compile-time spec count guard. That prototype was designed around the FreeMarker template approach (see ["Rejected Alternatives"](#rejected-alternatives)), where generated classes were compiled at test time via `ToolProvider.getSystemJavaCompiler()` and loaded via `URLClassLoader`. With the source-ownership approach the classes are already on the classpath at build time, so only the class-loading mechanism changes - the verification logic itself carries over unchanged.
+
+These tests live in the Kroxylicious repo alongside the generated and copied classes, with `kafka-clients` as a test-scoped dependency. They run as part of the normal CI pipeline - when `kafka-clients` is bumped, they immediately report any incompatibility.
```
k-wall: What module do this tests live in?

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Drift in source-copied classes
+
+The non-generated classes (`MemoryRecords`, `RecordBatch`, protocol infrastructure, etc.) are copied source that may diverge from upstream over time as Kafka evolves. Drift is detected through two complementary mechanisms:
+
+1. **`japicmp` API comparison:** `japicmp-maven-plugin` compares by fully-qualified class name, so we cannot point it directly at our jar and `kafka-clients` - the package names do not match. Instead, we shade `kafka-clients` with the same `org.apache.kafka` → `io.kroxylicious.kafka` relocation into a throwaway reference artifact, then run `japicmp` comparing our source-built jar against that reference. Both jars now have matching package names, and `japicmp` reports exactly which APIs have diverged from upstream - method renames, signature changes, removed types. Kroxylicious already uses `japicmp` in the main build, so the plugin infrastructure is in place.
```
k-wall: I question whether we really need this.  The important thing is wire level compatibility.

proposals/116-kafka-api-migration.md
k-wall: Let's keep focus on what we have to do. This proposal is big enough already.  If we decide we want to propose it separately, after this proposal is delivered, but before the release, fine.

k-wall:
Could we add a high level implementation plan sketching out the path to being done?   We will want to chunk this work up, so let's think about it up-front.

SamBarker:
Building on the review comments from @tombentley (module naming) and @k-wall (module structure, whether a separate module is needed, where tests live), I think the proposal would benefit from spelling out the module structure more concretely. The current text says "copied into Kroxylicious repo as a new module" and "invoked during the build" but does not cover the relationships between modules, how the generator is invoked, or where the generated source and tests land.

One way this could look:

- **`kroxylicious-kafka-protocol`** — the runtime artifact that filter authors and Kroxylicious internals depend on. Generated `*Data` classes, copied protocol infrastructure (`ApiMessage`, `Readable`, `Writable`, schema model), record classes, `common.*` types. The unifying principle is purpose: everything in this module exists to represent or operate on the Kafka wire protocol. How the code got there — generated, copied from upstream, or eventually written from scratch — is an implementation detail, so the name should not encode provenance.
- **`kroxylicious-kafka-protocol-generator`** — a library module (not a plugin) containing the copied `MessageGenerator` source. Analogous to the generation logic inside `kroxylicious-krpc-plugin`, but producing Kafka's protocol message types rather than Kroxylicious filter interfaces/invokers/decoders.
- **`kroxylicious-krpc-plugin`** — gains a new goal that depends on the generator library and invokes it during the build to produce `*Data` source into `kroxylicious-kafka-protocol`. This keeps all code generation orchestrated through a single Maven plugin rather than introducing a second one.
- **Fidelity tests** in `kroxylicious-kafka-protocol` with `kafka-clients` as a test-scoped dependency.

I think making this structure explicit in the proposal would address several of the open questions — @k-wall's points about whether a separate module is needed and where tests live, and @tombentley's question about the module name.

@piotrpdev what do you think? If you see a better structure or have concerns about any of this, happy to discuss.

proposals/116-kafka-api-migration.md
SamBarker: Agree with @k-wall — this is a good idea but let's keep it out of scope for this proposal. The goal here is to take ownership of the protocol surface and unblock Kafka version upgrades. Type-safe pairing can be a follow-up once the foundation is in place.

proposals/116-kafka-api-migration.md
SamBarker: Good idea, but I'd keep it out of scope for this proposal unless it simplifies the package move itself. If eliminating the `ApiKeys` dependency from KrpcGenerator makes the migration easier — fewer things to relocate, fewer cross-module dependencies to untangle — then it's worth doing as part of this work. If it's just a nice cleanup, it can come after.

proposals/116-kafka-api-migration.md
SamBarker: I think wire compatibility is necessary but not sufficient on its own. These are Kafka's internal classes — Kafka doesn't consider them public API and won't reliably document changes to them. The 4.3.0 `record.*` move that motivated this proposal is a good example: even reading the JIRA (https://issues.apache.org/jira/browse/KAFKA-20128) you wouldn't pick it as a class relocation — you'd have to read the actual PR (https://github.com/apache/kafka/pull/21412) to connect the dots. No reasonable level of release note monitoring would have caught it proactively.

`japicmp` is the only automated signal for structural drift that isn't wire-visible — a renamed method, a changed signature, a removed type. Without it, the alternative is manually diffing our copies against upstream, which won't happen consistently.

That said, it's a transitional safety net. As each copied class gets replaced with a Kroxylicious-native implementation, we stop running `japicmp` against it — we've deliberately diverged at that point. But for classes that are still essentially Kafka's code, the drift signal remains valuable. The scope shrinks progressively, class by class.

SamBarker:
I think this is out of scope for this proposal — 116 is about owning the types that appear in the Kroxylicious API surface. Filters are a separate domain and are free to depend on `kafka-clients` directly for internal classes they need.

For your specific case, option 3 (shared module) actually makes sense to me. It's not just your SASL termination work — we have SASL initiation alongside the existing OAUTHBEARER usage as well, so there's a natural set of consumers for those security classes. That said, I'd keep it as a separate effort rather than adding it to 116's scope.

proposals/116-kafka-api-migration.md
tombentley: > Type-safe pairing can be a follow-up once the foundation is in place.

Not without breaking compatibility again. We if we want to do this separately we'd need to land a PR for it in the same release.

proposals/116-kafka-api-migration.md
piotrpdev: I decided against it because `kroxylicious-api` looks to entirely consist of interfaces and very simple/short classes, so I thought putting implementation-heavy code in there would be a bad idea. Let me know if this is an incorrect assumption though.

proposals/116-kafka-api-migration.md
piotrpdev: Done in [`207c48b` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/207c48b1ba7a3e8b146949154c6a47abe05d7ae8)

proposals/116-kafka-api-migration.md
piotrpdev: Done in [`207c48b` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/207c48b1ba7a3e8b146949154c6a47abe05d7ae8)

proposals/116-kafka-api-migration.md
piotrpdev: Mentioned in [`207c48b` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/207c48b1ba7a3e8b146949154c6a47abe05d7ae8)

proposals/116-kafka-api-migration.md
piotrpdev: Mentioned in [`207c48b` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/207c48b1ba7a3e8b146949154c6a47abe05d7ae8)

piotrpdev:
@SamBarker Thanks, I mentioned/named the modules more explictly in [`40a6db8` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/40a6db8176d6ac1f16edaca3559817c93467636c) and [`207c48b` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/207c48b1ba7a3e8b146949154c6a47abe05d7ae8) :+1:.

Also, I mentioned using `exec-maven-plugin` during the `generate-sources` phase, to mirror Kafka's approach instead of adding a goal to `kroxylicious-krpc-plugin`. I don't know if mixing the FreeMaker and generator approaches by running them under the same module would be a good idea.
[`02102a0` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/02102a0ea0d8f5459a017472d538e3a9a1ced75a)

piotrpdev:
> Could we add a high level implementation plan sketching out the path to being done? We will want to chunk this work up, so let's think about it up-front.

Done in [`b88eda3` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/b88eda36f30028634025be2e7adcd2b26cebd493) :+1:

proposals/116-kafka-api-migration.md
k-wall: Resolving - we are no longer considering patches.

proposals/116-kafka-api-migration.md
k-wall: In the case of bug fixes, probably yes. 
In the case of API changes, as it would be our API surface, probably not.

proposals/116-kafka-api-migration.md
k-wall: In this context, I took forking to mean "copy the code" into our repo, rather than depending on it from theirs.

proposals/116-kafka-api-migration.md
k-wall: >> If we decide we want to propose it separately, after this proposal is delivered, but before the release, fine.

> Not without breaking compatibility again. We if we want to do this separately we'd need to land a PR for it in the same release.

Let's aim to do this, for the same release, but let's deliver it as a follow on PR.  My motivation here is smaller steps = easier review.

proposals/116-kafka-api-migration.md
k-wall: We already have implementation heavy code in this module (the generated *Filter classes) generated by the krpc-plugin.  What's the difference?   I think keeping the number of public facing modules small is advantageous -  modules are part of the public API too.

We've also just learnt today that more module leads to MC limits problems too https://github.com/kroxylicious/kroxylicious/issues/4467

proposals/116-kafka-api-migration.md
k-wall: > If eliminating the ApiKeys dependency from KrpcGenerator makes the migration easier — fewer things to relocate, fewer cross-module dependencies to untangle — then it's worth doing as part of this work. If it's just a nice cleanup, it can come after.

My intuition is that it is the former.    It should be a change we can deliver separately, which then simplifies the rest of the problem.  Let me see if I can brush up a simple PR to test if my intuition is correct.

proposals/116-kafka-api-migration.md
k-wall: > I think wire compatibility is necessary but not sufficient on its own. 

I don't see this.  Once we've made the fork the Data classes become our public API.  If Kafka refactor their Data classes to add/remove/change methods, we don't care.  The only thing we care about is that the serialised byte array they produce is mutually intelligible.

I think we risk doing a load of  japicmp work that we'll immediately consign to bin.

proposals/116-kafka-api-migration.md
k-wall: I've put up a draft PR exploring this idea: https://github.com/kroxylicious/kroxylicious/pull/4472

SamBarker:
One though that occurred to me while thinking about https://github.com/kroxylicious/kroxylicious/issues/1121 : how will the runtime handle API Version negotiation once we break the Kafka Clients dependency. Do we still get what ever property we use to determine the proxies supported key set and versions?

proposals/116-kafka-api-migration.md
robobario: closing, shading is now a rejected alternative.

proposals/116-kafka-api-migration.md
robobario: comment is now outdated by latest content

proposals/116-kafka-api-migration.md
robobario: thread is now outdated and solution has evolved, closing

proposals/116-kafka-api-migration.md
robobario: we've moved away from shading, closing thread

proposals/116-kafka-api-migration.md
robobario: I agree with starting with it in `kroxylicious-api`, this contains the API filters need to implement which will include all the forked classes down to MemoryRecords and so on.

One argument for keeping them separate is that other technology could take advantage of this library if they want a Kafka serialization library with public-API type compatibility guarantees. Very niche so I think we'd be better starting with them in `kroxylicious-api` and could extract later into independent module if anyone ever asked. (or they can just depend on api and live with a few irrelevant classes, it's not a big footprint)

robobario:
> I'm currently working on [kroxylicious/kroxylicious#4391](https://github.com/kroxylicious/kroxylicious/issues/4391). The code which I have currently uses some of Apache Kafka's internal classes in `org.apache.kafka.common.security.***.internals`. I'm trying to figure out whether that means my work has a dependency on this proposal as it stands.
> 
> ... snip options for forking more of kafka internals ...
> 
> Thoughts?

Does that percolate into any public API, or is it all contained within some handler classes? I'm wondering, do we really need to stop depending on `kafka-clients` for Filter internals, vs public API. Yeah it's a bit precarious to depend on the internals but maybe for implementation details it's worth the risk.

edit: looks like Sam already answered the same.

proposals/116-kafka-api-migration.md
SamBarker: I get the argument — once we've forked, Kafka's API changes are irrelevant to us. My concern is narrower: if Kafka fixes a bug in one of the copied classes, how do we find out and pull it in?

I've convinced myself there's no good automated answer to that. japicmp only tracks signature changes — invisible to a logic fix inside a method body. A scheduled diff against upstream trunk breaks down the moment we deliberately don't absorb a change, because every subsequent diff is permanently dirty. Wire fidelity tests catch serialisation bugs but not behavioural ones.

So I think you're right to drop japicmp for the copied source classes — it wasn't giving us what I thought it was. The gap (non-wire logic bugs in copied source) is real but unautomatable. We should acknowledge it explicitly in the proposal and accept it as a known limitation of the source-ownership approach. The japicmp usage for the generated *Data classes is a separate matter and unaffected by this.

proposals/116-kafka-api-migration.md
SamBarker: Good challenge. I want to own something first: I previously proposed `kroxylicious-kafka-protocol` as a separate module from `kroxylicious-api` and argued for it on cohesion grounds — "everything in this module exists to represent or operate on the Kafka wire protocol." I still think that's a coherent principle, but cohesion alone isn't enough justification for a module boundary. The question is whether anything has a reason to depend on the protocol types *without* also depending on `kroxylicious-api`, and I can't find a convincing case for that.

The MC limits argument cuts both ways for me — I agree they're a real problem, but I don't think "fewer modules" is the right fix on its own. Modules should have a single reason to change. The issue is that the protocol types and the filter API types *don't* have separate reasons to change — they move together. So the boundary isn't load-bearing.

I think you're right: fold them together.

SamBarker:
Looking at the code, `ApiVersionsServiceImpl` calls `apiKey.latestVersion(true)`, `apiKey.oldestVersion()`, `apiKey.messageType.lowestSupportedVersion()`, `apiKey.messageType.lowestDeprecatedVersion()`, `ApiKeys.hasId()` and `ApiKeys.forId()` to intersect the broker's `ApiVersionsResponse` with the proxy's own supported ranges. All of these are derivable from the IDL specs, so as long as the copied/generated `ApiKeys` carries those methods the runtime version negotiation continues to work unchanged. The `kafka-clients` runtime dependency for this purpose is fully replaced by the generated `ApiKeys` in the new module.

SamBarker (approved):
LGTM — modulo collapsing `kroxylicious-kafka-common` into `kroxylicious-api` as discussed in https://github.com/kroxylicious/design/pull/116#discussion_r3624177265

robobario:
> Do we still get what ever property we use to determine the proxies supported key set and versions?

I believe so, we need a copy of ApiKeys to keep our Filter APIs unchanged (method signatures like `shouldHandleRequest(ApiKeys apiKey, short apiVersion)`), I assume our ApiKeys enum will have similar/identical properties for determining the range of versions known to the Proxy.

proposals/116-kafka-api-migration.md
robobario: Another interesting case is record batch versions and compression types that are allowed per RPC version (see ProduceRequest#validateRecords), that logic won't even be in the classes we are taking ownership of. Is that validation part of the wire protocol?  By the wire protocol you can say that new compression types could pop up in any version. This is true of the current proxy code anyway, we decode at a lower level and don't use those request classes. I guess we will just need tests we can handle all compression types and keep a close eye out for new magic versions appearing in RecordBatch.

proposals/116-kafka-api-migration.md
robobario: I think this is worth mentioning in the proposal since it breaks a potential cycle. kroxylicious-api depends on the generator to build the Filter interfaces, so we want to avoid krpc generator depending on kroxylicious-api. (think we've got a consensus that folding all the new bits into kroxylicious-api is the preference now)

robobario (approved):
LGTM, as said with the preference for rolling into kroxylicious-api, but also not totally against a separate module if we found that improves something.

proposals/116-kafka-api-migration.md
k-wall: I've marked https://github.com/kroxylicious/kroxylicious/pull/4472 as ready for review.  If it is accepted, I suggest we merge that and rebase this proposal.

proposals/116-kafka-api-migration.md
k-wall: > Another interesting case is record batch versions and compression types that are allowed per RPC version (see ProduceRequest#validateRecords), that logic won't even be in the classes we are taking ownership of. Is that validation part of the wire protocol?

Oh - good point.  I think we actually have that problem today.  There's nothing to stop a ProduceResponse Filter Author generating a 'bad' batch.  https://github.com/kroxylicious/kroxylicious/issues/4489

k-wall:
@piotrpdev what do you think to the latest comments?   I think we are heading towards consensus.   If you happy can you reflect the comments in the proposal and we can finally get this approved and start cutting code.

proposals/116-kafka-api-migration.md
k-wall: Just to be clear, I'm happy for this to stay in the proposal , but lets say it will be delivered by a separate phase.

k-wall:
There's a conversation rumbling away on https://github.com/kroxylicious/kroxylicious/issues/1121 about whether this proposal's scope needs to include generating the JsonConverters into a test module (possible kroxylicious-filter-test-support).  Why: authorization, entity-isolation and multi-tenant tests rely on the existence of these classes - see io.kroxylicious.testing.filter.requestresponsetestdef.KafkaApiMessageConverter

robobario:
Thinking on that, maybe it's good for our tests to keep using apache kafka code for test clients. Keep this design as limited as we can to the filter apis.

edit: ah we probably have test support for the filter apis, I see, yeah that does sound worth considering

k-wall:
> Thinking on that, maybe it's good for our tests to keep using apache kafka code for test clients. Keep this design as limited as we can to the filter apis.
> 
> edit: ah we probably have test support for the filter apis, I see, yeah that does sound worth considering

It wouldn't work.  ProduceRequestDataJsonConverter references ProduceRequestData.  That really has to be "our" ProduceRequestData, as we feed those to our Filter implementations.

proposals/116-kafka-api-migration.md
piotrpdev: Emphasized in [`0c70c80` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/0c70c8090721bad0ad0bc296c01764638ca3638c)

proposals/116-kafka-api-migration.md
piotrpdev: Folded `kroxylicious-kafka-common` into `kroxylicious-api`: [`8ae28fc` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/8ae28fc98f35881f6e60a6b3f84d1452953774c1).

(Kept `kroxylicious-kafka-message-generator` and `kroxylicious-kafka-common-tests`)

proposals/116-kafka-api-migration.md
piotrpdev: Removed some mentions of `ApiKeys` [`ed16ac9` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/ed16ac996d0bea794c762f8e0f22344842dd3b58) + [`d125e3d` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/d125e3d9037be605be8b38ac0e4d2a7774176623)

proposals/116-kafka-api-migration.md
piotrpdev: Removed `japicmp` in [`c5e7087` (this PR)](https://github.com/kroxylicious/design/pull/116/commits/c5e70877f30d78a40555a17977c3120e14869a05)

proposals/116-kafka-api-migration.md
```diff
@@ -0,0 +1,228 @@
...
+This makes `FilterContext.sendRequest()` type-safe - `context.sendRequest(header, produceRequest)` returns `CompletionStage<ProduceResponseData>` with no cast required. The pairing information already exists in `ApiKeys`; emitting the `implements` clause is a small generator change. It is not required for correctness and can be deferred if it complicates the initial delivery, but the decision to include or exclude it must be made upfront.
+
+#### Post-parity improvements
+
+These can be delivered after the initial release:
+
+- **Javadoc from IDL `about` strings:** The JSON specs already include documentation strings on most fields. The generator can emit these as Javadoc on getters/setters.
+- **`aliases` field in the IDL:** An optional field on spec field definitions (e.g. `"aliases": ["oldFieldName"]`) allowing Kroxylicious to maintain stable getter/setter names even if Kafka renames a field in a future IDL version. Requires schema model changes and generator work.
+
+### Migration for filter developers
```
k-wall, line 127: Non-blocking - @SamBarker would this be a good candidate for OpenRewrite?  I'm not suggesting we need this in the proposal.

k-wall (approved):
Fantastic work @piotrpdev

proposals/116-kafka-api-migration.md
gunnarmorling: Not Sam, but yes, that's a perfect use case for OpenRewrite. That said, any LLM-based coding agent will get this job done equally well, so I wouldn't spend much time on creating a recipe.

PR #118, opened 2026-06-30
Routing SPI: Default forwarding to the assigned upstream broker

hrishabhg:
Revises the Router plugin API (proposal 070) so a request on a broker-bound connection forwards to the broker that endpoint represents by default, and a router declares only the API keys it must intercept. Replaces staticRoutes().

Addresses the staticRoutes() gaps in kroxylicious#4177: the lost forward-to-assigned-broker default, the VC-level vs per-connection scope mismatch, and the unpopulatable key->route map.

SPI: intercepts(apiKey, ctx) gate (default = bootstrap connections only), shouldDecodeRequest(apiKey, ctx) (default false), and onRequest(apiKey, RequestFrame, ctx). Response decode is lazy via ResponseFrame.body(); no shouldDecodeResponse and no onResponse callback. Route binding is established at connection-creation time, which keeps bound-connection forwarding unambiguous and permits multiple routes to one cluster.

Relates-to: kroxylicious/kroxylicious#4177
Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/proposals/118-default-forward-to-assigned-broker.md
+++ b/proposals/118-default-forward-to-assigned-broker.md
@@ -99,13 +99,11 @@ interface Router {
 router, and responses fetched via `sendRequest` return decoded (`CompletionStage<ApiMessage>`), as
 070 specifies. This proposal changes only the gate.
 
-**Deferred: decode depth on the interception path.** A router that intercepts but only needs the
-header (client-id, subject) shouldn't force a full body decode, and a relay shouldn't pay to decode
-a response it never reads. Those optimisations (a `shouldDecodeRequest` declaration, lazy
-request/response frames) are real but separable: they optimise what happens *after* interception and
-have no bearing on the default destination. They are left to a follow-on proposal so this one stays
-focused on the gate. The gate already delivers the big win by itself — non-intercepted keys are
-never decoded at all (see the `DecodePredicate` wiring below).
+**Deferred: request decode depth on the interception path.** A router that intercepts but only needs
+the header (client-id, subject) shouldn't force a full body decode. That optimisation is real but
+separable: it optimises what happens *after* interception and has no bearing on the default
+destination, so it is left to a follow-on proposal. The gate already delivers the big win by
+itself — non-intercepted keys are never decoded at all (see the `DecodePredicate` wiring below).
 
 ### Three request shapes (one `onRequest`)
 
@@ -286,14 +284,13 @@ is to be revisited once the SPI settles and is not fixed by this proposal.
 
 ## Open questions
 
-- **Gate naming: `intercepts` vs `requiresDynamicRouting`.** One term should be used everywhere —
-  SPI, metrics, logs, docs. The gate covers rerouting and fan-out (both routing), which favours
-  `requiresDynamicRouting` and matches the existing `routingMode` vocabulary; alternatively, keep
-  `intercepts` and align the metric vocabulary to it when the metrics are revisited.
+- **Gate naming: `intercepts` vs `requiresDynamicRouting`.** Whichever term wins should be used
+  consistently across SPI, metrics, logs, and docs.
 
 ## Follow-on work (out of scope here)
 
-- **Decode depth on the interception path** — a `shouldDecodeRequest`-style declaration so a router
-  that routes on the header (client-id, subject) doesn't force a body decode, and lazy
-  request/response frames so a relay doesn't pay to decode a response it never reads. Separable from
-  the gate; deserves its own proposal.
+- **Request decode depth on the interception path** — a `shouldDecodeRequest`-style declaration so a
+  router that routes on the header (client-id, subject) doesn't force a body decode. Separable from
+  the gate; deserves its own proposal. (Response-side laziness was considered and dropped: virtual
+  node-id translation already decodes the node-reference-bearing responses, `Fetch` included, leaving
+  too narrow a case to justify widening `sendRequest`.)
--- a/proposals/118-default-forward-to-assigned-broker.md
+++ b/proposals/118-default-forward-to-assigned-broker.md
@@ -1,7 +1,7 @@
-# 118 - Default forwarding to the assigned upstream broker
+# 118 - Default forwarding to the assigned virtual node
 
 Revise the `Router` plugin API (proposal [070](./070-routing-api.md)) so that a request on a
-broker-bound connection is, by default, forwarded to the broker that endpoint already represents,
+node-bound connection is, by default, forwarded to the virtual node that endpoint already represents,
 and a router declares only the API keys it needs to **intercept**. Replaces `staticRoutes()`.
 
 Addresses the gaps raised in
@@ -29,14 +29,14 @@ implementation `onRequest()` is not yet wired, so an API key absent from `static
 Three problems with `staticRoutes()`, drawn from
 [kroxylicious#4177](https://github.com/kroxylicious/kroxylicious/issues/4177):
 
-1. **The useful default was lost.** Before routing, every broker was a virtual endpoint and traffic
-   to that endpoint _always_ flowed to the upstream broker it represents (`b0.proxy -> b0.cluster0`,
-   `b10000.proxy -> b0.cluster1`). `staticRoutes()` makes a router opt _into_ forwarding by listing
-   API keys, instead of inheriting that default.
+1. **The useful default was lost.** Before routing, every upstream broker was represented by its own
+   virtual endpoint and traffic to that endpoint _always_ flowed to the broker it represents
+   (`b0.proxy -> b0.cluster0`, `b10000.proxy -> b0.cluster1`). `staticRoutes()` makes a router opt
+   _into_ forwarding by listing API keys, instead of inheriting that default.
 
 2. **Scope mismatch.** `staticRoutes()` is a fixed `ApiKeys -> route name` map, but `Router`
-   instances are per-connection. A route name cannot express "forward to the broker _this connection_
-   represents" — that target (`virtualNode()`) is only known per connection.
+   instances are per-connection. A route name cannot express "forward to the implied target virtual
+   node" — that target (`virtualNode()`) is derived initially from the connection.
 
 3. **No router can populate the map usefully.** For a pass-through router it means enumerating ~80
    keys; for client-id, subject, or topic routers it is empty or near-empty, because nothing can be
@@ -44,34 +44,43 @@ Three problems with `staticRoutes()`, drawn from
 
 ## Proposal
 
-Invert the model: the runtime forwards a bound connection's traffic to its assigned broker by
+Invert the model: the runtime forwards a bound connection's traffic to its assigned virtual node by
 default; a router declares only what it must **intercept**.
 
 ### Route binding is established when the connection is created
 
-A connection reaches the proxy on a specific endpoint — a bootstrap endpoint, or a broker endpoint
-for one `(route, broker)`. The runtime binds the connection to that route and builds its upstream
+A connection reaches the proxy on a specific endpoint — a bootstrap endpoint, or a node endpoint for
+one `(route, virtual node)`. The runtime binds the connection to that route and builds its upstream
 filter chain at connection-creation time, exactly as the non-router proxy already builds a
-per-connection pipeline. Two consequences:
+per-connection pipeline. Three consequences:
 
-* A bound (broker-endpoint) connection's forward target is unambiguous: it already knows its
-  `(route, broker)` and filter chain, so `intercepts == false` simply forwards there — no per-request
-  route recovery.
+* A bound (node-endpoint) connection's forward target is unambiguous: it already knows its
+  `(route, virtual node)` and filter chain, so `shouldIntercept == false` simply forwards there — no
+  per-request route recovery.
+* "Derived initially from the connection" is deliberate: routers compose as a DAG, and each router in
+  the DAG sees its own virtual node id mapping. Router A may see a request destined for
+  virtualNodeId 2 while Router B sees the same request destined for virtualNodeId 0. A router reads
+  `ctx.virtualNode()` to learn which virtual node the request is associated with *at its own position
+  in the DAG*; it is not a single global identity. Nor is a virtual node a physical broker: the 1:1
+  mapping the runtime happens to use today is not something the SPI promises, and future topologies
+  need not preserve it.
 * This does **not** require one route per cluster. Two routes fronting the same cluster are two
   distinct endpoints; a connection's route is fixed by the endpoint it arrived on, not by decoding
   the request. The only thing out of scope is a single endpoint representing *several* routes at once
   (070's shared node-id mapping, opt-in via `RouterFactoryContext.allowSharedClusterTargets()`),
   which would require per-request decomposition.
 
-### Control plane vs data plane
+### Bootstrap and node endpoints are a structural distinction
 
-Kafka makes no protocol distinction between a bootstrap broker and a data broker. The router world
-must, because the connection's endpoint is the only signal for whether a default broker exists:
+Kafka makes no protocol distinction between a bootstrap broker and a data broker, but the proxy must,
+because the endpoint a connection arrived on is the only signal for whether a default destination
+exists. That makes the distinction a structural property the SPI depends on, not an implementation
+convenience, and it should be visible in the API:
 
-* **bootstrap endpoint** → connection is _unbound_ (`RouterContext.virtualNode()` is empty) → the
-  **control plane**: discovery, `METADATA`, coordinator lookup, the routing decisions.
-* **broker endpoint** → connection is _bound_ to one `(route, broker)` → the **data plane**:
-  data-plane traffic to a known broker.
+* **bootstrap endpoint** → connection is _unbound_: `RouterContext.virtualNode()` is empty, there is
+  no default destination, so the router must participate.
+* **node endpoint** → connection is _bound_ to one `(route, virtual node)`: the default destination
+  is that virtual node, so the router need not be consulted at all.
 
 ### The SPI
 
@@ -80,20 +89,28 @@ interface Router {
 
     /**
      * Whether the router must be invoked for this request. When false the runtime forwards the
-     * frame to the connection's assigned broker (virtualNode()) without calling onRequest().
-     * Default: intercept only when there is no assigned broker (bootstrap connections).
+     * frame to the connection's assigned virtual node (virtualNode()) without calling onRequest().
+     * Default: intercept only when there is no assigned virtual node (bootstrap connections).
      */
-    default boolean intercepts(ApiKeys apiKey, RouterContext ctx) {
+    default boolean shouldIntercept(ApiKeys apiKey, short apiVersion, RouterContext ctx) {
         return ctx.virtualNode().isEmpty();
     }
 
-    /** Called only when intercepts() is true. The router chooses the destination(s). */
+    /** Called only when shouldIntercept() is true. The router chooses the destination(s). */
     CompletionStage<RouterResponse> onRequest(short apiVersion, ApiKeys apiKey,
             RequestHeaderData header, ApiMessage request, RouterContext ctx);
 
     default void close() {}
 }
 ```
+
+`apiVersion` is part of the gate so a router can intercept only the version ranges it cares about,
+and so the gate composes directly with `DecodePredicate.shouldDecodeRequest(apiKey, apiVersion)`.
+
+The gate is a `boolean` rather than a richer action type. A `sealed RouteAction` (forward-to-implicit-node,
+intercept, forward-to-named-node, multicast) was considered and rejected as YAGNI: the boolean is
+sufficient for every router shape below, and it keeps symmetry with the filter API's `shouldHandle*`
+predicates. Richer actions can be introduced when a use case demands them.
 
 `onRequest` keeps 070's signature unchanged: an intercepted request is decoded and handed to the
 router, and responses fetched via `sendRequest` return decoded (`CompletionStage<ApiMessage>`), as
@@ -112,9 +129,9 @@ no extra methods:
 
 | shape | gate | onRequest does | example |
 |---|---|---|---|
-| **pass-through** | `intercepts == false` | _(not called; forward to bound broker)_ | — |
-| **pinned** | `intercepts == true` | send the whole request to one fixed route | `GetTelemetrySubscriptions` |
-| **fan-out** | `intercepts == true` | decompose, send to each route, recompose | `Metadata` |
+| **pass-through** | `shouldIntercept == false` | _(not called; forward to bound virtual node)_ | — |
+| **pinned** | `shouldIntercept == true` | send the whole request to one fixed route | `GetTelemetrySubscriptions` |
+| **fan-out** | `shouldIntercept == true` | decompose, send to each route, recompose | `Metadata` |
 
 Pinned is the degenerate case of fan-out where the decomposition yields a single `(route, request)`
 entry; it needs no separate code path.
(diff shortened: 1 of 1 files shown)


SamBarker:
Thanks for writing this up — and for pushing on `staticRoutes()` in #4177. The problems you identified are real and the core of this proposal is the right answer.

I want to share a direction I've been thinking about as context. I haven't fully worked it out, so this isn't a counter-proposal — it's a lens I'd like us to evaluate the proposal through.

**The default destination for every request is the broker the connection already represents.** That's the foundation I think we should be building on. I've been calling this "identity routing" in my head — every request has a default destination (the assigned broker), and routing is about when and why you deviate from that. Whether the identity router ends up as a concrete class, a runtime behaviour, or something else, I don't know yet. But the principle feels right: the assigned broker is always the answer unless something says otherwise.

Under that lens, `intercepts()` is well-aligned. The default gate (`ctx.virtualNode().isEmpty()`) says exactly the right thing: on a bound connection there's a default destination, so don't bother the router. On bootstrap there isn't, so the router must participate. The three shapes table and the four example routers all work cleanly against this. This is the strongest part of the proposal.

Where I'd like to push is whether the decode depth mechanics need to be part of this proposal or whether they're a follow-on. Thinking through each one against the same lens:

- **`shouldDecodeRequest`** — a router that intercepts but only needs the header (client-id, subject) shouldn't force a full decode. That's a real need. But it's a need of the interception path, not of the default destination. Rob made the point in #4177 that deserialization is a runtime decision, not a router decision. Could this be separated?

- **`RequestFrame`/`ResponseFrame` and lazy response decoding** — these widen 070's `sendRequest` return type and introduce new abstractions. They're well thought through, but they're about optimising what happens after a router intercepts. The default destination doesn't need frames — the request just goes where it was going. Could these follow once `intercepts()` is in place?

I'm not saying these are wrong. I'm asking whether they're separable from `intercepts()`, because if they are, this proposal gets smaller and lands faster — and the decode depth work gets a proposal that can focus on it properly.

Is there a dependency I'm missing that ties the decode depth to the routing default?

I've run out of time for a deeper look right now but will come back to this properly.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,315 @@
...
+
+```java
+interface Router {
+
+    /**
+     * Whether the router must be invoked for this request. When false the runtime forwards the
+     * frame to the connection's assigned broker (virtualNode()) without calling onRequest().
+     * Default: intercept only when there is no assigned broker (bootstrap connections).
+     */
+    default boolean intercepts(ApiKeys apiKey, RouterContext ctx) {
```
robobario: we should include `short apiVersion` in the signature, allowing for Routers that only need to intercept specific version ranges

hrishabhg:
@SamBarker 
> Is there a dependency I'm missing that ties the decode depth to the routing default?

No dependency — I've slimmed the proposal to just the gate and moved decode depth to follow-on work.

SamBarker:
One more thought on the decode depth side, specifically lazy response decoding.

Given the work in kroxylicious#4257, `shouldDecodeResponse` in `DelegatingDecodePredicate` is already symmetric with `shouldDecodeRequest` — and the router already drives response decoding declaratively for node ID translation. Metadata, FindCoordinator, DescribeTopicPartitions, Fetch (`preferredReadReplica`, `currentLeader`), and the other responses that carry node references are decoded regardless of what the router asks for, because virtual node ID mapping requires it.

Fetch is the primary response by volume and size, and it's already decoded. So lazy `ResponseFrame.body()` would only save decode cost for a response that a router intercepts but doesn't need to read, AND that nothing else in the pipeline (node ID translation, filters) has already decoded. Is that case real enough to justify widening `sendRequest`'s return type and introducing an asymmetry between the request and response decode paths?

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,298 @@
...
+- **Router plugin authors:** affected. The no-op default is "forward to the assigned broker," a safe
+  and usually-correct starting point.
+- **Filter API and existing filter configurations:** not affected.
+- **Operator / CRDs:** not affected — SPI/runtime only.
+
+## Compatibility
+
+Proposal [070](./070-routing-api.md) is not yet released, so `staticRoutes()` has no users to break;
+this revises an in-flight API. The `routingMode` metric (`static`/`dynamic`) is preserved:
+`intercepts == false` is reported as `static`, `intercepts == true` as `dynamic`.
```
SamBarker: The `static`/`dynamic` metric categories will need revisiting once the SPI settles — I wouldn't block on getting the metric vocabulary right in this proposal.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,298 @@
...
+  decode depth; `PASS_THROUGH` has no destination on bootstrap (so it needs the bound gate anyway).
+  Decode depth is deferred to a follow-on proposal in any case.
+- **`staticRoute(apiKey, ctx)` / `staticRouteForApiKey(apiKey)` returning a route** — a route is
+  cluster-granular, so it can't name the specific represented broker that `sendRequest` (070) targets
+  by node; loses leader affinity.
+
+## Open questions
+
+- **`intercepts` vs `requiresDynamicRouting` naming.** `intercepts` better describes the gate's real
+  job (the router may intercept to reroute, to fan out, _or_ to rewrite a response without
```
SamBarker: I'm not sure I understand the use case for a router rewriting a response. A route has a filter chain — that's where rewriting should occur. If this use case belongs in the filter chain rather than the router, it doesn't work as justification for the `intercepts` name.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,298 @@
...
+  rerouting). `requiresDynamicRouting` aligns with the `routingMode` metric vocabulary. The metric
+  name and method name need not match.
```
SamBarker: I'd push back on "the metric name and method name need not match." We strive for ubiquitous language — the same term should mean the same thing in the SPI, the metrics, the logs, and the docs. If the metric says `dynamic` and the method says `intercepts`, every reader has to mentally translate between two vocabularies. That translation cost compounds across the codebase. We pick one and use it everywhere.

hrishabhg:
> One more thought on the decode depth side, specifically lazy response decoding.
> 
> Given the work in kroxylicious#4257, `shouldDecodeResponse` in `DelegatingDecodePredicate` is already symmetric with `shouldDecodeRequest` — and the router already drives response decoding declaratively for node ID translation. Metadata, FindCoordinator, DescribeTopicPartitions, Fetch (`preferredReadReplica`, `currentLeader`), and the other responses that carry node references are decoded regardless of what the router asks for, because virtual node ID mapping requires it.
> 
> Fetch is the primary response by volume and size, and it's already decoded. So lazy `ResponseFrame.body()` would only save decode cost for a response that a router intercepts but doesn't need to read, AND that nothing else in the pipeline (node ID translation, filters) has already decoded. Is that case real enough to justify widening `sendRequest`'s return type and introducing an asymmetry between the request and response decode paths?

@SamBarker Not much gain in the decode path due to nodeIdTranslation, agreed and removed from the proposal.

Two version notes:

- NodeEndpoints/currentLeader arrive only in Fetch v16+ (Kafka 3.7).
- PreferredReadReplica (v11+, Kafka 2.3) forces Fetch response decoding anyway — and Kroxylicious currently does not observe whether a deployment actually uses rack-aware follower fetching, so the decode can't be skipped where it's off.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Whether the router must be invoked for this request. When false the runtime forwards the
+     * frame to the connection's assigned broker (virtualNode()) without calling onRequest().
+     * Default: intercept only when there is no assigned broker (bootstrap connections).
+     */
+    default boolean intercepts(ApiKeys apiKey, RouterContext ctx) {
+        return ctx.virtualNode().isEmpty();
+    }
+
```
SamBarker, line 98: What was your thinking behind changing the name from the question the runtime asks of filters? From where I'm sitting, this gate is asking the same question as `shouldDecodeRequest` on the filter API — "should the runtime decode this request rather than passing it through as raw bytes?" The proposal already shows it being OR'd into the `DecodePredicate`, which reinforces that it's the same question asked of a different participant.

SamBarker (approved):
I think the proposal is good, but we need a second maintainer to have a review before we can merge.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+```java
+interface Router {
+
+    /**
+     * Whether the router must be invoked for this request. When false the runtime forwards the
+     * frame to the connection's assigned broker (virtualNode()) without calling onRequest().
+     * Default: intercept only when there is no assigned broker (bootstrap connections).
+     */
+    default boolean intercepts(ApiKeys apiKey, RouterContext ctx) {
```
robobario: One idea we had in 070 was to support a richer set of actions here than intercept or forward-to-implicit virtualnode, for example you had your own ideas about forwarding on without implying a decode was required. Tom also speculated about broadcast/multicast possibilities, forwarding the same bytes to different upstreams. I wonder if we should leave the API open for extension towards that. Something like:

``` java
sealed interface RouteAction {
  ...
}

default RouteAction route(ApiKeys apiKey, short apiVersion, RouterContext ctx) {
    if (!ctx.virtualNode().isEmpty()) {
        return RouteAction.FORWARD_TO_IMPLICIT_NODE;
    } else {
        return RouteAction.INTERCEPT;
    }
}
```

In one way I like that this brings the implicit node target language into the API. Currently I don't think it's clear in the Router API javadoc where messages "forwarded to a route" are sent. One hesitation though is that future topologies might not have an implied node target, so adding the language might not extend well to those situations.

In future we could come up with further `RouteActions` like `forwardToNode(VirtualNode node)` implying we don't need to decode the message if we don't need to modify it or use it's contents for routing. Or `multiCastToNodes` and so on.

What do you think of this vs the boolean?

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+[kroxylicious#4177](https://github.com/kroxylicious/kroxylicious/issues/4177):
+
+1. **The useful default was lost.** Before routing, every broker was a virtual endpoint and traffic
+   to that endpoint _always_ flowed to the upstream broker it represents (`b0.proxy -> b0.cluster0`,
+   `b10000.proxy -> b0.cluster1`). `staticRoutes()` makes a router opt _into_ forwarding by listing
+   API keys, instead of inheriting that default.
+
+2. **Scope mismatch.** `staticRoutes()` is a fixed `ApiKeys -> route name` map, but `Router`
+   instances are per-connection. A route name cannot express "forward to the broker _this connection_
+   represents" — that target (`virtualNode()`) is only known per connection.
```
robobario:
```suggestion
   instances are per-connection. A route name cannot express "forward to the implied target virtual node" — that target (`virtualNode()`) is derived initially from the connection.
```

I think we have to be careful with the language since we can have a DAG of `Routers` which will see different virtual node id mappings, so Router A may know that virtualNodeId 2 goes to route x and then Router B may see that the request is destined for virtualNodeId 0. A Router knows contextually which virtualNode() is associated with the request.

And also we don't want to suggest that there will always be a 1:1 mapping between a virtual node id and some physical upstream broker, leaving room for future topologies.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+
+3. **No router can populate the map usefully.** For a pass-through router it means enumerating ~80
+   keys; for client-id, subject, or topic routers it is empty or near-empty, because nothing can be
+   pinned to a route purely by API key. The map either lists everything (bad DX) or returns `{}`.
+
+## Proposal
+
+Invert the model: the runtime forwards a bound connection's traffic to its assigned broker by
+default; a router declares only what it must **intercept**.
```
robobario: generally for this and the following section I wonder if we should talk about assigned virtual node vs broker.

From each Router's point of view it is aware of VirtualNodes, the connection implies a virtual node, then it's a runtime detail that we can happen to map it 1:1 to a broker. So I wonder if the proposal should use language throughout, rather than baking in this assumption. It's pedantic I know but it might help keep the concepts clear for future readers.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+* A bound (broker-endpoint) connection's forward target is unambiguous: it already knows its
+  `(route, broker)` and filter chain, so `intercepts == false` simply forwards there — no per-request
+  route recovery.
+* This does **not** require one route per cluster. Two routes fronting the same cluster are two
+  distinct endpoints; a connection's route is fixed by the endpoint it arrived on, not by decoding
+  the request. The only thing out of scope is a single endpoint representing *several* routes at once
+  (070's shared node-id mapping, opt-in via `RouterFactoryContext.allowSharedClusterTargets()`),
+  which would require per-request decomposition.
+
+### Control plane vs data plane
```
robobario: not quite sure what this section is here for, is it to justify solidifying inclusion of bootstrap and broker-bound endpoints as a Proxy neccessity? Something that should be visible in our APIs, rather than just a convenience. That seems like a good thing to have noted, I don't get the relationship to the plane types.

robobario (commented):
Thanks for the work @hrishabhg , I'm generally supportive of this. 

I wonder if we could make this more open to extension than a `boolean` decision and am wondering if the language in the proposal should be changed to talk less about "broker" and more about "virtual node" because future topologies may not have the "1 virtualNodeId: 1 upstreamNode" relationship.

proposals/118-default-forward-to-assigned-broker.md
tombentley: We have `shouldHandle*Request()` and `shouldHandle*Response()` in the filter API, so I think a better name would be `shouldIntercept` or `shouldHandle`.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+- **Gate naming: `intercepts` vs `requiresDynamicRouting`.** Whichever term wins should be used
+  consistently across SPI, metrics, logs, and docs.
+
+## Follow-on work (out of scope here)
+
+- **Request decode depth on the interception path** — a `shouldDecodeRequest`-style declaration so a
+  router that routes on the header (client-id, subject) doesn't force a body decode. Separable from
+  the gate; deserves its own proposal. (Response-side laziness was considered and dropped: virtual
+  node-id translation already decodes the node-reference-bearing responses, `Fetch` included, leaving
+  too narrow a case to justify widening `sendRequest`.)
```
tombentley: I find the idea appealing, but I'm not sure how feasible it would be given the `RequestData` and `ResponseData` APIs we have. I guess if the root object had a reference to the `ByteBuf` we could, in theory, deserialize progressively only when a called tried to access a struct which hadn't been decoded yet.

tombentley (commented):
@hrishabhg thanks for this!

Do you have a POC of this, or is it purely design currently? If you had a POC I'd like to take it for a spin and try it out.

proposals/118-default-forward-to-assigned-broker.md
k-wall: +1

proposals/118-default-forward-to-assigned-broker.md
k-wall: YAGNI - I think boolean is good enough for now. I like the symmetry with the filter API too.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+
+    default void close() {}
+}
+```
+
+`onRequest` keeps 070's signature unchanged: an intercepted request is decoded and handed to the
+router, and responses fetched via `sendRequest` return decoded (`CompletionStage<ApiMessage>`), as
+070 specifies. This proposal changes only the gate.
+
+**Deferred: request decode depth on the interception path.** A router that intercepts but only needs
```
k-wall, line 119: It's a good point.  If you were making routing decisions based on the client-id, then you could avoid the need to decode the body and route it in its undecoded form.  It is worth raising a separate issue just to register that the idea is come to light.  I'd wait for a use-case before we consider it further.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,292 @@
...
+    public CompletionStage<RouterResponse> onRequest(...) { throw new AssertionError("unused"); }
+}
+```
+
+**ClientId-based** — key is in the header; route chosen on the bootstrap connection.
+
+```java
+class ClientIdRouter implements Router {
+    // on a bound connection, the client-id is already known and the route is fixed; no need to intercept.
+    // dev can additional add `ApiVersionsRequestFilter` or decode ApiVersionRequest to validate clientId once in connection lifetime.
```
k-wall: I think the implementation is clear.  The comment "on a bound connection, the client-id is already known" didn't really help me understand. 

- It is the bootstrap phase where the client gets routed to the right cluster (`ctx.virtualNode().isEmpty() == true).  They'll get a metadata response containing the brokers of the cluster they are being route to. 
- The client then connects to the node(s) it needs. 
- Bound connections don't need to be routed at all.  `ctx.virtualNode().isEmpty() == false` so no routing takes place.

k-wall:
Thanks for the work @hrishabhg.   I'm +1 to this too.  It'd be great it you could address the open comments and we can get this over the line.

robobario:
I've got a prototype of this design in progress, should be able to share on Monday.

proposals/118-default-forward-to-assigned-broker.md
robobario: sounds good to me, looks like we have consensus about that now. Go ahead with boolean and `should*` style naming.

proposals/118-default-forward-to-assigned-broker.md
robobario: I'll resolve, I think the design is comprehensible even if the language isn't perfect

proposals/118-default-forward-to-assigned-broker.md
robobario: same as other comment, I think the language is good enough

robobario:
Proof-of-concept branch https://github.com/kroxylicious/kroxylicious/pull/4510

I think it's actually exposed some problems in `main` so it was a useful exercise (out-of-band requests from VC filters or Router Filters combined with Dynamic Routing has issues).

proposals/118-default-forward-to-assigned-broker.md
hrishabhg: rephrased the comment.

proposals/118-default-forward-to-assigned-broker.md
hrishabhg: It was a bad heading. Renamed to `Bootstrap and Node Endpoints`

proposals/118-default-forward-to-assigned-broker.md
hrishabhg: While agreeing on separate proposal, how do we check for a use-case e.g is it like customer asked for a client-id-based-routing?

SamBarker (approved):
LGTM

Thanks @hrishabhg and @robobario (for the prototype)

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,328 @@
...
+   node" — that target (`virtualNode()`) is derived initially from the connection.
+
+3. **No router can populate the map usefully.** For a pass-through router it means enumerating ~80
+   keys; for client-id, subject, or topic routers it is empty or near-empty, because nothing can be
+   pinned to a route purely by API key. The map either lists everything (bad DX) or returns `{}`.
+
+## Proposal
+
+Invert the model: the runtime forwards a bound connection's traffic to its assigned virtual node by
+default; a router declares only what it must **intercept**.
```
tombentley, line 48: I'm concerned that by switching to an open-by-default model for routers we'd be making it easier to write a router today which, when used tomorrow with a new version of Kafka (with added a new key), would break horribly because the router defaulted to forwarding when it should do something else.

Consider Kafka adding a new API key that carries node IDs in responses (like METADATA, FETCH, etc. do today). The runtime's NODE_ID_TRANSLATION_APIS set (RouterDispatchHandler.java:68-76) wouldn't include it yet. Under the current model, the request reaches onRequest() and the router has a chance to notice something is off. Under the proposed model, the request silently forwards, the response comes back with untranslated node IDs, and the client gets confused — with no indication anything went
  wrong.

proposals/118-default-forward-to-assigned-broker.md
```diff
@@ -0,0 +1,328 @@
...
+* A bound (node-endpoint) connection's forward target is unambiguous: it already knows its
+  `(route, virtual node)` and filter chain, so `shouldIntercept == false` simply forwards there — no
+  per-request route recovery.
+* "Derived initially from the connection" is deliberate: routers compose as a DAG, and each router in
+  the DAG sees its own virtual node id mapping. Router A may see a request destined for
+  virtualNodeId 2 while Router B sees the same request destined for virtualNodeId 0. A router reads
+  `ctx.virtualNode()` to learn which virtual node the request is associated with *at its own position
+  in the DAG*; it is not a single global identity. Nor is a virtual node a physical broker: the 1:1
+  mapping the runtime happens to use today is not something the SPI promises, and future topologies
+  need not preserve it.
```
tombentley, line 66: I think the change you're proposing contradicts the original intent of the SPI by assuming there's a 1:1 relationship between the virtual node that a client connects to and a target node. 

That is true today, but part of the reason for introducing VirtualNode was to allow us to change that in the future. For example, I've long wanted to keep the door open to a different network model where the proxy instances form a cluster, each presenting as a logical broker with a single endpoint accepting client connections. In that model there would be no such 1:1 relationship. As an example:

1. Each proxy instance presents as a logical broker with a single endpoint
2. Client does METADATA and gets back addresses of the proxy instances
3. Client connects to proxy instance 0 for data operations — this is a "node" connection
4. The runtime would set virtualNode() to be non-empty (it knows which logical broker the client connected to)
5. But "forward to that virtual node" is meaningless — the proxy instance IS the virtual node; there's no single upstream broker it maps to
6. PRODUCE requests need to go to the partition leader; FETCH requests to the relevant replica — routing is always content-dependent

The proposed default behaviour of `shouldIntercept` only makes sense when `VirtualNode` is a physical broker. 

Making the proposed change would make it easier to implement routers today, but those routers would be incompatible with such a networking model. (Yes you could still write a router that worked with the alternative model, but by default all the existing ones would not work). Moreover, there would be no way to determine that incompatibility, for example during proxy startup. You would have to assume that all routers were broken under the new model (so you couldn't hope to just reuse the existing implementations which existed at that point in time).

I think, in practice, if we adopted this proposal we're effectively saying were not interested in such an alternative networking model (and, by extension, all the potential features which would be feasible with it which are currently not feasible).

Maybe now is a good time to discuss that, but we should do so explicitly and having enumerated those potential features which we're saying we're not interested in.

hrishabhg: Thanks @tombentley — the alternative networking model is a fair challenge and I don't want to foreclose it. Two things I'd like to pin down.

1. `EndpointBinding` --> `VirtualNode` --> `Physical Broker(s)`. What exactly does VirtualNode represent? Is it the virtualization of an individual broker, the frontend EndpointBinding, or both? Today we have `ctx.send(ctx.virtualNode())`, where `ctx.virtualNode()` conveys a frontend abstraction, while `.send(node)` makes the same object appear to represent a backend node. That duality feels confusing. If `VirtualNode` is intended to model the frontend, send should probably accept something like an `upstreamNode` rather than a `VirtualNode`.

2. I also don't yet understand the performance implications of inspecting high-volume Produce/Fetch traffic versus treating it as opaque. My understanding was that this trade-off is the primary reason we have both opaque and decoded request frames (one of the key enablers for a high-performance proxy). There should still be a path that supports that optimization. If static routing is implemented using Router, it should have all the advantage that we have in current model in my opinion.

PR #119, opened 2026-07-03
Proposal for AuthN/Z API Refactor

tomncooper:
This PR proposes extracting identity concepts from `kroxylicious-api` into a new lightweight module, `kroxylicious-identity-api`, so that `kroxylicious-authorizer-api` can be consumed independently of the full proxy API and its transitive dependencies (kafka-clients, jackson-annotations, compression codecs). Apicurio Registry has already resorted to [copying the authorizer API source](https://github.com/Apicurio/apicurio-registry/pull/7829) to avoid this dependency tree.

The new module contains four types in `io.kroxylicious.identity`: a `Principal` interface, a `@SingularPrincipal` annotation, a deprecated-at-birth `Identity` bridge interface, and a `Subject` record. It has no compile-scope dependencies beyond `spotbugs-annotations` (provided scope).

The approach uses a phased migration to minimise blast radius. Only `Authorizer.authorize()` and `AuthorizeResult` break immediately (switching to the `Identity` bridge interface). All other API surfaces (`FilterContext`, `RouterContext`, subject builders) remain unchanged because the existing `Subject` record now implements `Identity`. At 1.0, the bridge types are removed and all APIs migrate to the new `Subject` record directly.

A PoC branch containing these proposed changes is available on my fork: [https://github.com/tomncooper/kroxylicious/tree/auth-api-refactor](https://github.com/tomncooper/kroxylicious/tree/auth-api-refactor-v2)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 .claude/settings.local.json
 .scratch/
+*.swp
--- a/proposals/119-auth-api-refactor.md
+++ b/proposals/119-auth-api-refactor.md
@@ -3,7 +3,8 @@
 The Kroxylicious authorizer API provides a general-purpose abstraction for access control decisions, deliberately designed to be agnostic of specific `Principal` and `ResourceType` implementations.
 However, the API currently depends on `kroxylicious-api`, which transitively pulls in Kafka client libraries, Jackson, and compression codecs. 
 This makes it less appealing for non-Kroxylicious projects to reuse.
-This proposal extracts the `Subject` and `Principal` concepts into a new lightweight module, `kroxylicious-identity-api`, so that `kroxylicious-authorizer-api` can be consumed independently of `kroxylicious-api` module.
+This proposal extracts identity concepts into a new lightweight module, `kroxylicious-identity-api`, containing a `Principal` interface, a `Subject` record, a deprecated-at-birth `Identity` interface (to aid in the migration to the new module) and a `@SingularPrincipal` annotation.
+The existing types in `kroxylicious-api` will be deprecated and gain super-types from the new module, enabling a phased migration where only the `Authorizer` API breaks immediately while `FilterContext` and other consuming APIs remain unchanged until version 1.0.
 
 ## Current situation
 
@@ -45,7 +46,6 @@ Any non-Kroxylicious project that wants to implement or consume the `Authorizer`
 
 As detailed in [Current situation](#current-situation) section, the authorizer API's usage of `Subject` is narrow: implementations only call `principals()`, `name()`, and `getClass()`.
 Yet this narrow usage forces consumers to accept `kafka-clients`, `jackson-annotations`, and compression codec libraries as transitive dependencies.
-The relative cost of importing the dependency does not match the value consumed.
 
 ### Authentication concepts are misplaced in the module hierarchy
 
@@ -60,41 +60,57 @@ The project's pre-1.0 status does not justify the change on its own. It should b
 However, the 0.x window does significantly reduce the cost of making a change that is justified on its own merits.
 The authorizer API was introduced relatively recently, so external adoption is likely to be minimal.
 Post-1.0, this same change would require deprecation cycles, compatibility shims, and migration documentation.
-Given the demonstrated external demand and the narrow usage pattern, the migration cost is justified now in a way that would be harder to justify later.
+Given the demonstrated external demand and the narrow usage pattern, the migration cost seems justified now in a way that would be harder to justify later.
 
 ## Proposal
 
-Introduce a new module, `kroxylicious-identity-api`, containing the `Subject` and `Principal` interfaces and the `@Unique` annotation.
-The `Subject` interface defines one abstract method (`principals()`) and provides default convenience methods (`uniquePrincipalOfType`, `allPrincipalsOfType`, `isAnonymous`) and a static factory (`anonymous()`), giving any implementation full subject-querying capability without additional dependencies.
+Introduce a new module, `kroxylicious-identity-api`, containing four types: a `Principal` interface, a `@SingularPrincipal` annotation, a deprecated-at-birth `Identity` bridge interface, and a `Subject` record.
 
-The existing concrete `Subject` record in `kroxylicious-api` is renamed to `ProxySubject` and implements the new `Subject` interface.
-API surfaces that only consume subjects (`FilterContext.authenticatedSubject()`, `RouterContext.authenticatedSubject()`, `FilterContext.clientSaslAuthenticationSuccess()`) change their type to the `Subject` interface, while subject-constructing APIs (`TransportSubjectBuilder`, `SaslSubjectBuilder`) retain `ProxySubject`.
-`kroxylicious-authorizer-api` switches its dependency from `kroxylicious-api` to the new module.
+The existing `Subject` record and `Principal` interface in `kroxylicious-api` will remain.
+However, the existing `Principal` gains `extends io.kroxylicious.identity.Principal`, the existing `Subject` gains `implements io.kroxylicious.identity.Identity`, and both will be deprecated along with the existing `@Unique` annotation.
+
+`kroxylicious-authorizer-api` switches its dependency from `kroxylicious-api` to the new module and its method parameters change to use the `Identity` bridge interface.
+This is the only immediately breaking change.
+
+API surfaces in `kroxylicious-api` that consume subjects (`FilterContext.authenticatedSubject()`, `RouterContext.authenticatedSubject()`, `FilterContext.clientSaslAuthenticationSuccess()`) are unchanged because the existing `Subject` record now implements `Identity` and flows into `Authorizer.authorize(Identity)` without conversion.
+Subject-constructing APIs (`TransportSubjectBuilder`, `SaslSubjectBuilder`) are also unchanged.
+
+At 1.0, the deprecated bridge types are removed and all APIs migrate to using the new `Subject` record directly.
 
 ### New module: `kroxylicious-identity-api`
 
-A new module in the `io.kroxylicious.identity` package containing three types:
+We will create a new module in the `io.kroxylicious.identity` package containing four types:
 
 `Principal` interface: a single method, `String name()`.
 The Javadoc contract (implementations must override `hashCode`/`equals` based on class and name) is carried forward from the existing `Principal`.
 
-`Subject` interface: one abstract method, `Set<Principal> principals()`, plus default convenience methods and a static factory:
+`@SingularPrincipal` annotation: `@Retention(RUNTIME)`, `@Target(TYPE)`.
+This is a renamed version of the existing `@Unique` annotation, with a name that more clearly describes its purpose: marking `Principal` implementations that should have at most one instance in a subject.
+This annotation is co-located in the identity module because the `uniquePrincipalOfType` default method on `Identity` will directly depend on it: the method will check `@SingularPrincipal` at runtime to validate that the requested principal type supports the "at most one" invariant.
 
-- `<P extends Principal> Optional<P> uniquePrincipalOfType(Class<P> uniquePrincipalType)`: default method that returns the unique principal of a given `@Unique`-annotated type, or empty. Throws `IllegalArgumentException` if the type is not annotated with `@Unique`.
+`Identity` interface: a deprecated-at-birth bridge interface with one abstract method, `Set<? extends Principal> principals()`, plus default convenience methods and a static factory:
+
+- `<P extends Principal> Optional<P> uniquePrincipalOfType(Class<P> uniquePrincipalType)`: default method that returns the unique principal of a given `@SingularPrincipal`-annotated type, or empty. Throws `IllegalArgumentException` if the type is not annotated with `@SingularPrincipal`.
 - `<P extends Principal> Set<P> allPrincipalsOfType(Class<P> principalType)`: default method that returns all principals matching a given type.
 - `boolean isAnonymous()`: default method that returns `true` when the principals set is empty.
-- `static Subject anonymous()`: static factory method that returns a `Subject` with no principals.
+- `static Identity anonymous()`: static factory method that returns an `Identity` with no principals, backed by a lightweight private implementation.
 
-The interface retains the `@FunctionalInterface` annotation: default and static methods do not count toward the single abstract method requirement, so trivial implementations remain possible (`() -> myPrincipalSet`).
+The `Identity` interface is annotated `@Deprecated` from its introduction.
+It exists solely as a bridge type so that both the existing `Subject` record (in `kroxylicious-api`) and the new `Subject` record (in this module) can be passed to `Authorizer.authorize()`.
+The wildcard return type on `principals()` (`Set<? extends Principal>`) is necessary so the existing `Subject` record's `Set<io.kroxylicious.proxy.authentication.Principal>` accessor satisfies the interface through covariant return types, given that the existing `Principal` gains `extends io.kroxylicious.identity.Principal`.
+The `Identity` interface will be removed at 1.0.
 
-Placing these methods on the interface is motivated by evidence from [Apicurio Registry's prototype][apicurio-pr], which copied the Kroxylicious `Subject` and re-implemented equivalent convenience methods (`principalOfType`, `isAnonymous`, `anonymous()`) for their `GrantsAuthorizer`.
-This demonstrates that these methods are useful for working with `Subject`, not proxy-specific behaviour.
-Providing them as defaults means external consumers get full subject-querying capability without writing boilerplate or duplicating logic.
+Placing the convenience methods on `Identity` is motivated by evidence from [Apicurio Registry's prototype][apicurio-pr], which copied the Kroxylicious `Subject` and re-implemented equivalent convenience methods (`principalOfType`, `isAnonymous`, `anonymous()`) for their `GrantsAuthorizer`.
+This demonstrates that these methods are useful for working with subjects, not proxy-specific behaviour.
+Providing them as defaults means all implementations of `Identity`, including both the existing and new `Subject` types, get full subject-querying capability without writing boilerplate or duplicating logic.
 
-`@Unique` annotation: `@Retention(RUNTIME)`, `@Target(TYPE)`.
-Marks `Principal` implementations that should have at most one instance in a `Subject`.
-This annotation is moved from the main `kroxylicious-api` because the `uniquePrincipalOfType` default method on `Subject` directly depends on it: the method checks `@Unique` at runtime to validate that the requested principal type supports the "at most one" invariant.
-Co-locating the annotation with the method that enforces it keeps the module self-contained. If `@Unique` remained in `kroxylicious-api`, the identity module would depend on the proxy module, defeating the purpose of the extraction.
+`Subject` record: a concrete record implementing the `Identity` interface.
+Its constructor validates `@SingularPrincipal` uniqueness: if a `Principal` implementation is annotated with `@SingularPrincipal`, the constructor rejects any principal set containing more than one instance of that type.
+This is the intended final type for all consumers of the identity API.
+External consumers such as Apicurio should target this type directly.
+
+The `Subject` record has its own `static Subject anonymous()` factory method that returns a `Subject` with no principals.
+This is separate from `Identity.anonymous()` because static methods on interfaces are not inherited in Java — when `Identity` is removed at 1.0, `Subject.anonymous()` must already exist for code that has migrated to the new type.
 
 The module has no compile-scope dependencies beyond `spotbugs-annotations` (provided scope, for package-level null-safety annotations).
 This means the transitive dependency tree for consumers of the authorizer API, which imports from this new module, becomes:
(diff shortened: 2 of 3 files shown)


carlesarnal:
Hey Tom, thanks for the mention. I can confirm the statements about Apicurio Registry are factually correct.

For our fine-grained per-resource authorization (https://github.com/Apicurio/apicurio-registry/issues/7724) draft work (PR #7829 (https://github.com/Apicurio/apicurio-registry/pull/7829)), we evaluated the Kroxylicious Authorizer API and found the interface design to be exactly what we needed, with the Authorizer, AuthorizeResult, Action, Decision, ResourceType all fit our use case well. However, the transitive dependency on kroxylicious-api (and through it kafka-clients, Jackson, compression codecs) was a non-starter for us, since our authz module is meant to be a lightweight, zero-dependency library that can be consumed by Registry and potentially other projects.

We ended up copying the authorizer API source into our own io.apicurio.authz module and stripping out the dependent code. It works, but it means we're now maintaining a fork of interfaces that could be shared. Our ADR explicitly notes: "If the Kroxylicious team extracts their API, migrating is mechanical."

This proposal does exactly that. With kroxylicious-authentication-api as a standalone module, we could drop our copied interfaces and depend on the real thing. +1 from the Apicurio Registry side.

.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea
 .claude/settings.local.json
+*.swp
```
tomncooper, line 4: For those lonely few who still use the one-true-editor.

SamBarker: Quit it! Oh wait you can't... 😜

SamBarker:
Thanks for writing this up, Tom — the core motivation is solid and the proposal is thorough.

I've left some initial inline comments but I'm out of time for a full review today. I'll come back with further feedback.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Authentication concepts are misplaced in the module hierarchy
+
+`Subject` and `Principal` are general authentication concepts. 
+They represent identity, not proxy behaviour.
+Their current placement in `kroxylicious-api` mixes identity types with proxy infrastructure.
+
+### The 0.x window makes this the right time
+
+The project is at version 0.x, where breaking changes can be expected and carry lower migration cost.
```
SamBarker: Semver does allow breaking changes at 0.x, but it's worth noting that the project has never actually shipped a breaking change to date. That's a track record that matters — it's built trust with adopters, and lowering that bar has a cost beyond the mechanical migration.

I'm not saying this isn't the right time — it probably is, and the proposal makes a good case. But I think we should be explicit that this would be the project's first breaking change rather than framing 0.x as blanket permission. What standard are we setting for future proposals that want to make breaking changes? Is "we're 0.x" sufficient justification on its own, or do we want a higher bar (like demonstrable external demand, which this proposal does have)?

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,267 @@
...
+The authorizer API was introduced relatively recently, so external adoption is likely to be relatively minimal.
+Post-1.0, this same change would require deprecation cycles, compatibility shims, and migration documentation.
+Doing it now avoids most of that overhead.
+
+## Proposal
+
+Introduce a new module, `kroxylicious-authentication-api`, containing minimal interfaces for `Subject` and `Principal`, plus the `@Unique` annotation (which is used for marking `Principals` which should have only one instance per subject).
```
SamBarker: The core argument for the new module is "put types where they belong so consumers don't pull in things they don't need." I agree with that, and the \`Principal\`/\`Subject\` interfaces clearly belong here — they're the contract external consumers need.

But \`@Unique\` seems like it's travelling along by proximity, not by domain. It's a general-purpose cardinality constraint ("at most one of this type in the set"), not an authentication concept. The annotation is only meaningful in combination with validation code, and that validation lives in \`ProxySubject\`'s constructor — which stays in \`kroxylicious-api\`.

External consumers who pull in the new module get the annotation but no enforcement. If they want enforcement, they write their own validation anyway. So the annotation doesn't actually help them by being here.

I think the cleanest outcome is: \`@Unique\` stays in \`kroxylicious-api\` (or \`io.kroxylicious.proxy.authentication\`), \`ProxySubject\` already depends on that module for validation, and \`kroxylicious-authentication-api\` stays genuinely minimal — just the two interfaces.

Is there a case I'm missing where an external consumer needs \`@Unique\` from the new module specifically?

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,267 @@
...
+```
+kroxylicious-authorizer-api
+├── kroxylicious-authentication-api   (compile)
+│   └── spotbugs-annotations          (provided)
+└── ...
+```
+
+### Use of a new package
+
+The new types live in `io.kroxylicious.authentication`, not `io.kroxylicious.proxy.authentication`.
```
SamBarker: The rationale for avoiding a split package is sound. But \`io.kroxylicious.authentication\` is a very broad package name for a module that currently contains two interfaces and an annotation.

If the Kroxylicious project later introduces other authentication-related types — say, credential providers, token validators, or authentication flow abstractions — the natural home for those would also be \`io.kroxylicious.authentication\`. If they land in a different module, we're back to a split package. If they're forced into this module to avoid the split, the "minimal, zero-dependency" property is lost.

Does the package need a narrower name that more precisely reflects what's in this module — the identity model rather than authentication as a whole? Something like \`io.kroxylicious.authentication.identity\` or \`io.kroxylicious.identity\` would leave room for other authentication concerns to live in sibling packages without splitting.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,267 @@
...
+The existing concrete `Subject` record in `kroxylicious-api` is renamed to `ProxySubject` and implements the new `Subject` interface, and `kroxylicious-authorizer-api` switches its dependency from `kroxylicious-api` to the new module.
+
+### New module: `kroxylicious-authentication-api`
+
+A new module in the `io.kroxylicious.authentication` package containing three types:
+
+`Principal` interface: a single method, `String name()`.
+The Javadoc contract (implementations must override `hashCode`/`equals` based on class and name) is carried forward from the existing `Principal`.
+
+`Subject` interface: a single method, `Set<? extends Principal> principals()`.
```
SamBarker: The original design in [proposal 009](https://github.com/kroxylicious/design/blob/main/proposals/009-authorizer.md) used `Set<Principal>` without a wildcard, and explicitly anticipated diverse `Principal` implementations:

> There's strong no need, at the API level, to be prescriptive about the implementations of `Principal` that are allowed. In particular, we don't want to limit `Principal` implementations to being provided by the API module.

The diversity is handled by `Principal` being an interface — you put whatever implementations you want *into* the `Set<Principal>`. A wildcard on the set would only matter if an implementor of `Subject` wanted to declare their return type as `Set<SomeConcreteSubtype>` rather than `Set<Principal>`. This proposal says `ProxySubject`'s component type changes to `io.kroxylicious.authentication.Principal` (the interface), so it returns `Set<Principal>` directly — no covariance is involved.

Is there a concrete case driving the wildcard, or can this stay as `Set<Principal>` as it was in the original design?

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,267 @@
...
+  `User`, `PrincipalFactory`, and test types (`FakeUniquePrincipal`, `FakeMultiplePrincipal`) add an explicit `import io.kroxylicious.authentication.Principal` since the same-package type no longer exists.
+
+- The `japicmp` configuration is updated with:
+  - `<exclude>` entries for: the removed `Principal` class, the removed `Unique` annotation, the removed `Subject` class (renamed to `ProxySubject`), the changed `PrincipalFactory#newPrincipal` return type, and the changed method signatures in `TransportSubjectBuilder#buildTransportSubject`, `SaslSubjectBuilder#buildSaslSubject`, `FilterContext#clientSaslAuthenticationSuccess`, `FilterContext#authenticatedSubject`, and `RouterContext#authenticatedSubject`.
+  - An `<ignoreMissingClassesByRegularExpressions>` entry for `io.kroxylicious.proxy.authentication.Principal`, because `japicmp` cannot resolve old bytecode signatures that reference the deleted class without this.
+
+- The concrete `ProxySubject` record retains all its existing behaviour, including the `User`-principal validation in its constructor, and its `uniquePrincipalOfType`, `allPrincipalsOfType`, and `isAnonymous` methods.
```
SamBarker: The proposal keeps `uniquePrincipalOfType`, `allPrincipalsOfType`, `isAnonymous`, and `anonymous()` on the concrete `ProxySubject`, treating them as proxy-specific. But looking at the Apicurio PoC ([apicurio-registry#7829](https://github.com/Apicurio/apicurio-registry/pull/7829)), they copied our `Subject` and kept these same methods — `principalOfType`, `isAnonymous`, `anonymous()` — because they need them in their own `GrantsAuthorizer` to extract a username from a `Subject`.

That suggests these methods are fundamental to working with `Subject`, not proxy-specific behaviour. If the primary external consumer of this API needs them, they belong on the `Subject` interface in the new module.

The rename to `ProxySubject` is still right — having a `Subject` interface and a `Subject` class would be painful regardless of package. But if the convenience methods move to the interface, `ProxySubject` doesn't need to appear in the signatures of `FilterContext.authenticatedSubject()`, `RouterContext.authenticatedSubject()`, etc. — those could return `Subject` (the interface). `ProxySubject` becomes an internal implementation detail rather than a type that leaks across the whole proxy API surface, and the blast radius of the rename shrinks from "every module that touches authentication" to "just the places that construct a `ProxySubject`".

proposals/119-auth-api-refactor.md
tomncooper: You make a fair point. Honestly, I included `@Unique` as I thought "external users might want to do the same". But there is not actual proven demand for that ATM and, as you say, it was only half the job anyway and would be easy to implement this externally if there was a need. 

I will revert `@Unique` to staying in `kroxylicious-api`.

proposals/119-auth-api-refactor.md
tomncooper: Right, I probably leaned too heavily on the "we are 0.x, so lets YOLO this!". 

While I do think the 0.x argument is valid, it should not be the primary one. The external demand should be the main reason. I will reframe the argument.

proposals/119-auth-api-refactor.md
tomncooper: I agree with you that we need a more focused name for the package. `io.kroxylicious.identity` seems the cleanest. The module could then be called `kroxylicious-identity-api` and remain standalone if other authentication related types are introduced in future that the authorizer API does not strictly need.

proposals/119-auth-api-refactor.md
tomncooper: So no, there isn't a concrete case for this. It will be better to reduce the diff with the original setup. I will revert to just using `Set<Principal>`.

proposals/119-auth-api-refactor.md
tomncooper: You make a good point. I wanted to keep the interface as simple as possible. However, given that Apicurio are actually using those methods and how moving them to the interface would actually reduce the changes to the rest of the codebase I think it would make sense to move them. This would mean that `@Unique` would have to move to the package aswell.

I will update the proposal.

SamBarker: Thinking about the @Unique tension between this thread and the earlier one: @Unique's javadoc is already framed entirely in terms of Principal and Subject — it's not actually generic, just generically named. If we rename it to something like `@UniqueIdentity` it belongs naturally in the identity module and the T2 concern goes away. @Unique in kroxylicious-api doesn't need to exist.

SamBarker:
The downstream changes section says `Subject.anonymous()` becomes `ProxySubject.anonymous()`, but the new `Subject` interface already has a `static Subject anonymous()` factory. Consuming code (filters, authorizers) can keep calling `Subject.anonymous()` with just an import change — consistent with the consume-vs-construct split applied to the method signatures.

More broadly: is `ProxySubject.anonymous()` needed at all? The value of `ProxySubject` over `Subject` is the `User`-principal validation in its constructor, but an anonymous subject has no principals — there's nothing to validate. If all anonymous-subject call sites can use `Subject.anonymous()`, that's one fewer thing on `ProxySubject` and a smaller downstream diff.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,284 @@
...
+
+#### Notable implications
+
+- **`kroxylicious-runtime` gains a direct dependency on `kroxylicious-identity-api`**: its source directly references the identity-api `Principal` type, so this must be an explicit compile-scope dependency.
+
+- **Maven dependency analyzer false positive**: some modules need `kroxylicious-identity-api` on the classpath to resolve `ProxySubject`'s super-interface or `Subject` return types, but their bytecode may not directly reference identity-api types.
+  This triggers Maven's analyzer.
+  Because several API surfaces (e.g. `FilterContext.authenticatedSubject()`) directly reference `io.kroxylicious.identity.Subject`, some modules will have genuine bytecode references that make the dependency required rather than a false positive. The exact set of modules needing an `ignoredNonTestScopedDependencies` override will be determined during implementation.
+
+- **Dependency enforcer allowlists**: `kroxylicious-identity-api` must be added to `bannedDependencies` allowlists in the `kroxylicious-filters`, `kroxylicious-kms-providers`, and `kroxylicious-kubernetes` parent POMs.
```
SamBarker: Nit: I found this paragraph hard to follow. It describes a false positive (bytecode doesn't reference identity-api types directly, but needs them for super-interface resolution), then says some modules *will* have genuine bytecode references.

But for modules that only use `ProxySubject` — do they actually need to declare `kroxylicious-identity-api` at all? It'll be a transitive compile-scope dependency of `kroxylicious-api` (because `ProxySubject implements Subject`). If they don't declare it explicitly, the analyzer has nothing to complain about. And if they do declare it unnecessarily, the analyzer is right to flag it — that's not a false positive.

I think this section needs reworking to clarify which modules actually need an explicit dependency and why.

SamBarker:
I've been thinking about the breaking change story here. If this only broke the authorizer API, I'd be comfortable with a clean break — the adoption is minimal and the case is well made. But the change leaks into `FilterContext`, which has real external adoption, and that's a higher bar.

The rejected alternative ("extract concrete types into the new module") correctly identifies the split-package problem with keeping the same package across two modules. But there's a third option the proposal doesn't consider: **deprecate and migrate**. This doesn't require a split package because the old types stay where they are:

1. Create `kroxylicious-identity-api` with new `Subject` interface, `Principal` interface, `@Unique` in `io.kroxylicious.identity` — no split package.
2. Old `Subject` record stays named `Subject` in `kroxylicious-api`, adds `implements io.kroxylicious.identity.Subject`.
3. Old `Principal` adds `extends io.kroxylicious.identity.Principal`.
4. Deprecate old `Principal` and old `@Unique`.
5. `FilterContext.authenticatedSubject()` keeps returning the concrete `Subject` — **no breaking change on the filter plugin API**.
6. `Authorizer` API switches to the new `Subject` interface — breaking, but minimal adoption as acknowledged.
7. Future release: rename `Subject` → `ProxySubject`, remove deprecated types.

The authorizer API gets its lightweight module and Apicurio gets what they need, without breaking the filter plugin contract. The cost is carrying deprecated bridge types for a release cycle and living with `Subject` (record) vs `Subject` (interface) name ambiguity across packages.

I think that trade-off is worth it to preserve FilterContext stability.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,284 @@
...
+
+The Kroxylicious authorizer API provides a general-purpose abstraction for access control decisions, deliberately designed to be agnostic of specific `Principal` and `ResourceType` implementations.
+However, the API currently depends on `kroxylicious-api`, which transitively pulls in Kafka client libraries, Jackson, and compression codecs. 
+This makes it less appealing for non-Kroxylicious projects to reuse.
+This proposal extracts the `Subject` and `Principal` concepts into a new lightweight module, `kroxylicious-identity-api`, so that `kroxylicious-authorizer-api` can be consumed independently of `kroxylicious-api` module.
+
+## Current situation
+
+The `Authorizer` interface and `AuthorizeResult` record in `kroxylicious-authorizer-api` reference `io.kroxylicious.proxy.authentication.Subject`, which is defined in `kroxylicious-api`.
+This means any project that wants to implement the `Authorizer` plugin interface must depend on `kroxylicious-api`, which transitively pulls in `kafka-clients`, `jackson-annotations`, and compression codec libraries (`zstd-jni`, `lz4-java`, `snappy-java`).
```
k-wall: aside: I expect that #116 will allow the  kroxylicious-api' s dependency on kafka-clients to go away as  it should eliminate the use of Kafka's ApiKeys, RequestHeaderData, ApiMessage etc.

That's not to say I don't appreciate that a dedicated -tightly defined- kroxylicious-identity-api may be more attractive for re-use.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,284 @@
...
+The authorizer API was introduced relatively recently, so external adoption is likely to be minimal.
+Post-1.0, this same change would require deprecation cycles, compatibility shims, and migration documentation.
+Given the demonstrated external demand and the narrow usage pattern, the migration cost is justified now in a way that would be harder to justify later.
+
+## Proposal
+
+Introduce a new module, `kroxylicious-identity-api`, containing the `Subject` and `Principal` interfaces and the `@Unique` annotation.
+The `Subject` interface defines one abstract method (`principals()`) and provides default convenience methods (`uniquePrincipalOfType`, `allPrincipalsOfType`, `isAnonymous`) and a static factory (`anonymous()`), giving any implementation full subject-querying capability without additional dependencies.
+
+The existing concrete `Subject` record in `kroxylicious-api` is renamed to `ProxySubject` and implements the new `Subject` interface.
```
k-wall: Could we avoid the need to rename io.kroxylicious.proxy.authentication.Subject (and thus the Kroxylicious API change), by just having the existing record implement io.kroxylicious.identity.Subject?  io.kroxylicious.proxy.authentication.Principal would need to extend io.kroxylicious.identity.Principal

kroxylicious-api would need to depend on kroxylicious-identity-api

Edit:  Ah - it is the annotations that make this awkward.  Annotations can't extend annotations so we couldn't have io.kroxylicious.proxy.authentication.Unique extend io.kroxylicious.identity.Unique

k-wall: I think I'm basically suggesting the same thing as https://github.com/kroxylicious/design/pull/119#issuecomment-4976223515

tombentley:
Thanks for the proposal @tomncooper.

@SamBarker you are right to be concerned about the compatibility story. However, the new `Subject` being an interface makes it hard to use:
* Everyone will need their own implementation. 
* We'd need the contract to define how `equals`, `hashCode` and `toString` work. This is extra annoying because those methods are implemented on `java.lang.Object` so declaring them as `abstract` methods on an interface does not force implementors into have to define those methods again, instead the compiler sees the ones inherited from Object and decides they're implemented.
* The contract is made more awkward by the interface having to make statements about what the constructor of the implementation should do wrt `@Unique`. 

I wondered if we could plan to use an interface (let's call it something else than `Subject`, e.g. `Identity`) only temporarily, with a new `Subject` record in the new module as the intended replacement right from the start? Claude came up with [a more detailed plan](https://gist.github.com/tombentley/e967d3390c4fe4177208e915b7ce8165) along these lines. Wdyt?

SamBarker:
@tombentley thanks for thinking through the compatibility story — your concerns about `Subject`-as-interface are fair, and I agree we should reverse the "rejected alternative" and include a concrete `Subject.of()` factory alongside the interface (the `List.of()` pattern). That addresses the equals/hashCode and "everyone needs their own impl" points.

On `@Unique` constructor enforcement — that's a proxy construction choice (`ProxySubject`), not essential to the `Subject` contract. Query time is the last responsible moment to enforce the rule, and an interface keeps that option open.

My main concern however is with the `Identity` bridge removal story. It solves release N, but at N+2 (or 1.0) `authenticatedIdentity()` returns a type that no longer exists, and none of the options for resolving that are clean. `Subject`-as-interface avoids this — same method name, same type name through the whole lifecycle, filter authors update one import at removal time. The bridge also breaks authorizer implementors twice (to `Identity`, then back to `Subject`) vs once.

[Full analysis here](https://gist.github.com/SamBarker/08be05f94ff774cd4d1b12dd7310d190) if you want the detailed walkthrough.

tombentley:
@SamBarker 

> On @Unique constructor enforcement — that's a proxy construction choice (ProxySubject), not essential to the Subject contract. Query time is the last responsible moment to enforce the rule, and an interface keeps that option open.

Not sure what exactly you mean by query time, but I don't think it's a "proxy construction choice" at all. I think it's an aspect of a particular `Principal` implementation class. If I could statically enforce that a `Subject` never has more than one `User` I would. Checking in the constructor at runtime is the next best thing, and correlates to authn time. Your alternative of expecting users of `Subject` to check for themselves seems far too weak -- I think callers will easily forget to add checks like that.

> `authenticatedIdentity()` returns a type that no longer exists

Where did `authenticatedIdentity()` come from? I don't think I mentioned such a thing. I assume you meant `authenticatedSubject()`. 

> none of the options for resolving that are clean.

That's intentional and specifically deferred to a major version, where breaking changes don't need to be clean. The benefit is that the end state is much cleaner than having a pointless, and difficult to understand interface lingering in the API (and being confusing by its existence) forever.

tombentley:
FWIW, since I had the plan anyway I thought I'd let Claude implement it and https://github.com/kroxylicious/kroxylicious/pull/4408 is the result.

SamBarker:
@tombentley

> Where did `authenticatedIdentity()` come from?

You're right — that was a Claude hallucination I didn't catch. Apologies for the confusion. The gist's analysis of the N+2 removal story still holds when you substitute the correct method name, but I should have reviewed that more carefully.

> I don't think it's a "proxy construction choice" at all. I think it's an aspect of a particular Principal implementation class.

Nobody's disputing that `Subject` validates `@Unique` — the question is *when*. `ProxySubject` validates at construction time, which is the right call for the proxy. `uniquePrincipalOfType()` is the natural place for query-time enforcement — it has the right name and signature, it just doesn't validate today because the current implementation assumes construction-time rejection. Construction-time enforcement is an optimisation (fail early), not the only valid approach, and the API contract in `kroxylicious-identity-api` shouldn't mandate it for all implementations.

Impersonation is a concrete example of why. A subject carrying two `User` principals — the admin and the impersonated user — with `uniquePrincipalOfType()` returning only the impersonated user is a valid design. A record with construction-time uniqueness enforcement rules that out. Impersonation is a well-established concept in identity management, which is exactly the domain we're extracting here. And with the ongoing SASL termination work moving us into direct OAUTHBEARER authentication, impersonation moves from structurally unlikely to plausible even within Kroxylicious itself. There needs to be a strong argument for why the API should foreclose on it.

> The benefit is that the end state is much cleaner than having a pointless, and difficult to understand interface lingering in the API (and being confusing by its existence) forever.

I don't accept that characterisation. I've laid out the case for `Subject`-as-interface: wider reuse without mandating construction-time constraints, single break for authorizer implementors, final type name from day one, doesn't foreclose on established identity patterns like impersonation. What concrete advantage does the record give over an interface + `Subject.of()` factory?

The interface isn't pointless — it *is* the final type. `Subject.of(principals)` gives consumers everything they need out of the box. `ProxySubject` adds construction-time validation. This is an API explicitly designed for reuse across projects with different requirements — an interface with a factory is the natural shape for that. A concrete record couples every consumer to one implementation's choices.

I don't question that the `Identity` bridge can work. My disagreement is that it's the right choice. The [gist](https://gist.github.com/SamBarker/08be05f94ff774cd4d1b12dd7310d190) walks through the migration in detail, but the core issue is: the bridge adds a temporary type and two breaks (for authorizer implementors) to arrive at a concrete record, when `Subject`-as-interface lands on the final type from day one with one break. The bridge is more migration work for a less flexible end state.

tombentley:
@SamBarker it feels that I'm arguing as which with Claude as with you, which isn't great from my PoV. 

> Nobody's disputing that Subject validates @Unique — the question is when.

Yes, and I'm saying that query time is too late, because it would be a programmer error to create a Subject which violates the uniquess. Nothing is to be gained from not failing fast.

> Impersonation is a concrete example of why. A subject carrying two User principals — the admin and the impersonated user — with uniquePrincipalOfType() returning only the impersonated user is a valid design.

That simply doesn't work with the current code where `Subject` is a record because it's impossible to distinguish which User instances is the impersonated user and which is the impersonator. To implement that with the current code you'd have two different types in principal: A `User` and an `Impersonator`.

And if you're deviating from the current code in pursuit of this then we should recognise that you're now in effect redesigning the API, which is not @tomncooper's objective. 

>  What concrete advantage does the record give over an interface + Subject.of() factory?

[Broken access control is the #1](https://owasp.org/Top10/2025/A01_2025-Broken_Access_Control/) on the OWASP top ten. I think we should be prioritizing the safety of any API we're providing here. As I've tried to explain, an interface is simply not constrained enough: It allows users of the API to try to write their own implementation, which is likely to break `Authorizers` if not done correctly. Using type safety such that the only kind of Subject is the Subject record, which is tested and can be the only implementation which `Authorizers` have been tested with is a significant benefit. OTOH, opening the door to person A writing their own Subject which accidentally doesn't work with the Authorizer written by person B is a huge disadvantage.

> This is an API explicitly designed for reuse across projects with different requirements — an interface with a factory is the natural shape for that.

This seems to be assuming that subclass inheritance is a kind of flexibility which is relevant to what the API is intended for. 

> A concrete record couples every consumer to one implementation's choices.

And that is intentional.

tomncooper:
Thanks for the discussion folks. I've read through both your positions and the detailed gists. From my understanding of what you've both said, I think the way forward should be the deprecation @SamBarker originally suggested, but using the approach suggested by @tombentley. I agree that the concrete `Subject` record is the safer option for the Authorizer API, although we will have to accept the risk of using the `Identity` interface for the duration of the deprecation (pre-1.0) period. 

Given that, I have listed below what I think we should do:

### A new module: `kroxylicious-identity-api` (package `io.kroxylicious.identity`)

Containing four types with zero external dependencies:

- `Principal` interface  with `String name()` method
- `@SingularPrincipal` annotation:  Which is a better name for `@Unique`, which marks principal types with at-most-one-per-subject semantics
- `Identity` deprecated-at-birth bridging interface with `Set<? extends Principal> principals()` and default convenience methods: `isAnonymous()`, `uniquePrincipalOfType()`, `allPrincipalsOfType()` and the `anonymous()` factory. This will be removed in a future 1.0 release.
- `Subject` record class implementing the `Identity` interface with a constructor that validates `@SingularPrincipal` uniqueness. This is the intended final type for all consumers of the API.

### Changes to `kroxylicious-authorizer-api`

- `Authorizer.authorize()` parameter changes from the old `Subject` to `Identity`
- `AuthorizeResult` subject component changes from the old `Subject` to `Identity`
- Module dependency switches from `kroxylicious-api` to `kroxylicious-identity-api`
- This will be binary-incompatible, but should be source-compatible for callers and implementors will only have to make a mechanical signature change

### Changes to `kroxylicious-api`

- Old `Principal` gains `extends io.kroxylicious.identity.Principal`
- Old `Subject` record gains `implements Identity`
- `@User` switches from using `@Unique` to `@SingularPrincipal`
- Old `Subject`, `Principal` and `@Unique` are all deprecated
- `FilterContext.authenticatedSubject()` is unchanged. The old `Subject` it returns is now an `Identity`, so when that is passed to the authorizer, there should be no breaking change for filter authors to deal with 

### Cleanup at 1.0

- Remove the deprecated-at-birth `Identity` interface (the bridge) from the `kroxylicious-identity-api` module
- Remove the `Subject` record, `Principal` interface and `@Unique` annotation from the `kroxylicious-api` module
- `Authorizer.authorize()` now changes from accepting the `Identity` interface to the new `Subject` record (from `kroxylicious-identity-api`) directly

Does the above sound like a reasonable way forward?

SamBarker:
@tombentley

> it feels that I'm arguing as much with Claude as with you

I use Claude as a drafting tool, the same way you used it for your counter-proposal and PoC. The arguments are mine. If hallucinations slip through, as with `authenticatedIdentity()`, I'm happy to acknowledge the mistake and correct it, as I did. That's not grounds for suggesting I've not engaged with this review.

On the substance: @tomncooper's synthesis works for me. I'm happy to move forward with `Subject`-as-record in `kroxylicious-identity-api` and the deprecated-at-birth `Identity` bridge.

k-wall:
@tomncooper the summary provided in https://github.com/kroxylicious/design/pull/119#issuecomment-5048167073 looks good. Looking forward to seeing the proposal updated.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Proposal
+
+Introduce a new module, `kroxylicious-identity-api`, containing four types: a `Principal` interface, a `@SingularPrincipal` annotation, a deprecated-at-birth `Identity` bridge interface, and a `Subject` record.
+
+The existing `Subject` record and `Principal` interface in `kroxylicious-api` will remain.
+However, the existing `Principal` gains `extends io.kroxylicious.identity.Principal`, the existing `Subject` gains `implements io.kroxylicious.identity.Identity`, and both will be deprecated along with the existing `@Unique` annotation.
+
+`kroxylicious-authorizer-api` switches its dependency from `kroxylicious-api` to the new module and its method parameters change to use the `Identity` bridge interface.
+This is the only immediately breaking change.
```
tombentley: So this must also impact `kroxylicious-authorizer-acl` too, but that's not called out.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,309 @@
...
+
+- This is a binary-incompatible change.
+  However, it is source-compatible for callers: since the existing `Subject` record now implements `Identity`, code that passes an existing `Subject` to `authorize()` compiles without changes.
+  `Authorizer` implementations must update the parameter type in their `authorize()` method from `io.kroxylicious.proxy.authentication.Subject` to `io.kroxylicious.identity.Identity`.
+  The fix is mechanical (change one import and the parameter type).
+
+### Changes to downstream modules
+
+The only downstream modules that need source changes are those containing `Authorizer` implementations, which must update their `authorize()` method signature from the existing `Subject` to `Identity`.
+Two such implementations exist in the codebase.
```
tombentley:
```suggestion
Two such implementations exist in the codebase, and no usages outside the project are known.
```

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,309 @@
...
+## Affected/not affected projects
+
+### Affected
+
+- `kroxylicious`: the main repository; all changes are within this repo.
+
+### Not affected
+
+- `kroxylicious-junit5-extension`
+- `kroxylicious-operator`
```
tombentley: This is confusing `kroxylicious-junit5-extension` (a repo) with `kroxylicious-operator` (a module)

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,309 @@
...
+- `AuthorizeResult`'s `subject` component type changes from `io.kroxylicious.identity.Identity` to `io.kroxylicious.identity.Subject`.
+- `FilterContext.authenticatedSubject()` return type changes from `io.kroxylicious.proxy.authentication.Subject` to `io.kroxylicious.identity.Subject`.
+- `RouterContext.authenticatedSubject()` return type changes from `io.kroxylicious.proxy.authentication.Subject` to `io.kroxylicious.identity.Subject`.
+- `FilterContext.clientSaslAuthenticationSuccess()` parameter type changes from `io.kroxylicious.proxy.authentication.Subject` to `io.kroxylicious.identity.Subject`.
+- `TransportSubjectBuilder.buildTransportSubject()` return type changes from `CompletionStage<io.kroxylicious.proxy.authentication.Subject>` to `CompletionStage<io.kroxylicious.identity.Subject>`.
+- `SaslSubjectBuilder.buildSaslSubject()` return type changes from `CompletionStage<io.kroxylicious.proxy.authentication.Subject>` to `CompletionStage<io.kroxylicious.identity.Subject>`.
+
+By this point, the deprecated types will have been available for at least one release cycle, giving consumers time to migrate.
+The deprecation warnings serve as documentation of the migration path.
```
tombentley: Presumably those are **compile time** deprecation warnings, so they're visible to the developer, but not the end user?

tombentley (commented):
I _think_ what this is describing is a reasonable trade-off, and reflects what we've discussed in the comments. However, Claude (I assume) has been pretty verbose in how it's described, repeating things across sections, describing code changes in lengthy paragraphs, rather than just showing the code (with elided method bodies , and annotates changes). So TBH I'm not _entirely_ certain. 

I think the proposed changes is just two sections: "Initial changes", and "Changes in the following major version". Those could then just show code with what changes in one subsection, and another subsection discussing the impact of those changes (_what_ breaks, _who_ needs to do _what_ by _when_ in order to unbreak it). 

E.g.

> # Initial changes
> 
> ## What changes
> 
> In `kroxylicious-identity-api`:
> ```java
> package io.kroxylicious.identity; // new package in new `kroxylicious-identity-api` model 
> 
> @Deprecated
> interface Identity {
>     // same methods as the current io.kroxylicious.proxy.authentication.Subject, except in terms of io.kroxylicious.identity. Principal 
> }
> 
> record Subject implements Identity {
> }
> 
> interface Principal {
>   // same methods as io.kroxylicious.proxy.authentication.Princpal
> }
> ```
>
> In kroxylicious-api: 
>
> ```java
> package io.kroxylicious.proxy.authentication;
> record Subject implements Identity { // add implements
> }
> ```
>
> In `kroxylicious-authorizer-api`:
> ...
> In `kroxylicious-authorizer-acl`:
> ...
> 
> ## Impact
> 
> ....

I think if you adopted that structure it would be a lot easier for reviewers to understand what they're being asked to accept.

k-wall:
@tomncooper I'll get to re-reviewing this tomorrow.

proposals/119-auth-api-refactor.md
tomncooper: Yes, good point, I've added that.

proposals/119-auth-api-refactor.md
tomncooper: Good point, I'll add more detail on that. 

I am switching to your suggested code based format so I'll add `@deprecated` annotations with `forRemoval=true` to make it more concrete.

tomncooper:
Thanks @tombentley, I agree that this is too verbose. It is a reasonably low-level proposal so I think your code format suggestion is a good one. I will push the reformatted version soon.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,298 @@
...
+
+In `kroxylicious-api`:
+
+```java
+package io.kroxylicious.proxy.authentication;
+
+// Subject record: removed
+// Principal interface: removed
+// @Unique annotation: removed
+// User switches from using @Unique to @SingularPrincipal
```
k-wall: Presumably User now implements `io.kroxylicious.identity.Principal` too

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,298 @@
...
+```
```
k-wall, line 239: Presumably PrincipalFactory type bound change from io.kroxylicious.proxy.authentication.Principal to identity.Principal

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,298 @@
...
+SaslSubjectBuilder.buildSaslSubject()           // returns CompletionStage<io.kroxylicious.identity.Subject>
+```
+
+#### Impact
+
+All types migrate to `io.kroxylicious.identity`.
+Filter, router and authorizer plugin authors update imports.
+By this point, the deprecated types will have been available for at least one release cycle, giving consumers time to migrate.
+
+## Rejected alternatives
```
k-wall, line 279: Could we add a sections talking about

- test strategy - for instance at stage 1, I'd expect we'll want to add tests ensuring that `Authorizer.authorize` works with both new and old subject/identities.
- approach to japicmp - we are going to need to relax the japicmp rules to let these changes happen.

proposals/119-auth-api-refactor.md
tomncooper: Good point. In phase 1 `User` gets this transitively since the existing Principal gains `extends io.kroxylicious.identity.Principal`. But at 1.0 when the old Principal is removed, `User` would need to switch to `implements io.kroxylicious.identity.Principal` directly. I'll add this to the Phase 2 section.

proposals/119-auth-api-refactor.md
tomncooper: Again, a good point! In 1.0 when the old `Principal` is removed, `PrincipalFactory`'s type bound would need to change from `<P extends io.kroxylicious.proxy.authentication.Principal>` to `<P extends
  io.kroxylicious.identity.Principal>`. I'll add this to the Phase 2 section.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,298 @@
...
+TransportSubjectBuilder.buildTransportSubject() // returns CompletionStage<io.kroxylicious.identity.Subject>
+SaslSubjectBuilder.buildSaslSubject()           // returns CompletionStage<io.kroxylicious.identity.Subject>
+```
+
+#### Impact
+
+All types migrate to `io.kroxylicious.identity`.
+Filter, router and authorizer plugin authors update imports.
+By this point, the deprecated types will have been available for at least one release cycle, giving consumers time to migrate.
+
```
k-wall, line 278: We suspect there will be filter authors out there.  Let's ship an OpenRewrite rule for this.

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,298 @@
...
+Compatible changes (no action required):
+
+- Adding `extends io.kroxylicious.identity.Principal` to the existing `Principal` interface
+- Adding `implements Identity` to the existing `Subject` record and widening type parameter bounds
+- Introducing `kroxylicious-identity-api` as a new module
+- `kroxylicious-identity-api` must be added to `bannedDependencies` allowlists in relevant parent POMs
+
+All other modules (including those that use `FilterContext.authenticatedSubject()` or `RouterContext.authenticatedSubject()`) require no source changes.
+These modules will see compile-time deprecation warnings for usages of the existing `Subject`, `Principal` and `@Unique`, visible to developers during builds but not to end users.
+
```
k-wall, line 213: So if I understand things right, a third party (line Apicurio) would be free to start using the new API from Phase 1.    If they write there own private Authorizer impls, they'd use new identity.Subject from the outset - so there would be no breaking change for them at Phase 2.

Do I get this right?

k-wall (approved):
Couple of nits, but otherwise LGTM

proposals/119-auth-api-refactor.md
tomncooper: Good point on mentioning testing. The `kroxylicious-authorizer-api` test suite seems like the place for those tests  this since we can verify that both the existing and new `Subject`s are accepted as `Identity` without needing a real `Authorizer` implementation. I'll add a testing section to the proposal.

On `japicmp`, I'll confess I am not that familiar with that tool. But from what I can tell from my testing on the [PoC branch](https://github.com/tomncooper/kroxylicious/tree/auth-api-refactor-v2), the Phase 1 changes to `kroxylicious-api` are all binary-compatible (adding super-interfaces, changing type parameter bounds where the erasure is unchanged and deprecations), so `japicmp` shouldn't flag them. The breaking changes are on `kroxylicious-authorizer-api` which doesn't appear to have japicmp enforcement?

proposals/119-auth-api-refactor.md
tomncooper: Good point. If #116 removes `kafka-clients` from `kroxylicious-api`, then the transitive dependency cost does become less of an issue. However, as you say, the dependency argument is only one of the motivations. External consumers like Apicurio would still need to depend on `kroxylicious-api` (with whatever dependencies remain or are added in future) just to get `Principal` and `Subject`. 

Again, as you say, a dedicated module with zero compile-scope dependencies is a cleaner contract for reuse, regardless of what happens with #116.

proposals/119-auth-api-refactor.md
tomncooper: I wasn't aware of OpenRewrite, yes, we should definitely plan to ship a rule for this in the 1.0 release. I will add a note to Phase 2. Have we used this before in Kroxy? Is there any prior-art I could look at?

proposals/119-auth-api-refactor.md
tomncooper: Yes, that's the idea. From Phase 1 they'd depend on `kroxylicious-identity-api` and use `io.kroxylicious.identity.Subject` directly. Their `authorize()` method would accept `Identity` during Phase 1, then change to `Subject` at Phase 2 when the bridge is removed, but that's just a mechanical signature change since they'd already be working with the new `Subject` type throughout.

proposals/119-auth-api-refactor.md
tomncooper: @k-wall should `japicmp` be enabled on the `authorizer-api`? If so, I assume we would deal with that outside this proposal?

proposals/119-auth-api-refactor.md
tomncooper: I added a line about including a rule for the 1.0 release.

SamBarker (approved):
Thanks for keeping at this one @tomncooper

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,308 @@
...
+record Subject(Set<Principal> principals) implements Identity {
+    // These methods originally had bounds <P extends old.Principal>.
+    // Identity's defaults have bounds <P extends new.Principal>.
+    // Both erase to the same JVM signature, but Java requires identical
+    // bounds for a valid override (JLS §8.4.8.1), not just a subtype
+    // relationship. Without changing the bounds, compilation fails with
+    // a name clash. The change is safe: since old.Principal extends
+    // new.Principal, any type that satisfied the old bound also satisfies
+    // the new one, so callers are unaffected.
+    @Override <P extends io.kroxylicious.identity.Principal> Optional<P> uniquePrincipalOfType(Class<P> type) { ... }
```
tombentley, line 163: Will this method's implementation will work with both the new and old annotation types, or does `User` get annotated with both between now and the next major version?

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,308 @@
...
+FilterContext.clientSaslAuthenticationSuccess() // accepts io.kroxylicious.identity.Subject
+TransportSubjectBuilder.buildTransportSubject() // returns CompletionStage<io.kroxylicious.identity.Subject>
+SaslSubjectBuilder.buildSaslSubject()           // returns CompletionStage<io.kroxylicious.identity.Subject>
+```
+
+#### Impact
+
+All types migrate to `io.kroxylicious.identity`.
+Filter, router and authorizer plugin authors update imports.
+By this point, the deprecated types will have been available for at least one release cycle, giving consumers time to migrate.
```
tombentley: Do you mean

```suggestion
By this point, the new types will have been available for at least one release cycle, giving consumers time to migrate.
```

proposals/119-auth-api-refactor.md
```diff
@@ -0,0 +1,308 @@
...
+
+```java
+FilterContext.authenticatedSubject()            // returns io.kroxylicious.identity.Subject
+RouterContext.authenticatedSubject()            // returns io.kroxylicious.identity.Subject
+FilterContext.clientSaslAuthenticationSuccess() // accepts io.kroxylicious.identity.Subject
+TransportSubjectBuilder.buildTransportSubject() // returns CompletionStage<io.kroxylicious.identity.Subject>
+SaslSubjectBuilder.buildSaslSubject()           // returns CompletionStage<io.kroxylicious.identity.Subject>
+```
+
+#### Impact
```
tombentley, line 264: Does this section capture the full extent of the changes needed @1.0? Consider someone with this code

```java
new.AuthorizeResult r = ...
Identity i = r.subject()
```

That's not going to be fixed just by changing the `import`s, which is what this section makes it sound like. The need to change `Identity` to `new.Subject`, too, right?

tombentley (commented):
Thanks @tomncooper, that's much clearer now! I had a couple more questions.

proposals/119-auth-api-refactor.md
tomncooper: So I admit I went back and forth a lot on this overloaded annotations issue when working on the proposal.

To answer your question: neither. The old `Subject`'s unique Principal method (which now overrides the `Identity` interface) would keep its current implementation and continue to check only the `@Unique` annotation. `User` wouldn't get annotated with both old and new.

The two annotation systems are (largely) separate during Phase 1. The old `Subject`'s constructor takes `Set<old.Principal>`, so principals defined purely against the new identity API can't be passed in. AFAICS all the internal APIs still return the old `Subject` type, so new `io.kroxylicious.identity.Subject` instances shouldn't be passed into Kroxylicious internals either.

Going in the other direction, putting `User` (which only has `@Unique`) into a new `Subject` and calling `uniquePrincipalOfType(User.class)` would also need to fail. The proposal should specify the behaviour of `Identity.uniquePrincipalOfType` when its called with a type that lacks `@SingularPrincipal`. The existing `Subject` already throws `IllegalArgumentException` when the type lacks `@Unique`, so the new `Identity` default should follow the same pattern. I'll add this to the proposal.

However, with this current approach there is a gap. A principal that implements old `Principal` but uses `@SingularPrincipal` instead of `@Unique` _could_ be passed into the old `Subject` implementation. In that case, the old `Subject`'s constructor wouldn't enforce its uniqueness, and `uniquePrincipalOfType` would reject it. It's a contrived scenario that requires mixing deprecated classes and new annotations from different modules/packages, but it _could_ happen.

So this raises the question of whether we should take a belt-and-braces approach and have both the old and new `Subject` check for both `@Unique` and `@SingularPrincipal` in their constructor validation and `uniquePrincipalOfType` methods, as well as the other internal APIs? 

What do you think, is it worth accounting for, or is the scenario too contrived to worry about?

proposals/119-auth-api-refactor.md
tomncooper: Good point. Outside Authorizer implementations it should just be an import change but within they would need to change the type from `Identity` to the new `Subject` type. I will add more detail to the impact section.

proposals/119-auth-api-refactor.md
tombentley: I think it _is_ worth doing. We might might think it's a contrived scenario, but it's not especially hard (low cost).

tomncooper: Right, I've updated the proposal to cover handling both annotations in both the old and new `Subject`. 

I have proposed that they share the same validation logic, via a utility class in the new `identity-api` module, so `@Unique` and `@SingularPrincipal` are recognised everywhere they could be mixed up. 

I have also proposed adding `@SingularPrincipal` as a meta-annotation on `@Unique`, that way the utility does a one-level meta-annotation scan so the new module never needs to import `@Unique` directly. The whole shared utility class is deprecated from birth and gets removed at 1.0 when `@Unique` goes away. 

I've also called out in the compatibility section that `PrincipalEntityNameMapper` in the entity isolation filter switches to using this shared utility too.

PR #120, opened 2026-07-06
Propose rack-aware SNI broker address patterns

EJASKHAN:
This proposal adds rack-aware-derived broker address generation to the SNI node identification strategy.

It allows advertised broker address patterns to include Kafka broker rack metadata using `$(rackId)`. The rack value can be translated through configured `rackIdMappings` into DNS/load-balancer-friendly advertised address labels, with `rackIdDefault` used when the upstream rack value is missing or unmapped.

The motivation is to support topology-aware DNS/load-balancer routing and reduce avoidable cross-zone traffic for high-volume Kafka produce/fetch workloads, without changing Kafka protocol rack metadata or cloud-provider placement semantics.

### DIFF
--- a/proposals/120-rack-aware-sni-broker-address-patterns.md
+++ b/proposals/120-rack-aware-sni-broker-address-patterns.md
@@ -1,4 +1,4 @@
-# 000 - Rack-aware SNI broker address patterns
+# 120 - Rack-aware SNI broker address patterns
 
 This proposal adds optional rack-aware broker address generation to the SNI node identification strategy.
 It allows an advertised broker address pattern to include broker rack metadata learned from Kafka metadata responses.
--- a/proposals/120-rack-aware-sni-broker-address-patterns.md
+++ b/proposals/120-rack-aware-sni-broker-address-patterns.md
@@ -64,7 +64,7 @@ Instead, it should use the rack value that Kafka already exposes.
 
 ## Proposal
 
-Extend the SNI node identification strategy so `advertisedBrokerAddressPattern` may use these additional tokens:
+Extend the SNI node identification strategy so `advertisedBrokerAddressPattern` may use this additional token:
 
 ```text
 $(rackId)
@@ -139,8 +139,7 @@ Provider-specific values such as availability-zone IDs or subnet-derived placeme
 Today the existing node identification strategies perform basic pattern, token, port, and URI-style checks.
 They do not perform full DNS label validation for generated advertised broker hostnames.
 This proposal does not change that validation model.
-If an advertised address fails the existing validation checks, address generation fails.
-Otherwise, Kroxylicious treats hostname suitability as an operator responsibility, as it does for existing advertised broker address patterns.
+Kroxylicious treats hostname suitability as an operator responsibility, as it does for existing advertised broker address patterns.
 Operators should use `rackIdMappings` to translate arbitrary upstream rack values into advertised address labels that are suitable for their DNS, certificate, and load-balancer conventions.
 
 ### Usage model
@@ -182,7 +181,9 @@ For a SNI configuration using `$(rackId)`, eager endpoint creation uses `rackIdD
 Once upstream metadata has been observed, rack-aware endpoints should be updated by metadata-driven endpoint reconciliation.
 
 If a broker rack value changes, Kroxylicious should treat that as a change in the generated advertised broker address during normal endpoint reconciliation.
-The new hostname should be generated from the updated metadata, and stale endpoint bindings generated from the previous rack value should be removed according to the existing reconciliation lifecycle.
+The new hostname should be generated from the updated metadata, and reconciliation must detect this as an endpoint change even though the node ID is unchanged.
+This requires rack-aware reconciliation logic; the current node-ID membership reconciliation is not sufficient on its own.
+This may be implemented by tracking the effective generated broker address, or by treating `(nodeId, effectiveRackLabel)` as the endpoint identity for rack-aware SNI bindings.
 Existing client connections can continue using the connection they already established, but a client that tries to open a new connection using a cached stale hostname may need to refresh metadata and retry.
 
 ### Endpoint registration and reverse mapping
@@ -211,7 +212,8 @@ Embedding rack labels into broker hostnames does not provide the same routing me
 This is a scoping choice rather than a fundamental limitation.
 If future work defines a useful rack-aware lifecycle and routing model for port-based listeners, it can be considered separately.
 It does not introduce cloud-provider integrations.
-It does not require users to configure rack-aware patterns.
+It does not introduce full DNS label validation for generated advertised broker hostnames.
+It does not require existing users to configure rack-aware patterns.
 
 ## Affected/not affected projects
 
--- a/proposals/120-rack-aware-sni-broker-address-patterns.md
+++ b/proposals/120-rack-aware-sni-broker-address-patterns.md
@@ -1,7 +1,7 @@
 # 120 - Rack-aware SNI broker address patterns
 
 This proposal adds rack-derived broker address generation to the SNI node identification strategy.
-It uses broker rack metadata learned from Kafka metadata responses and configured DNS label mappings to build advertised broker addresses without changing Kafka protocol rack metadata.
+It uses broker rack metadata learned from Kafka metadata responses, a required `rackIdDefault`, and optional `rackIdMappings` to build advertised broker addresses without changing Kafka protocol rack metadata.
 
 ## Current situation
 
@@ -42,11 +42,12 @@ In a multi-zone deployment, the client-facing load balancer, the Kroxylicious po
 For example:
 
 ```text
-Client → Network Load Balancer (NLB) endpoint in Zone A
-*(assuming cross-zone load balancing is disabled between the NLB and Kroxylicious pods)*
-→ Kroxylicious pod in Zone A
-→ Kafka broker or partition leader in Zone C
+client -> Network Load Balancer (NLB) endpoint in Zone A
+       -> Kroxylicious pod in Zone A
+       -> Kafka broker or partition leader in Zone C
 ```
+
+This example assumes cross-zone load balancing is disabled between the NLB and Kroxylicious pods.
 When traffic crosses zone boundaries between these hops, operators may pay inter-zone data transfer charges and may also see additional latency.
 This is especially relevant for Kafka because produce and fetch traffic can be high-volume.
 If broker-specific advertised hostnames include a rack or zone label, operators can configure DNS, load balancer targets, ingress policy, or deployment topology so that clients prefer a Kroxylicious endpoint in the same zone as the broker rack or failure domain where possible.
@@ -70,8 +71,8 @@ $(rackId)
 ```
 
 When Kroxylicious rewrites a metadata response, it uses the broker rack value from the metadata response for `$(rackId)`.
-If `rackIdMappings` is configured, Kroxylicious first translates the upstream rack value into an advertised address label.
-If the upstream rack value is missing or does not have a mapping, Kroxylicious uses `rackIdDefault` if configured.
+If `rackIdMappings` contains the upstream rack value, Kroxylicious translates it into the configured advertised address label.
+If the upstream rack value is missing or does not have a mapping, Kroxylicious uses `rackIdDefault`.
 
 Example configuration:
 
@@ -96,8 +97,8 @@ That advertised hostname can then be backed by operator-managed DNS or load-bala
 For example, `az1` can resolve to proxy endpoints in zone `az1`, while `az2` resolves to proxy endpoints in zone `az2`.
 The exact DNS and networking implementation remains outside Kroxylicious.
 
-If the broker rack value is unavailable or unmapped and no `rackIdDefault` is configured, address generation fails.
-This makes misconfiguration visible rather than silently advertising an incomplete or incorrect hostname.
+Using `$(rackId)` requires `rackIdDefault`.
+This gives Kroxylicious a deterministic advertised address label when the upstream rack value is unavailable or unmapped.
 
 ### Protocol-level consistency
 
@@ -123,13 +124,10 @@ The hostname label controls the network path to Kroxylicious.
 The Kafka protocol rack value continues to control Kafka features that depend on rack identity.
 Because Kroxylicious does not rewrite protocol rack values, features such as KIP-392 rack-aware fetching continue to use the upstream Kafka rack values and do not require reverse mapping by the proxy.
 
-If no `rackIdMappings` is configured, Kroxylicious uses the upstream Kafka rack value directly as the advertised address label.
-Operators should only use this mode when their broker rack values are already suitable for their advertised hostname pattern.
-
-If `rackIdMappings` is configured and the upstream rack value is missing or unmapped, Kroxylicious uses `rackIdDefault` if configured.
+If the upstream rack value is missing or unmapped, Kroxylicious uses `rackIdDefault`.
 The default is an advertised address fallback label.
 It should be used for cases where the operator prefers deterministic DNS behavior over failing address generation for unknown rack values.
-This means that, when mappings are configured, an unmapped upstream rack value is not passed directly into the advertised hostname.
+This means that an unmapped upstream rack value is not passed directly into the advertised hostname.
 
 ### DNS validity
 
@@ -169,8 +167,7 @@ Other response types can contain broker endpoints without rack metadata.
 For example, `FindCoordinatorResponse` identifies the coordinator by node ID, host, and port, but does not include broker rack information.
 
 For responses that contain broker endpoints but do not include rack metadata, Kroxylicious should use a rack value previously learned from metadata for the same node ID.
-If no learned rack value is available, Kroxylicious uses `rackIdDefault` if configured.
-If no learned rack value is available and no default is configured, address generation fails.
+If no learned rack value is available, Kroxylicious uses `rackIdDefault`.
 Using `rackIdDefault` can cause an early control-plane response to use the default rack label before metadata has been observed, but it preserves routing correctness because SNI reverse mapping remains based on node ID.
 
 ### Lifecycle
@@ -181,8 +178,8 @@ For other responses that contain broker endpoints, Kroxylicious can only generat
 
 This has implications for SNI endpoint lifecycle.
 Configurations that eagerly create broker endpoints from configured node ID ranges cannot generate rack-aware broker endpoints before upstream metadata has been observed.
-For a SNI configuration using `$(rackId)`, eager endpoint creation can only use `rackIdDefault`.
-If no default is configured, rack-aware endpoints should be created by metadata-driven endpoint reconciliation once upstream metadata has been observed.
+For a SNI configuration using `$(rackId)`, eager endpoint creation uses `rackIdDefault`.
+Once upstream metadata has been observed, rack-aware endpoints should be updated by metadata-driven endpoint reconciliation.
 
 If a broker rack value changes, Kroxylicious should treat that as a change in the generated advertised broker address during normal endpoint reconciliation.
 The new hostname should be generated from the updated metadata, and stale endpoint bindings generated from the previous rack value should be removed according to the existing reconciliation lifecycle.
@@ -244,8 +241,8 @@ The new behavior is opt-in:
 * Strategies that do not use rack metadata continue to generate broker addresses from node ID alone.
 
 There are compatibility considerations for `$(rackId)`.
-If the upstream Kafka cluster does not expose broker rack metadata, a pattern using `$(rackId)` cannot be resolved.
-Operators can configure `rackIdDefault` when they need deterministic fallback behavior for missing or unmapped rack values.
+If the upstream Kafka cluster does not expose broker rack metadata, or exposes a rack value without a configured mapping, `rackIdDefault` is used.
+Because of this, `rackIdDefault` is required when `$(rackId)` is used.
 
 The reverse SNI mapping must continue to extract exactly the node ID from the advertised broker hostname.
 Rack labels must not become part of the node identity.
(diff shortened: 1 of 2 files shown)


proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,224 @@
...
+Kroxylicious should not call cloud-provider APIs to interpret those values.
+Instead, it should use the rack value that Kafka already exposes and let operators optionally map raw rack values to DNS-friendly labels.
+
+## Proposal
+
+Extend the SNI node identification strategy so `advertisedBrokerAddressPattern` may use these additional tokens:
+
+```text
+$(rackId)
+$(rackId:<default>)
```
SamBarker: Defaulting in the token syntax will set a user expectation that all tokens
support inline defaults — `$(nodeId:<default>)`,
`$(virtualClusterName:<default>)`, etc. That's not the intent.

If we keep `rackIdMappings` in scope (see top-level comment for why I'd
suggest deferring it), defaulting belongs there — e.g. a reserved entry for
unknown/unmapped values — not in the templating language. If we ship
`$(rackId)` without mappings, the question becomes simpler: rack unavailable
means address generation fails, which is already the behaviour for
`$(rackId)` without a default.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,224 @@
...
+For example, both patterns still resolve node ID correctly:
+
+```text
+broker-$(nodeId)-$(rackId).example.net
+broker-$(rackId).$(nodeId).example.net
+```
+
+### Scope
+
+This proposal is limited to the SNI node identification strategy.
```
SamBarker, line 223: Could you add a sentence on why port mode is excluded? This states the
scope cut but not the rationale. Even something brief would help readers
understand whether this is a fundamental limitation or a scoping choice —
especially since solving the "rack unknown at generation time" lifecycle
(which the proposal needs to address for SNI's eager path anyway) would
largely remove the obstacle for port mode too.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,224 @@
...
+
+```yaml
+rackIdMappings:
+  euc1-az1: az1
+```
+
+This lets operators keep stable DNS labels even when raw Kafka rack values differ between environments or providers.
+It also avoids making Kroxylicious aware of cloud-provider-specific naming rules.
+
+If no mapping exists for a non-empty rack value, the raw rack value is used as-is.
```
SamBarker: This passthrough means an unmapped rack value goes straight into a DNS
hostname with no validation. Kafka `broker.rack` values are arbitrary
strings — `rack/shelf-3:unit_7` is a legal broker.rack value but not a
legal DNS label. The proposal should specify what happens when a rack value
(mapped or unmapped) would produce an invalid hostname.

(This also applies to `$(rackId)` without mappings — see top-level comment.)

SamBarker (commented):
Thanks for this — rack-aware addressing is something we've been asked about
and the use case is compelling. The decision to keep Kroxylicious out of
cloud-provider placement semantics and work from Kafka's own rack metadata
is exactly right.

My main concern is that the proposal doesn't yet present a coherent
abstraction for rack-aware addressing.

A few things that need to hang together but currently don't:

**Protocol-level consistency.** `rackIdMappings` maps rack values in the
advertised address but leaves the Kafka protocol `rack` field (in
MetadataResponse, DescribeClusterResponse) unmapped. A client connecting to
`broker-0-az1.example.net` sees `rack=euc1-az1` in the protocol — two
naming schemes for the same concept. And if we do map the protocol field,
we also need to reverse-map `rackId` in FetchRequest for KIP-392 rack-aware
consumption to keep working. The proxy should present one consistent view
of a broker's rack identity to clients, not a mix of mapped and unmapped
values depending on where you look.

**Lifecycle.** The `nodeIdRanges` eager path creates endpoints before any
upstream metadata exists, so "rack unknown at generation time" is a real
path through the SNI strategy as proposed. The compatibility section
acknowledges this via `$(rackId:<default>)` but doesn't specify the full
lifecycle: what's advertised initially, how and when it updates if the real
value differs from the default, what happens to clients that resolved the
previous hostname. Similarly, a broker can change racks — what does the
proxy do with endpoints that were generated from a now-stale rack value?
These aren't edge cases — they're what it means to own a rack-aware
identity on behalf of the upstream cluster.

**DNS validity.** Raw rack values go straight into hostnames (either
directly, or via the `rackIdMappings` passthrough for unmapped values).
Kafka `broker.rack` values are arbitrary strings — the proposal should
specify validation of the resulting hostname and what happens when a value
would produce an invalid DNS label.

**Port-mode exclusion.** `portIdentifiesNode` faces the same "rack unknown
at generation time" situation, just unconditionally — it has no reactive
discovery path at all. Whatever mechanism solves the lifecycle for SNI's
eager case would work for port mode too. I'm not asking this proposal to do
both, but it would help to state *why* port mode is excluded and acknowledge
that it becomes a natural follow-on once the lifecycle question is answered.

I think there's a solid feature here — it may be that the right path is to
reduce the initial scope so that what ships is fully coherent, and build up
from there. But I'd rather hear your thoughts on how to address the
consistency and lifecycle questions than prescribe a particular
decomposition.

proposals/120-rack-aware-sni-broker-address-patterns.md
EJASKHAN: Thanks for the detailed feedback. I have updated the proposal to make the abstraction and lifecycle clearer.

The proposal remains focused on **rack-aware advertised address generation**, not Kafka rack identity translation. `rackIdMappings` maps upstream rack values to labels used in advertised broker hostnames for DNS, certificate, and load-balancer routing. It does not modify rack information in Kafka protocol requests or responses.

For example, a client may connect to `broker-0-az1.example.net` while still seeing `rack=euc1-az1` in Kafka metadata. Here, `az1` is the network routing label, while `euc1-az1` remains the Kafka rack identity.

I also clarified the lifecycle. Broker addresses are generated from the latest rack metadata known to Kroxylicious. If the rack is not yet known, `rackIdDefault` is used if configured; otherwise, address generation fails. When rack metadata changes, the advertised hostname and SNI bindings are updated through normal reconciliation. Existing connections can continue, while clients discover the new hostname through normal Kafka metadata refresh and reconnect behavior.

Kafka rack values may not be DNS-safe. Operators can use `rackIdMappings` to map arbitrary rack values to labels suitable for their DNS, certificate, and load-balancer setup. If mappings are configured and a rack value is not mapped, `rackIdDefault` is used if configured; otherwise, address generation fails.

Finally, I clarified why `portIdentifiesNode` is excluded. This proposal focuses on SNI-based routing, where the hostname is used for routing. In port mode, broker identity is primarily determined by the port number, typically using a shared hostname. Port-based rack-aware addressing could be considered separately if a clear routing and lifecycle model is identified.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,283 @@
...
+For metadata responses, the rack value is available in the response being rewritten.
+For other responses that contain broker endpoints, Kroxylicious can only generate a rack-aware address after it has previously learned the broker's rack from metadata.
+
+This has implications for SNI endpoint lifecycle.
+Configurations that eagerly create broker endpoints from configured node ID ranges cannot generate rack-aware broker endpoints before upstream metadata has been observed.
+For a SNI configuration using `$(rackId)`, eager endpoint creation uses `rackIdDefault`.
+Once upstream metadata has been observed, rack-aware endpoints should be updated by metadata-driven endpoint reconciliation.
+
+If a broker rack value changes, Kroxylicious should treat that as a change in the generated advertised broker address during normal endpoint reconciliation.
+The new hostname should be generated from the updated metadata, and stale endpoint bindings generated from the previous rack value should be removed according to the existing reconciliation lifecycle.
```
SamBarker: The current endpoint reconciliation in the runtime is keyed purely on node-ID set membership — nodes appearing in or disappearing from metadata. If a node ID stays constant but its rack value changes, `doReconcile` sees the node ID is already bound and does nothing. The existing SNI binding is left untouched.

So "existing reconciliation lifecycle" doesn't cover the rack-change case. If broker 0's rack flips from `euc1-az1` to `euc1-az2`, the metadata response rewrite will correctly use the new rack value for that individual response, but the SNI endpoint registration still points at the old hostname (`broker-0-az1.example.net`). A client that receives the new advertised hostname (`broker-0-az2.example.net`) and tries to connect would fail because no SNI binding matches.

This proposal needs new reconciliation logic — keying on `(nodeId, rack)` rather than just `nodeId`, or triggering a rebind when a tracked attribute changes. That's fine, but the proposal should call it out as new work rather than framing it as covered by existing infrastructure.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,283 @@
...
+Not every legal Kafka rack value is valid inside a DNS hostname.
+For example, `rack/shelf-3:unit_7` may be a legal broker rack value, but it is not a valid DNS label.
+Provider-specific values such as availability-zone IDs or subnet-derived placement labels may also be unsuitable for the operator's desired DNS naming scheme.
+
+Today the existing node identification strategies perform basic pattern, token, port, and URI-style checks.
+They do not perform full DNS label validation for generated advertised broker hostnames.
+This proposal does not change that validation model.
+If an advertised address fails the existing validation checks, address generation fails.
+Otherwise, Kroxylicious treats hostname suitability as an operator responsibility, as it does for existing advertised broker address patterns.
+Operators should use `rackIdMappings` to translate arbitrary upstream rack values into advertised address labels that are suitable for their DNS, certificate, and load-balancer conventions.
```
SamBarker: The position that hostname suitability is an operator responsibility is reasonable for this proposal — `$(rackId)` doesn't change the validation model, it just widens the surface for values that aren't DNS-safe.

But it's worth noting that the runtime doesn't currently validate DNS validity of generated advertised hostnames for *any* token, including existing ones like `$(virtualClusterName)`. If an operator configures a virtual cluster name or a rack mapping value that isn't a valid DNS label, the first signal is a DNS resolution failure at the client — which is a late and confusing place to discover a configuration mistake.

I don't think this proposal needs to solve that — it's a pre-existing gap. I've raised [kroxylicious#4323](https://github.com/kroxylicious/kroxylicious/issues/4323) to track validating DNS validity of generated advertised broker hostnames generally.

proposals/120-rack-aware-sni-broker-address-patterns.md
EJASKHAN: Added the rationale in the Scope section.

proposals/120-rack-aware-sni-broker-address-patterns.md
EJASKHAN: Updated the proposal so raw rack values are not passed through directly.

proposals/120-rack-aware-sni-broker-address-patterns.md
EJASKHAN: Agreed. Updated the lifecycle section to call this out as new rack-aware reconciliation work.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,285 @@
...
+broker-$(nodeId)-$(rackId).example.net
+broker-$(rackId).$(nodeId).example.net
+```
+
+### Scope
+
+This proposal is limited to the SNI node identification strategy.
+It does not change the port-based node identification strategy.
+Port mode is excluded because this proposal is about hostname/SNI-based routing.
+In `portIdentifiesNode`, broker identity is primarily represented by the port number, often using a shared DNS name for all brokers.
```
SamBarker, line 226: Thanks for adding the rationale. The shared-hostname case makes sense — if all brokers share a hostname, there's nothing useful to put a rack label in.

But PPN already supports per-broker hostnames via `advertisedBrokerAddressPattern` with `$(nodeId)` — e.g. `broker-$(nodeId).example.net` gives `broker-0.example.net:9093`, `broker-1.example.net:9094`. The rationale assumes PPN means shared hostnames, but it doesn't have to. Behind a load balancer with per-broker hostnames, `broker-0-az1.example.net:9093` would be just as useful for DNS/routing as it is in SNI mode.

The heavy lifting in this proposal is the lifecycle and reconciliation model — learning rack values from metadata, tracking changes, rebinding endpoints. That work applies to both strategies. PPN already has the hostname customisation infrastructure; it just needs the `$(rackId)` token added.

I'd like to hear from the other maintainers on whether we should include PPN while we're here or whether scoping to SNI only is the right call.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,285 @@
...
+If `rackIdMappings` contains the upstream rack value, Kroxylicious translates it into the configured advertised address label.
+If the upstream rack value is missing or does not have a mapping, Kroxylicious uses `rackIdDefault`.
+
+Example configuration:
+
+```yaml
+sniHostIdentifiesNode:
+  bootstrapAddress: "cluster.example.net:9192"
+  advertisedBrokerAddressPattern: "broker-$(nodeId)-$(rackId).example.net:9192"
+  rackIdDefault: az1
```
SamBarker: The `rackId` prefix on these config fields (`rackIdDefault`, `rackIdMappings`) suggests they're about rack identity, but they're not — they're about address generation. The proposal goes to some length in the Protocol-level consistency section to explain why protocol rack fields aren't rewritten, but the config naming works against that message. A reader seeing `rackIdMappings` reasonably asks "if we're mapping rack IDs, why aren't we mapping them everywhere?"

Consider `$(rackAddress)` for the token — it says what gets substituted into the hostname. The config fields follow naturally: `defaultRackAddress` for the fallback, and `rackAddressMappings` for the lookup table. In the expanded form, the mapping direction is self-documenting:

```yaml
advertisedBrokerAddressPattern: "broker-$(nodeId)-$(rackAddress).example.net:9192"
defaultRackAddress: az1
rackAddressMappings:
  - rackId: euc1-az1
    rackAddress: az1
  - rackId: euc1-az2
    rackAddress: az2
```

An operator reading this immediately sees: `rackId` is the Kafka rack value, `rackAddress` is the hostname label. The naming carries the intent — these are address labels derived from rack metadata, not rack identities — and the Protocol-level consistency section becomes supporting detail rather than a necessary correction of a first impression.

EJASKHAN: Valid suggestion, updated the same.

proposals/120-rack-aware-sni-broker-address-patterns.md
SamBarker: I think this is good as it stands. I'd rather we were consistent between identity schemes, but will go with the consensus from others.

SamBarker (approved):
I think this is good as it stands. I'd rather we were consistent between identity schemes, but will go with the consensus from others.

proposals/120-rack-aware-sni-broker-address-patterns.md
```diff
@@ -0,0 +1,302 @@
...
+
+When handling Kafka metadata responses, Kroxylicious already rewrites upstream broker host and port values into client-facing broker addresses.
+This proposal extends that path to pass the broker rack value to the node identification strategy when generating the advertised broker address.
+
+The rack value is only available in responses that carry broker rack metadata, such as metadata responses.
+Other response types can contain broker endpoints without rack metadata.
+For example, `FindCoordinatorResponse` identifies the coordinator by node ID, host, and port, but does not include broker rack information.
+
+For responses that contain broker endpoints but do not include rack metadata, Kroxylicious should use a rack value previously learned from metadata for the same node ID.
+If no learned rack value is available, Kroxylicious uses `defaultRackAddress`.
```
k-wall, line 172: I don't think this can actually happen.  The EagerMetadataLearner ensures that the channel has the upstream topology before real kafka work begins.  It monitors the RPCs going down the channel and interjects is own MetadataRequest if the client hasn't sent one itself.   So I think if the upstream Kafka cluster has rack information, the responses sent by the proxy will always include it.

k-wall:
Let me say first, I really like the proposal.  Thank you for the work.

However, I think there's an issue in the scenario where client connections exist when the proxy is restarted.  We've been here before.

1. Let's say client A has a connection to broker-0-az2.example.net.  Proxy is configured with `defaultRackAddress: az1`
2. Proxy restarted
3. Client A tries to reconnect to `broker-0-az2.example.net`.  It fails because the SNI used by the client broker-0-az2.example.net does not match any endpoint binding know to the proxy.  
4. Client A is now spinning trying to reconnect to `broker-0-az2.example.net`. 

Now if **another** client happens to connect the the proxy at a matching address (bootstrap or broker), the proxy will learn the metadata and from that point Client A's connection attempts to `broker-0-az2.example.net` will spontaneously succeed. Happy days.

But what if there is no other client in the system?  Client A is stuck.  Client A will need to be restarted to force it to begin again from bootstrap.  We learnt back in https://github.com/kroxylicious/kroxylicious/pull/447 that Kafka clients don't "give up on a broker address/revert to bootstrap" when disconnects occur.   They keep dialing the same broker address - over and over. (I tested librdkafka and Apache Kafka back then.  I haven't repeated the test today) 

I'm wondering if we could hook into the solution we found back then.  If the proxy accepted the connection with the mismatched rack but used a `MetadataDiscoveryBrokerEndpointBinding` to root the traffic to bootstrap, that allows the proxy to learn the metadata then forces the connection closed (EagerMetadataLeaner#73).  The client retries again.  This time the SNI matches.  Happy days.

WDYT?

SamBarker:
Great catch Keith, and thanks for connecting this back to #447. I agree that `defaultRackAddress` is problematic — thinking it through, it doesn't have a compelling role in any scenario. If brokers don't report rack, there's no reason to use `$(rackAddress)` in the pattern at all. If they do, `defaultRackAddress` creates transient wrong addresses on cold start and actively harmful ones on restart.

I think the right model is to drop `defaultRackAddress` entirely and wire in `MetadataDiscoveryBrokerEndpointBinding` for the pre-metadata phase. That handles both cold start and post-restart with stale clients through the same mechanism — no guessing a default rack.

proposals/120-rack-aware-sni-broker-address-patterns.md
SamBarker: Agreed — `EagerMetadataLearner` ensures rack info is available before real Kafka work begins, so the `defaultRackAddress` fallback path in response rewriting shouldn't be reachable. This ties into the issue comment discussion — dropping `defaultRackAddress` altogether in favour of `MetadataDiscoveryBrokerEndpointBinding` for the pre-metadata phase seems like the cleaner model.

k-wall:
> Great catch Keith, and thanks for connecting this back to #447. I agree that `defaultRackAddress` is problematic — thinking it through, it doesn't have a compelling role in any scenario. If brokers don't report rack, there's no reason to use `$(rackAddress)` in the pattern at all. If they do, `defaultRackAddress` creates transient wrong addresses on cold start and actively harmful ones on restart.
> 
> I think the right model is to drop `defaultRackAddress` entirely and wire in `MetadataDiscoveryBrokerEndpointBinding` for the pre-metadata phase. That handles both cold start and post-restart with stale clients through the same mechanism — no guessing a default rack.

Agreed,  I hadn't twigged when I wrote the comment that the need for `defaultRackAddress` goes away entirely.   But I agree it does.  One less lever! Hurray!

@EJASKHAN does this make sense to you?  Do you see any holes?

EJASKHAN:
> > Great catch Keith, and thanks for connecting this back to #447. I agree that `defaultRackAddress` is problematic — thinking it through, it doesn't have a compelling role in any scenario. If brokers don't report rack, there's no reason to use `$(rackAddress)` in the pattern at all. If they do, `defaultRackAddress` creates transient wrong addresses on cold start and actively harmful ones on restart.
> > I think the right model is to drop `defaultRackAddress` entirely and wire in `MetadataDiscoveryBrokerEndpointBinding` for the pre-metadata phase. That handles both cold start and post-restart with stale clients through the same mechanism — no guessing a default rack.
> 
> Agreed, I hadn't twigged when I wrote the comment that the need for `defaultRackAddress` goes away entirely. But I agree it does. One less lever! Hurray!
> 
> @EJASKHAN does this make sense to you? Do you see any holes?

Thanks Keith and Sam. Agreed on reusing the #447 metadata-discovery path whenever a broker SNI matches the configured pattern but has no registered endpoint, covering both cold starts and restart recovery. 

Keeping defaultRackAddress optional would provide an availability fallback for missing or unexpectedly unmapped rack values. When configured, routing may be non-optimal until the mapping is updated; when omitted, address generation would fail explicitly. Would this separation address the concern?

SamBarker:
> Thanks Keith and Sam. Agreed on reusing the #447 metadata-discovery path whenever a broker SNI matches the configured pattern but has no registered endpoint, covering both cold starts and restart recovery.
> 
> Keeping defaultRackAddress optional would provide an availability fallback for missing or unexpectedly unmapped rack values. When configured, routing may be non-optimal until the mapping is updated; when omitted, address generation would fail explicitly. Would this separation address the concern?

OHh right, yes thats a valid use case. We need to know what to do if a broker advertises its rack as `rack-A` but we only have mappings for `rack-1`, `rack-2` etc.

EJASKHAN:
> > Thanks Keith and Sam. Agreed on reusing the #447 metadata-discovery path whenever a broker SNI matches the configured pattern but has no registered endpoint, covering both cold starts and restart recovery.
> > Keeping defaultRackAddress optional would provide an availability fallback for missing or unexpectedly unmapped rack values. When configured, routing may be non-optimal until the mapping is updated; when omitted, address generation would fail explicitly. Would this separation address the concern?
> 
> OHh right, yes thats a valid use case. We need to know what to do if a broker advertises its rack as `rack-A` but we only have mappings for `rack-1`, `rack-2` etc.

Updated the design doc with the details.

SamBarker:
> 
> Updated the design doc with the details.  

The pr was closed when you added this comment @EJASKHAN I presume that was an accident?

EJASKHAN:
> > Updated the design doc with the details.
> 
> The pr was closed when you added this comment @EJASKHAN I presume that was an accident?

Yes, it was unintentional. I’ve reopened the PR now. Thanks for pointing it out!

SamBarker (approved):
LGTM

PR #121, opened 2026-07-08
Enforce minimum access rights on confidential files

piotrpdev:
Kroxylicious should validate that confidential files (TLS private keys, keystores, truststores,
password files, and KMS credential files) have suitably restrictive filesystem permissions before
reading them, similar to how the `ssh` command refuses to use a world-readable private key.

Related to: https://github.com/kroxylicious/kroxylicious/issues/1038

### DIFF
--- a/proposals/121-enforce-minimum-file-permissions.md
+++ b/proposals/121-enforce-minimum-file-permissions.md
@@ -1,0 +1,173 @@
+# 121 - Enforce minimum access rights on confidential files
+
+Kroxylicious should validate that confidential files (TLS private keys, keystores, truststores,
+password files, and KMS credential files) have suitably restrictive filesystem permissions before
+reading them, similar to how the `ssh` command refuses to use a world-readable private key.
+
+## Current situation
+
+Kroxylicious reads confidential material from the filesystem - TLS private keys, keystores,
+truststores, and passwords - without checking whether those files are accessible by users other
+than the owner. A world-readable private key file (`0644`) is silently accepted and used. This
+violates the principle of least privilege and increases the risk of credential exposure through
+over-permissive filesystem configurations.
+
+## Motivation
+
+Security best practices require that private key material is accessible only to the process that
+owns it. Tools such as `ssh`, `gpg`, and many TLS libraries enforce this by refusing to operate
+on files with group or other read bits set. Kroxylicious should provide equivalent protection.
+
+The threat model includes:
+- Accidental over-permissive file creation (e.g. default `umask` producing `0644`).
+- Multi-tenant environments where other users on the same host could read Kroxylicious credentials.
+- Kubernetes deployments where secret volumes default to world-readable `0644` unless explicitly
+  configured otherwise.
+
+## Proposal
+
+### File permission policy
+
+Introduce a configurable `security.filePermissions.policy` setting with three modes:
+
+- `STRICT` - files must be owner-only (equivalent to `chmod 400` or `chmod 600`). Any group or
+  other read/write/execute bits cause an `IllegalStateException` at startup. Mirrors SSH behaviour.
+- `RELAXED` - other-user bits are forbidden, but group bits are permitted. This supports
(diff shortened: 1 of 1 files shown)


proposals/121-enforce-minimum-file-permissions.md
```diff
@@ -0,0 +1,173 @@
...
+- Accidental over-permissive file creation (e.g. default `umask` producing `0644`).
+- Multi-tenant environments where other users on the same host could read Kroxylicious credentials.
+- Kubernetes deployments where secret volumes default to world-readable `0644` unless explicitly
+  configured otherwise.
+
+## Proposal
+
+### File permission policy
+
+Introduce a configurable `security.filePermissions.policy` setting with three modes:
```
tombentley: The RELAXED option has a bootstrapping problem: the file permission policy is itself stored in a config file. If that config file is writable by someone other than the owner, they can change `policy: STRICT` to `policy: DISABLED` (or `RELAXED`), and then the permissive secret files are accepted. The policy is only as strong as the protection on the config file that specifies it.

piotrpdev: Nice catch, mentioned this in [`e7459da` (this PR)](https://github.com/kroxylicious/design/pull/121/commits/e7459dac0ecdff027a73d00ab96043d2c5f1a138)

proposals/121-enforce-minimum-file-permissions.md
```diff
@@ -0,0 +1,282 @@
...
+- Kubernetes deployments where secret volumes default to world-readable `0644` unless explicitly
+  configured otherwise.
+
+## Proposal
+
+### File permission policies
+
+Three policy modes are available:
+
+- `STRICT` - files must be owner-only (equivalent to `chmod 400` or `chmod 600`). Any group or
```
k-wall, line 33: We have seen some users on Windows.  Whilst it is not a platform we currently test on, we don't want to disenfranchise them.    I think it would be accepted for this feature to apply only to a POSIX filesystem, but we should decide how the feature will behave on other systems.

k-wall:
There's quite a lot of friction between this proposal and https://github.com/kroxylicious/design/pull/96. 

#96 breaks up the config file.  There's a single passwords.yaml file for passwords and uses a directory tree for  plugin configuration fragments.  Plugins may have sidecar files (truststores, keystores etc).  Rather than individual KMS (say) having responsibility for loading keystore and truststores, there are new KeyMaterialProvider and TrustMaterialProviders which are loaded as plugins then referenced.  In #96 would the enforcement points for permission checks would be significantly different to today (org.apache.logging.log4j.core.net.ssl.PasswordProvider would be gone).

I think as a project we need to confirm that #96 is the direction we want (I think we've said yes already as we see it critical for v1).  If that's the case, then I think this proposal needs to be closed.  Once #96 is delivered, a new proposal can be raised implementing permissions restrictions in the new model.

I'd suggest a discussion in slack and/or a slot on the community call.

proposals/121-enforce-minimum-file-permissions.md
piotrpdev: Added compatibility note for this in: [`a7500a7` (this PR)](https://github.com/kroxylicious/design/pull/121/commits/a7500a71f079552c3d1b2c5a0434d4d122d13718)

piotrpdev:
> There's quite a lot of friction between this proposal and #96.

Hmm, I think those could just be resolved by:

- Changing where the `validate()` calls are made (the specific call sites aren't a core part of this proposal) 
  - `NettyKeyProvider`/`NettyTrustProvider` calls would move to the new `ResourceDeserializer` for `KeyMaterialProvider`/`TrustMaterialProvider` sidecar files
  - `FilePassword` validation would move to the `passwords.yaml` reader (validating permissions on that file)
- Moving `security.filePermissions`  into #96's `proxy.yaml` under global proxy settings

As far as I can tell, the validator, the config model, the exit code, and the operator condition are all reusable as is.

PR #122, opened 2026-07-16
docs: propose ClientAddressing sealed interface for RouterContext

SamBarker:
## Summary

Proposes a `ClientAddressing` sealed interface to replace the opaque `VirtualNode`
marker in the Router API (refining Proposal 070). The sealed hierarchy uses Kafka
domain types — `Bootstrap` for connections to the virtual cluster's bootstrap address,
`Broker` for broker-specific connections — making the distinction compiler-checkable
and self-documenting.

Also proposes:
- `clientAddressed()` replacing `Optional<VirtualNode> virtualNode()`
- `sendToRoute()` collapsing the `anyNode()` + `sendRequest()` two-step
- `brokerFor(int)` replacing `nodeForId(int)`
- `sendRequest(Broker, ...)` overload alongside deprecated `sendRequest(VirtualNode, ...)`

All deprecated methods receive default implementations that delegate to their
replacements — no runtime breaks during the migration window.

## Context

The bootstrap/broker distinction surfaced independently from two directions:
the port-binding separation work (#4147) arrived at a sealed two-case type
from the networking layer, and review of the dynamic routing PR (#4307,
[discussion](https://github.com/kroxylicious/kroxylicious/pull/4307#discussion_r3533660951))
raised the question of what state `VirtualNode` should expose. This proposal
addresses both by replacing `VirtualNode` with a domain-typed sealed hierarchy.

### DIFF
--- a/proposals/122-router-context-endpoint-type.md
+++ b/proposals/122-router-context-endpoint-type.md
@@ -1,4 +1,4 @@
-# 000 - Router API: `ClientAddressing` sealed interface
+# 122 - Router API: `ClientAddressing` sealed interface
 
 This proposal refines the Router API introduced in [Proposal 070](070-routing-api.md)
 by introducing a `ClientAddressing` sealed interface that replaces the opaque
--- a/proposals/000-router-context-endpoint-type.md
+++ b/proposals/000-router-context-endpoint-type.md
@@ -1,0 +1,366 @@
+# 000 - Router API: `ClientAddressing` sealed interface
+
+This proposal refines the Router API introduced in [Proposal 070](070-routing-api.md)
+by introducing a `ClientAddressing` sealed interface that replaces the opaque
+`VirtualNode` marker with domain-typed alternatives: `Bootstrap` for connections to the
+virtual cluster's bootstrap address and `Broker` for broker-specific connections.
+
+## Current situation
+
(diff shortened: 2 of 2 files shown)


proposals/122-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+extending `ClientAddressing`, with `Broker` and `Controller` as its subtypes. This made
+the `id()` contract explicit on `VirtualNode` and allowed pattern matching at two levels
+of granularity.
+
+This was rejected for two reasons. First, `Controller` is hypothetical — no current use
+case requires controller-specific connections, so the second level exists only to
+accommodate a speculative future type. Second, the flat hierarchy is strictly simpler and
+does not foreclose the two-level design: if `Controller` is needed later, a unifying
+`VirtualNode` interface can be introduced alongside it without breaking existing code
+that uses `Broker`.
```
SamBarker, line 304: Is `Controller` as a future peer subtype well-served by the flat hierarchy, or does the two-level hierarchy (rejected here) better justify itself despite `Controller` being hypothetical? The flat design doesn't foreclose the two-level option later, but reviewers familiar with the KRaft roadmap may have a stronger view on whether that extensibility point is worth pre-building.

PR #123, opened 2026-07-16
Replace opaque VirtualNode with domain-typed connection addressing

SamBarker:
## Summary

Proposes a `ClientAddressing` sealed interface to replace the opaque `VirtualNode`
marker in the Router API (refining Proposal 070). The sealed hierarchy uses Kafka
domain types — `Bootstrap` for connections to the virtual cluster's bootstrap address,
`Broker` for broker-specific connections — making the distinction compiler-checkable
and self-documenting.

Also proposes:
- `clientAddressed()` replacing `Optional<VirtualNode> virtualNode()`
- `sendToRoute()` collapsing the `anyNode()` + `sendRequest()` two-step
- `brokerFor(int)` replacing `nodeForId(int)`
- `sendRequest(Broker, ...)` overload alongside deprecated `sendRequest(VirtualNode, ...)`

All deprecated methods receive default implementations that delegate to their
replacements — no runtime breaks during the migration window.

## Context

The bootstrap/broker distinction surfaced independently from two directions:
the port-binding separation work (#4147) arrived at a sealed two-case type
from the networking layer, and review of the dynamic routing PR (#4307,
[discussion](https://github.com/kroxylicious/kroxylicious/pull/4307#discussion_r3533660951))
raised the question of what state `VirtualNode` should expose. This proposal
addresses both by replacing `VirtualNode` with a domain-typed sealed hierarchy.

### DIFF
--- a/proposals/123-router-context-endpoint-type.md
+++ b/proposals/123-router-context-endpoint-type.md
@@ -1,4 +1,4 @@
-# 122 - Router API: `ClientAddressing` sealed interface
+# 123 - Router API: `ClientAddressing` sealed interface
 
 This proposal refines the Router API introduced in [Proposal 070](070-routing-api.md)
 by introducing a `ClientAddressing` sealed interface that replaces the opaque
--- a/proposals/123-router-context-endpoint-type.md
+++ b/proposals/123-router-context-endpoint-type.md
@@ -104,17 +104,32 @@ public record Bootstrap() implements ClientDestination {}
 source-compatible for routers that pattern-match on `Bootstrap`.
 
 **`Broker`** — the client connected to a broker-specific address. `Broker` replaces the
-opaque `VirtualNode` marker with a domain-typed record carrying the virtual node ID:
+opaque `VirtualNode` marker with a domain-typed class. Construction is controlled by the
+runtime; router authors receive `Broker` instances but cannot instantiate them directly:
 
 ```java
-public record Broker(int id) implements ClientDestination {}
+public final class Broker implements ClientDestination {
+    private final int id;
+
+    // package-private constructor — only the runtime creates Broker instances
+    Broker(int id) { this.id = id; }
+
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof Broker other && this.id == other.id;
+    }
+
+    @Override
+    public int hashCode() {
+        return Integer.hashCode(id);
+    }
+}
 ```
 
-The `id` is the virtual (client-facing) node ID assigned by the proxy — it is not the
-upstream (target-cluster) node ID. The Kafka wire protocol serializes node identity as
-integers, so `brokerFor(int)` bridges from protocol messages into the typed handle and
-`id` provides the reverse when routers need the integer form. The Router API still
-prefers working in terms of `Broker` instances rather than raw integers.
+The internal node ID is not exposed as a public accessor. `Broker` is an opaque,
+equality-comparable handle — router authors use it to key maps (e.g. correlating
+`METADATA` responses with routing decisions) and pass it to `sendRequest(Broker, ...)`.
+`brokerFor(int)` remains the sole bridge from wire-protocol integers into `Broker` handles.
 
 ### Replace `virtualNode()` with `clientDestination()`
 
--- a/proposals/123-router-context-endpoint-type.md
+++ b/proposals/123-router-context-endpoint-type.md
@@ -1,7 +1,7 @@
-# 123 - Router API: `ClientAddressing` sealed interface
+# 123 - Router API: `ClientDestination` sealed interface
 
 This proposal refines the Router API introduced in [Proposal 070](070-routing-api.md)
-by introducing a `ClientAddressing` sealed interface that replaces the opaque
+by introducing a `ClientDestination` sealed interface that replaces the opaque
 `VirtualNode` marker with domain-typed alternatives: `Bootstrap` for connections to the
 virtual cluster's bootstrap address and `Broker` for broker-specific connections.
 
@@ -83,13 +83,13 @@ domain distinction between "a specific Kafka broker" and "a proxy discovery addr
 
 ## Proposal
 
-### `ClientAddressing` sealed interface
+### `ClientDestination` sealed interface
 
 ```java
-public sealed interface ClientAddressing permits Bootstrap, Broker {}
+public sealed interface ClientDestination permits Bootstrap, Broker {}
 ```
 
-`ClientAddressing` represents how the client addressed the proxy — either via a
+`ClientDestination` represents how the client addressed the proxy — either via a
 bootstrap discovery address or via a specific broker. The name reflects that this is
 determined by the client's choice of address, not by the router's decision.
 
@@ -97,7 +97,7 @@ determined by the client's choice of address, not by the router's decision.
 not a Kafka domain concept. There is no specific node and no node ID:
 
 ```java
-public record Bootstrap() implements ClientAddressing {}
+public record Bootstrap() implements ClientDestination {}
 ```
 
 `Bootstrap` is intentionally empty. Adding fields later (e.g. for observability) is
@@ -107,7 +107,7 @@ source-compatible for routers that pattern-match on `Bootstrap`.
 opaque `VirtualNode` marker with a domain-typed record carrying the virtual node ID:
 
 ```java
-public record Broker(int id) implements ClientAddressing {}
+public record Broker(int id) implements ClientDestination {}
 ```
 
 The `id` is the virtual (client-facing) node ID assigned by the proxy — it is not the
@@ -116,17 +116,17 @@ integers, so `brokerFor(int)` bridges from protocol messages into the typed hand
 `id` provides the reverse when routers need the integer form. The Router API still
 prefers working in terms of `Broker` instances rather than raw integers.
 
-### Replace `virtualNode()` with `clientAddressed()`
+### Replace `virtualNode()` with `clientDestination()`
 
 ```java
 // Replaces Optional<VirtualNode> virtualNode()
-ClientAddressing clientAddressed();
+ClientDestination clientDestination();
 ```
 
 Router authors use pattern matching:
 
 ```java
-switch (ctx.clientAddressed()) {
+switch (ctx.clientDestination()) {
     case Bootstrap b   -> /* discovery path — forward to any broker on a route */
     case Broker broker -> /* broker-specific path — forward to that broker */
 }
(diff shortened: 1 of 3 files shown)


proposals/123-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+extending `ClientAddressing`, with `Broker` and `Controller` as its subtypes. This made
+the `id()` contract explicit on `VirtualNode` and allowed pattern matching at two levels
+of granularity.
+
+This was rejected for two reasons. First, `Controller` is hypothetical — no current use
+case requires controller-specific connections, so the second level exists only to
+accommodate a speculative future type. Second, the flat hierarchy is strictly simpler and
+does not foreclose the two-level design: if `Controller` is needed later, a unifying
+`VirtualNode` interface can be introduced alongside it without breaking existing code
+that uses `Broker`.
```
SamBarker, line 320: Is `Controller` as a future peer subtype well-served by the flat hierarchy, or does the two-level hierarchy (rejected here) better justify itself despite `Controller` being hypothetical? The motivation for `Controller` is the proxy standing between members of the Kafka cluster — not just between clients and brokers. The flat design doesn't foreclose the two-level option later, but reviewers with a view on whether the proxy will need to mediate controller connections may have a stronger opinion on whether that extensibility point is worth pre-building.

proposals/123-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+first-class entities with integer node IDs. Bootstrap, by contrast, is not a Kafka
+domain concept but a proxy-level addressing mechanism — the proxy exposes a discovery
+address that does not correspond to a specific cluster node.
+
+The current API collapses both into a single opaque `VirtualNode` type, erasing the
+domain distinction between "a specific Kafka broker" and "a proxy discovery address."
+
+## Proposal
+
+### `ClientAddressing` sealed interface
```
k-wall: Not sure about the name `Addressing`.   Class/interface names are usually nouns rather than gerunds.  I wondered about:

- `ClientDestination` 
- `ClientAddressTarget`

proposals/123-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+envisaged `anyNode()` as a node-selection step separate from request dispatch, but no
+use case has emerged where a router selects a node and then decides later whether to
+actually send to it.
+
+### 3. `VirtualNode` carries no identity
+
+Because `VirtualNode` is an opaque marker, routers cannot reason about which broker
+they are targeting at the API level. This is intentional for some cases (the runtime
+selects the broker for bootstrap requests) but not for broker-specific connections,
+where the broker identity is meaningful and should be surfaced.
```
k-wall, line 72: "should be surfaced"

out of interest, have you a use-case in mind?

proposals/123-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+
+```java
+// Replaces Optional<VirtualNode> virtualNode()
+ClientAddressing clientAddressed();
+```
+
+Router authors use pattern matching:
+
+```java
+switch (ctx.clientAddressed()) {
```
k-wall: I do think this reads better.

proposals/123-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+**`Broker`** — the client connected to a broker-specific address. `Broker` replaces the
+opaque `VirtualNode` marker with a domain-typed record carrying the virtual node ID:
+
+```java
+public record Broker(int id) implements ClientAddressing {}
+```
+
+The `id` is the virtual (client-facing) node ID assigned by the proxy — it is not the
+upstream (target-cluster) node ID. The Kafka wire protocol serializes node identity as
+integers, so `nodeForId(int)` bridges from protocol messages into the typed handle and
```
k-wall: nit

```suggestion
integers, so `brokerFor(int)` bridges from protocol messages into the typed handle and
```

proposals/123-router-context-endpoint-type.md
```diff
@@ -0,0 +1,366 @@
...
+
+```java
+// Replaces: var node = ctx.anyNode(route); ctx.sendRequest(node, header, request)
+CompletionStage<ApiMessage> sendToRoute(String route, RequestHeaderData header, ApiMessage request);
+```
+
+`sendRequest(Broker, ...)` is retained for the broker-specific case, where the caller
+already holds a `Broker` from `clientAddressed()` or `brokerFor(int)`.
+
+### Migration
```
k-wall, line 187: Claude pointed out to be that the TopologyService#brokerInfo is in terms of the VirtualNode.  It is not implemented yet, but was described in Proposal 70 (Router SPI)

k-wall:
Thanks for raising this @SamBarker. On the whole I think this looks good.

k-wall (approved):
Approving with nits.

tombentley (commented):
This is well intended, and the reasoning has some merit in the proposal's own terms, but it's missing an important bigger picture issue. Kafka itself doesn't really have "bootstrap broker" or "bootstrap address" as a first class concept. Kafka clients are just configured with a list of normal brokers from which to bootstrap their discovery of the rest of the brokers in the cluster. Kafka brokers just expose their addresses to whichever clients connect to them. 

Kroxylicious added a bootstrap concept of its own, at least for the current networking models. I don't recall why that choice was made. Maybe it was to solve a fundamental issue (if so, we should record what that reason was somewhere). Or maybe we were just following the approach taken by Strimzi. 

One concern I have with this proposal is that it would commit us further to this concept that's not really fundamental to Kafka. And it's a troublesome concept to have to carry, because it means we end up in situation where the proxy doesn't know which broker the client thinks its connected to. That has annoying consequences and makes the proxy more complicated than (it seems) is necessary.

If anything I would prefer if the project was moving in the other direction to the one this proposal would take things: Try to remove bootstrap addresses as a concept.

Now it's quite possible that I'm missing some really great reason why we need to have bootstrap addresses, but I'd like to understand what that is before we go adding API which make it into more of a thing than it is already.

proposals/123-router-context-endpoint-type.md
SamBarker: Good point — `ClientDestination` is a better name. It's a noun, and it reads well on both the type (`ClientDestination`) and the accessor (`clientDestination()`). I'll rename.

proposals/123-router-context-endpoint-type.md
SamBarker: The use-case I have in mind is a rack-aware routing filter. It would intercept fetch requests heading to a broker in a remote AZ and re-route them to a local-AZ broker instead. The fetch request itself doesn't carry the target broker ID — the broker is implicit based on which connection the client sent the request on. So the filter needs the connection's client destination from the `RouterContext` to know which broker the client thinks it's talking to and decide whether to redirect.

proposals/123-router-context-endpoint-type.md
SamBarker: Fixed, thanks.

proposals/123-router-context-endpoint-type.md
SamBarker: Good catch — I've added `TopologyService.brokerInfo(VirtualNode)` → `TopologyService.brokerInfo(Broker)` to the migration table.

SamBarker:
It's news to me that bootstrap is disputed — thanks for raising it.

I think it does represent something real though. Kafka clients model the bootstrap-vs-broker distinction internally (bootstrap nodes get negative IDs, real brokers get non-negative ones), and KIP-899 and KIP-1102 extended it by adding rebootstrap mechanisms. It's not a protocol-level concept, but the ecosystem clearly treats it as a meaningful one.

I also think it helps with client UX — a single discovery address that's decoupled from broker identity is a nice property for a proxy to offer, and matches the Strimzi model.

But the main argument for this proposal is more pragmatic: given where the routing API is today, bootstrap exists in the proxy and router authors already deal with the distinction. This proposal improves the ergonomics by making that distinction type-safe and compiler-checkable rather than encoding it as Optional emptiness. If we later decide to move away from bootstrap, the sealed interface actually makes that easier — the compiler tells you every place that needs to change.

tombentley:
> It's not a protocol-level concept

And that's my point. It does not exist at the level of the protocol, so why are we making it a first class concept in the proxy's network level presentation to clients?

> I also think it helps with client UX — a single discovery address that's decoupled from broker identity is a nice property for a proxy to offer, and matches the Strimzi model.

It might be a nice property for a proxy to offer, but that doesn't in itself mean we should do it.

My problem with it is not some super philosophical point. It's coming from some of the real problems that it's causing us with routing. Problems that only exist because we've created this network address which does not correspond to a broker, but where we're obliged to route requests to a broker. 

So it might be 'nice', but it's coming at a real and increasing cost, in terms of proxy complexity. That's not a good thing.

> But the main argument for this proposal is more pragmatic: given where the routing API is today, bootstrap exists in the proxy and router authors already deal with the distinction.

Yes, undeniably true.

> This proposal improves the ergonomics by making that distinction type-safe and compiler-checkable rather than encoding it as Optional emptiness. If we later decide to move away from bootstrap, the sealed interface actually makes that easier — the compiler tells you every place that needs to change.

I'm not really convinced on that. It allows a developer to scatter a bunch of types around their code today, which we might deprecate in the future. Instead of a single warning on the method they might see a sea of warning about the type use. That's not objectively better.

And this proposal doesn't _just_ do that, it does a lot of other things too, so you can't just the possible deprecation advantage without weighing it against some of the costs of those other changes:

* It's also leaking information (in the form of `Broker.id`) which was _intentionally_ hidden. The virtual node id is not meaningful to developers. I'm not aware they get and benefit from being able to observe it. 
* It's letting developers instantiate their own `Broker` instances (with potentially illegal broker ids!), because the record necessarily has a public constructor. 

It probably wasn't very clearly communicated in the original proposal, but part of the motivation for having an empty `VirtualNode` type, and encapsulating the instantiation within the runtime, was to hide the internal state from developers, with the possible future option of not using plain integers as the underlying identification mechanism in the future -- for example if we supported other networking models.

Throwing this encapsulation away actually really limits our choices (as Kroxylicious developers) in the future.

SamBarker:
As a proxy we necessarily operate in a wider domain than just the wire protocol — the protocol is necessary but not sufficient for what we do. Bootstrap is part of that wider domain; it's how clients discover which brokers exist before they can connect to any of them. It's a real concept, it exists in the runtime today, and router authors handle it today. I'm not proposing to add it, I'm proposing to give something that already exists a proper type.

Given that it already exists, I'd rather it be a meaningful and explicit concept in the API than an implicit one encoded as Optional emptiness. The "sprawl" from type deprecation warnings isn't worse than a single method warning — it's the same callsites, just made visible by the type system rather than left for the developer to discover manually. I'd argue that's better, not worse.

On the encapsulation concerns: I've updated the proposal to make Broker a class with a package-private constructor and no public id accessor. That preserves the opaque nature of the original VirtualNode while still giving it a meaningful domain type. Does that resolve those points for you?

tombentley:
> As a proxy we necessarily operate in a wider domain than just the wire protocol — the protocol is necessary but not sufficient for what we do. Bootstrap is part of that wider domain; it's how clients discover which brokers exist before they can connect to any of them.

Except brokers and Kafka clusters do not have this idea, which demonstrates it is not a necessary concept, or something intrinsic to what we're doing. 

> It's a real concept, it exists in the runtime today, and router authors handle it today.

Yes, that's true. But my point is that the concept itself is problematic. We can make things more type safe as you're proposing, but in a sense you're solving the wrong problem because you're just embracing the wrong concept harder.

This is a small change that does offer some minor value to a router developer. But it does nothing for the larger problem which we should be aiming to address -- that developers have to deal with the whole concept of bootstraps  at all. The mental model needed to write a router is much harder than it needs to be. 

> I've updated the proposal to make Broker a class with a package-private constructor and no public id accessor.

Thanks. However, I don't think that will work in practice: The `Broker` type is in a package of the api module. We don't want a split package situation. So the runtime won't be able to instantiate a `Broker` to give to the router.

PR #125, opened 2026-07-21
Add signoff option to commit message in workflow

tombentley:
The helpful message about how to fix the proposal file number, when followed, meant that the `git push` would fail CI because the `git commit` lacked a signoff. Signing off on a file rename seems to be something that should always be acceptable. (Requiring signoff is about intellectual property, but I struggle to see an IP-related angle on the name of a file).


robobario (approved):
LGTM

PR #129, opened 2026-08-06
[proposal] Add ProtocolLogging filter for wire-level protocol tracing

AdityaThakur1998:
## Summary
 
Add a first-party `ProtocolLogging` filter that renders Kafka requests and responses as a readable, version-aware wire-level trace, giving operators and filter authors a way to see protocol traffic that the proxy currently decodes but never exposes.
 
## Approach
 
The filter implements the catch-all `RequestFilter`/`ResponseFilter` interfaces and forwards every message unchanged. Bodies are rendered by dispatching to Kafka's generated `*DataJsonConverter` classes via the existing generated `KafkaApiMessageConverter`, so there is no per-API formatting code:
 
- **Wire-level trace, not a curated view:** output shows what was transmitted for the configured API keys, rather than an interpreted summary of client behaviour. This is what makes it useful for problems nobody anticipated.
- **Version fidelity as a requirement:** a field that did not exist at the negotiated API version must not appear. Someone debugging version negotiation is exactly who would be misled by a field at a default value that was never sent.
- **Credential bodies structurally withheld:** `SaslAuthenticate`, `CreateDelegationToken`, `AlterUserScramCredentials` and `DescribeDelegationToken` never reach the converter. The envelope is still emitted so the handshake stays visible. The list is hardcoded, deliberately not configurable.
- **No record payloads:** Kafka's converters emit an empty byte array for records-typed fields regardless of content, which matches the intent — the value is in showing message flow, not message content.
- **Gated at `shouldHandleRequest`/`shouldHandleResponse`:** on the configured API key set and the log level, so uninteresting messages are never deserialised despite the catch-all interfaces.
Three configuration keys: `logLevel` (default `DEBUG`), `apiKeyNames` (default all), `maxBodyChars` (default 8192).
 
## Key Benefits
 
- Debugging traffic as the proxy sees it, before or after other filters in the chain — a vantage point broker request logging cannot provide
- Filter authors can see what the messages they handle actually contain
- Generic across all API keys and protocol versions, unlike the hand-written per-API approach of the existing [connect-filter](https://github.com/katheris/kroxylicious-connect-filter) proof of concept
- Same configuration deployable everywhere: output is enabled and disabled by tuning the logging backend on a running proxy, no restart, no config change
## Open points
 
- The three configuration keys become public API, so naming is worth settling here
- `KafkaApiMessageConverter` is currently generated into `kroxylicious-filter-test-support`; its eventual home is expected to be settled by #116, which is already discussing exactly this
- Rejected alternatives (Jackson `ObjectMapper`, configurable field redaction, a payload logging option, request/response latency) are covered in the proposal with reasoning
Relates-to: [kroxylicious/kroxylicious#1121](https://github.com/kroxylicious/kroxylicious/issues/1121)

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/proposals/129-protocol-logger-filter.md
+++ b/proposals/129-protocol-logger-filter.md
@@ -1,7 +1,7 @@
-# 129 - Protocol logging filter
+# 129 - Protocol logger filter
 
 Add a first-party filter that logs Kafka requests and responses passing through the proxy as a
-readable, wire-level protocol trace, for use in debugging and demonstrations.
+readable, wire-level protocol log, for use in debugging and demonstrations.
 
 ## Current situation
 
@@ -24,6 +24,13 @@ Kafka brokers have their own request logging. That is a different vantage point:
 traffic as the proxy sees it, before or after other filters in the chain have transformed it, and
 it is not available to someone running the proxy who does not administer the broker.
 
+## Non-goals
+
+This filter is a development and debugging tool. Production audit logging is explicitly not an
+intended use case: an always-on filter recording all traffic for compliance purposes would need
+sampling, back-pressure handling, a stable machine-readable schema and retention design, none of
+which are proposed here. Framing that as out of scope now avoids the feature drifting into it.
+
 ## Motivation
 
 Three audiences benefit:
@@ -31,10 +38,14 @@ Three audiences benefit:
 - **Debugging.** Seeing the exact requests a client sends, the versions it negotiates, and the
   errors it receives is often the fastest route to a diagnosis. This is particularly true for
   problems that only appear in the presence of the proxy.
-- **Filter development.** Contributors writing filters currently have no convenient way to see
-  what the messages they are handling actually contain.
-- **Demonstration and learning.** The Kafka protocol is not widely understood in detail. A trace
-  showing a client connecting, negotiating versions, authenticating, joining a group and fetching
+- **Filter and router development.** Contributors writing filters or routers currently have no
+  convenient way to see what the messages they are handling actually contain.
+- **Seeing what the proxy itself does.** Placing an instance at each end of the filter chain
+  shows the effect of the filters in between — that a topic policy filter really is rejecting
+  `CreateTopics` for `rf=1`, or what a record encryption filter actually transmits. This is a
+  view of proxy behaviour rather than client behaviour, and nothing else provides it.
+- **Demonstration and learning.** The Kafka protocol is not widely understood in detail. A protocol
+  log showing a client connecting, negotiating versions, authenticating, joining a group and fetching
   makes the protocol concrete in a way documentation does not.
 
 The proxy is unusually well placed to provide this. It already decodes the protocol, it sits
@@ -43,14 +54,14 @@ controls.
 
 ## Proposal
 
-Add a `ProtocolLogging` filter to `kroxylicious-filters`, shipped in the distribution.
+Add a `ProtocolLogger` filter to `kroxylicious-filters`, shipped in the distribution.
 
-### Scope: a wire-level trace, not a curated view
+### Scope: a wire-level protocol log, not a curated view
 
 The filter emits what was on the wire for the configured API keys, rather than a summarised or
 interpreted view of client behaviour. This is a deliberate choice. A curated view — of the kind
 the Connect filter proof of concept produces — is more readable but requires per-API code and
-editorial judgement about which fields matter. A trace is complete, generic, and useful for
+editorial judgement about which fields matter. A protocol log is complete, generic, and useful for
 problems nobody anticipated.
 
 A consequence is that the filter should be version-aware: a field that did not exist at the
@@ -58,6 +69,21 @@ negotiated API version must not appear in the output. For a tool whose purpose i
 was actually transmitted, a field appearing at a default value when the client never sent it is
 misleading, and version negotiation problems are exactly the kind of thing someone would use this
 filter to investigate.
+
+### Record contents are not logged
+
+The filter logs protocol structure, not the data being carried. For the four message types with
+records-typed fields, the record batch — keys, values and headers — does not appear in the output.
+
+This falls out of the implementation: Kafka's generated JSON converters emit an empty byte array
+for records-typed fields regardless of content. It is also the intended behaviour. Record contents
+are user data, potentially personal or regulated, and the value of this filter is in showing
+message flow rather than message content.
+
+Note this does not mean the output is small. Only four message types carry records at all, and
+protocol structure alone can be substantial — a `Metadata` response describing ten thousand
+partitions, or a `Produce` request spanning many partitions, contains no record data and is still
+large. That is what `maxPayloadChars` bounds.
 
 ### Implementation approach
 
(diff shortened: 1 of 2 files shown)


proposals/129-protocol-logger-filter.md
```diff
@@ -0,0 +1,242 @@
...
+- type: ProtocolLogging
+  config:
+    logLevel: DEBUG                                        # default
+    apiKeyNames: [METADATA, FIND_COORDINATOR, JOIN_GROUP]  # absent or empty = all
+    maxBodyChars: 8192                                     # default, must be > 0
+```
+
+| Key | Type | Default | Purpose |
+|---|---|---|---|
+| `logLevel` | SLF4J level | `DEBUG` | Level at which entries are emitted, and the level the backend must be enabled at |
```
robobario, line 175: thought: maybe we should have a higher default level like INFO as the purpose is debugging/educational/demonstrations. If I install this filter I'd like to see output without tweaking the logging config.

robobario (approved):
LGTM except I wonder if we should pump logs out by default. Depends how we imagine it being used, a DEBUG level is better if we imagine it being deployed to production in a quiet mode and then activated by changing the log level in their logging backend. 

But if our focus is on learning/demonstration/debugging-locally then it might make more sense to have it log at INFO by default so there's no co-ordinated log config change required. The user already has to consciously install this Filter for debug.

proposals/129-protocol-logger-filter.md
SamBarker: I'd prefer to set the defaults as Production safe so DEBUG is a very good starting point.

I think any example usage should provide a log4j and or logback snippet configuring the logger. Though it does make me wonder if we should use a dedicated logger for the protocol logging rather than the filters own logger. Allowing the filter to have its own logging?

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+
+Invalid API key names and a non-positive `maxBodyChars` are rejected at startup rather than
+discovered at log time.
+
+### Enabling and disabling at runtime
+
+The filter emits nothing unless the logging backend is also enabled at `logLevel` for the
+filter's logger. Because `shouldHandleRequest` and `shouldHandleResponse` are consulted per
+message, this check is live: an operator can deploy the same configuration everywhere and switch
+tracing on or off by changing the logging backend on a running proxy, without a restart and
```
SamBarker: nit

```suggestion
protocol logging on or off by changing the logging backend on a running proxy, without a restart and
```

I think tracing is a bit ambiguous when talking about logging.

SamBarker (approved):
Thanks @AdityaThakur1998 

A couple of nits but LGTM

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+  "topics" : [ {
+    "topicId" : "AAAAAAAAAAAAAAAAAAAAAA",
+    "name" : "test-logging"
+  } ],
+  "allowAutoTopicCreation" : true,
+  "includeTopicAuthorizedOperations" : false
+}
+```
+
+The envelope comes first so that the output can be scanned and grepped by API key. Proxy-side
```
k-wall: The fact that the header is not JSON is a real impediment to processing in other tools.  If the header were JSON it would be easy to construct a jq pipeline filter that say, extracted all activity for clientId `producer1` or all PRODUCE requests.  
In the current form we force the user into awk/js pipeline gymnastics which is the opposite of developer joy. 

Also in its current form, any addition becomes a breaking change API (say we wanted to log the header's `untaggedFields`) as the format of the space separate string changes. 

I would suggest something like: 

```
{
   "header": {
        type: "REQUEST"
        apiKey:
        apiVersion: 
        correlationId:
        clientId:  
   }
  "payload: {
      ...
   }
}
```

This leaves things open for extension - adding things like `authenticatedPrincipal` or `clientIP` becomes easy.  It will give us a place for new things like
https://cwiki.apache.org/confluence/spaces/KAFKA/pages/406624038/KIP-1313+Client+instance+ID+in+all+request+headers


Note that you probably **don't** want to just serialise the RequestDataHeader/ResponseDataHeader to JSON.  In Kafka the responses headers don't carry the apiKey etc.    We want it to be present so that say, your jq filter can extract produce request/responses.

```
jq -c 'select(.header.apiKey == "PRODUCE")'
```

proposals/129-protocol-logger-filter.md
```diff
@@ -0,0 +1,242 @@
...
+- **Filter development.** Contributors writing filters currently have no convenient way to see
+  what the messages they are handling actually contain.
+- **Demonstration and learning.** The Kafka protocol is not widely understood in detail. A trace
+  showing a client connecting, negotiating versions, authenticating, joining a group and fetching
+  makes the protocol concrete in a way documentation does not.
+
+The proxy is unusually well placed to provide this. It already decodes the protocol, it sits
+between client and broker, and it sees traffic at a point in the filter chain that the operator
+controls.
+
```
k-wall, line 54: It might be worth calling out that production 'audit' is not an intended use-case for the this feature.   This clearly sets expectations and avoids future scope creep

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+module from a shipped filter is not ideal; the longer-term home for this generation is expected
+to be settled as part of proposal 116, and the initial implementation can be repointed when that
+lands.
+
+### Output format
+
+Each entry is a human-readable envelope line followed by the message body as indented JSON:
+
+```
+REQUEST  METADATA v13  corr=1  client=producer-1
```
k-wall: There's a wrinkle with the correlation id, that the spec should point out.  The logged correlation id may or may not be a client correlation id. 

* Filters have the ability to dispatch out-of-band requests towards the upstream. FilterContext#sendRequest.
* Routers have analogous behaviour.  io.kroxylicious.proxy.router.RouterContext#sendRequest.
 
These requests will have 'synthetic' correlation ids (negative values).  Depending on the position of the filter in the chain, it may see this requests too.

Filters like this one have no "nice" way to know whether they are seeing a client or synthetic request.

I don't see this as a show stopper for this proposal, but we should note it in the text.

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+it is not available to someone running the proxy who does not administer the broker.
+
+## Motivation
+
+Three audiences benefit:
+
+- **Debugging.** Seeing the exact requests a client sends, the versions it negotiates, and the
+  errors it receives is often the fastest route to a diagnosis. This is particularly true for
+  problems that only appear in the presence of the proxy.
+- **Filter development.** Contributors writing filters currently have no convenient way to see
```
k-wall: nit: Filter and Router

proposals/129-protocol-logger-filter.md
```diff
@@ -0,0 +1,242 @@
...
+- **Debugging.** Seeing the exact requests a client sends, the versions it negotiates, and the
+  errors it receives is often the fastest route to a diagnosis. This is particularly true for
+  problems that only appear in the presence of the proxy.
+- **Filter development.** Contributors writing filters currently have no convenient way to see
+  what the messages they are handling actually contain.
+- **Demonstration and learning.** The Kafka protocol is not widely understood in detail. A trace
+  showing a client connecting, negotiating versions, authenticating, joining a group and fetching
+  makes the protocol concrete in a way documentation does not.
+
```
k-wall, line 50: I think there's a use-case to help one visualise the behaviour that your proxy is bringing to your Kafka system.

You could put a protocol logging filter at the front and back of filter chain to let you see that, yes your topic policy filter is truely, rejecting topic create requests for topics with rf=1.

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+connection, so the session is required to pair correctly.
+
+### Configuration
+
+```yaml
+- type: ProtocolLogging
+  config:
+    logLevel: DEBUG                                        # default
+    apiKeyNames: [METADATA, FIND_COORDINATOR, JOIN_GROUP]  # absent or empty = all
+    maxBodyChars: 8192                                     # default, must be > 0
```
k-wall: When you say body, what do you mean?  You've said produced/fetch records aren't logged, so I'm not really sure what is meant by body.

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+The envelope comes first so that the output can be scanned and grepped by API key. Proxy-side
+context — notably the session identifier — is emitted as structured key-values rather than in the
+envelope, since it is proxy state rather than protocol. Requests are paired with responses on
+session identifier plus correlation identifier; correlation identifiers are only unique within a
+connection, so the session is required to pair correctly.
+
+### Configuration
+
+```yaml
+- type: ProtocolLogging
```
k-wall: In the use-case where I want to put more than one ProtocolLogging in filter chain (say upstream end and downstream end), I've got no way to identify the output from one ProtocolLogging from the other.  This devalues the output considerably.  

Filters currently don't have a unique name. Nor can they tell their position in the filter chain or what router they are attached to.  That's really a filter/router API question (for discussion elsewhere).

In order not to bog this proposal down with that, I wonder if the ProtocolLogger was able to specify its own logger name, you'd be able to declare two instances of it which distinct logger names.  If the user wished.  I am not suggesting we try to validate the logger names they chose.  WDYT?

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+  showing a client connecting, negotiating versions, authenticating, joining a group and fetching
+  makes the protocol concrete in a way documentation does not.
+
+The proxy is unusually well placed to provide this. It already decodes the protocol, it sits
+between client and broker, and it sees traffic at a point in the filter chain that the operator
+controls.
+
+## Proposal
+
+Add a `ProtocolLogging` filter to `kroxylicious-filters`, shipped in the distribution.
```
k-wall: `ProtocolLogger` would be the noun-phrase.

proposals/129-protocol-logging-filter.md
```diff
@@ -0,0 +1,242 @@
...
+
+```
+REQUEST  SASL_AUTHENTICATE v2  corr=2147483642  client=producer-1
+<body withheld: credential-bearing API>
+```
+
+Making this list configurable was considered and rejected. A user with a concrete need can raise
+an issue; the default should not be something an operator can accidentally weaken.
+
+Record payloads are not logged. Kafka's generated JSON converters emit an empty byte array for
```
k-wall: The fact  the filter doesn't log records is an important point.  Let bring that out in the Scope section.

k-wall: I'd like to question on whether excluding records is the right thing.  Kafka exists to transport records about.   Records are things business care about. So I think most often users will want to see the records.

Imagine using the protocol logging filter to demonstrate the filters we have today. 
* record encryption (i want to see the bytes are encrypted!), 
* record validation (i want to see that the record that record validation has just rejected).

Maybe keeping it out of scope for the first version is the right thing.  But I can imagine adding a switch 'logRecords: true' which serialises produced and fetched records.

I'm not asking for an immediate change to scope - just let's discuss.

proposals/129-protocol-logging-filter.md
AdityaThakur1998: It means the serialised message excluding the header, so it's what gets truncated when a message is very large. Record contents aren't included, but the structure is, and that can still be substantial.Given your other comment about a JSON envelope with `header` and `payload`, `maxPayloadChars` names the thing it actually truncates.Will Change it to `maxPayloadChars` in configuration section to avoid confusion

proposals/129-protocol-logging-filter.md
AdityaThakur1998: It avoids inventing filter identity that doesn't exist. Adopting it would be a good solution.


```
filterDefinitions:
  - name: trace-downstream
    type: ProtocolLogger
    config:
      loggerName: protocol.downstream
  - name: trace-upstream
    type: ProtocolLogger
    config:
      loggerName: protocol.upstream
```

Two things it gets for free: the logger name already appears in the log line, so the instances are distinguishable without adding anything to the entry itself; and each becomes independently controllable from the logging backend, so you can enable just the upstream one while investigating.

Defaulting to the filter class name when unset, and agreed on not validating what the user picks.

Added in Configuration

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Your demo cases are convincing, and one of them reframes something from the earlier discussion(https://github.com/kroxylicious/kroxylicious/issues/1121) of chain ordering as an argument against logging records — that placed after the record encryption filter you'd see ciphertext rather than plaintext. In your framing that's the whole point: seeing that the bytes really are encrypted is the demonstration. Same behaviour, opposite conclusion, depending on whether records are being logged to inspect business data or to observe what a filter did to them.

**One thing worth putting into the discussion, because I think it changes the cost:** logRecords would not be a switch on the existing path. Kafka's generated converters hardcode an empty byte array for records-typed fields — I checked PartitionProduceDataJsonConverter.write in kafka-clients 4.2.0 and there's no mode that preserves the bytes. The three-argument overload emits recordsSizeInBytes instead, which is metadata rather than content.

So logging records means bypassing the converter for those fields and serialising MemoryRecords ourselves — iterating batches and records, and deciding how to render arbitrary bytes. That last part is a real design question rather than an implementation detail: record values may be Avro, protobuf, UTF-8 or ciphertext, and base64 would serve the encryption demo well while making the validation demo unreadable.

Only four message types carry records-typed fields (ProduceRequest, FetchResponse, ShareFetchResponse, FetchSnapshotResponse), and only three are reachable through a client-facing proxy, so the surface is small — but it's a distinct piece of work with its own design decisions rather than a flag.

Given that, I'd suggest keeping it out from v1. What do you think? @SamBarker @robobario

proposals/129-protocol-logging-filter.md
k-wall: Ok, as you'll see from the comments in the PR, I'm doubtful about whether we really want it.

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Fixed all of the occurrences

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Agreed on both counts  the jq-ability and, more importantly, that a positional space-separated format is frozen the moment anyone parses it. Restructured to {"header": {...}, "payload": {...}}, and taken your point about constructing the header rather than serialising ResponseHeaderData, since response headers carry only a correlation id and .header.apiKey needs to work on both directions.

One constraint worth stating alongside the jq example: the entry is JSON but the log line isn't, because the backend's timestamp, level and logger prefix surround it. Piping to jq needs an appender emitting %msg%n only, or a JSON layout for the whole event. Noted in the proposal and I'll cover it in the README.

Truncation changes shape as a result — cutting the serialised string would now corrupt the header too, so payload is truncated before composition and flagged, keeping the entry parseable.

Added in the Output format section

proposals/129-protocol-logger-filter.md
AdityaThakur1998: Added in the Non-goals section as a reminder

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Added

proposals/129-protocol-logger-filter.md
AdityaThakur1998: added in the use case

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Added in the scope

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Noted it in the Output section

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Fixed it on every occurrence

proposals/129-protocol-logging-filter.md
MarkNSweep: Just adding a +1 to the idea of a switch. I think that filter developers would likely want the ability to log the records. When writing custom business logic in filters a dev is as likely to be interested in the message as the protocol wrapper. I've found in the past that even the option to just include the record contents as a base 64 encoded field works. I can put that in a hex viewer and just visually pick out the text or change that I am interested in and whether it matches what I expected to see.

robobario: I agree @AdityaThakur1998 lets leave it as a future extension because we'd have to do some exploring about what to emit and how. I think the section Keith has commented on should be rephrased to say something about how it is a future extension, or move it to a Future Extensions section at the bottom.

base64 for record bodies and header values works for me but I think we may also want a total accounting of the record batch attributes so users can see which records are control records, what compression is used and so on. Having a high-fidelity YAML representation could also help our tests where we want to check that batching and record attributes are preserved. This will likely get easier to implement after #116 is implemented.

proposals/129-protocol-logging-filter.md
AdityaThakur1998: I have removed it and will make changes in the PR. We may include it in the future for the records.

proposals/129-protocol-logging-filter.md
AdityaThakur1998: Thanks Everyone for the discussion. I have added it as a future extension for this design proposal.

proposals/129-protocol-logger-filter.md
robobario: happy with DEBUG default, it's in the binary distro so keeping it safe works for me. The design covers letting users configure the logger names

proposals/129-protocol-logger-filter.md
```diff
@@ -0,0 +1,360 @@
...
+Note that decoding is shared across the chain, so this only avoids cost for API keys that no
+other filter in the chain requires.
+
+### Security
+
+Bodies of credential-bearing API keys are never logged. The exclusion list is hardcoded and
+deliberately not configurable:
+
+- `SaslAuthenticate` — carries credential material for PLAIN and SCRAM
+- `CreateDelegationToken` — the response carries the token HMAC, which is itself a credential
```
robobario, line 232: couple more to add: `RenewDelegationToken` and `ExpireDelegationToken` both also carry the HMAC

AdityaThakur1998: Fixed it in the latest commit

